PR #88, opened 2016-11-19
I18n pl

KrzysztofMadejski:
Some more Polish translation.

Hopefully in the future we can interest some Polish shelter with this solution.


bobintetley:
thank you for your efforts!

Issue #207, opened 2017-05-03
JSON web service for single animal data


inghamn:
It would be nice to have a service we could hit to return the data for a single animal.

While developing the Drupal integration, the general pattern is to have one screen that lists animals.  Each animal in that list would have a link to an animal detail screen.  Having a web service we can hit that matches would make this easier.

If the web service is read-only, can these be unauthenticated requests?

bobintetley:
animalview is not authenticated because it is a public visible call to return a page. The data owner has full control over the template and the data returned and nothing potentially sensitive is displayed.

the adoptable_animals calls are authenticated because they contain some private data such as details of the person who surrendered the animal. I don't see a problem with a call to retrieve one animal's data by ID but would have expected a caller to cache the complete adoptable set to avoid coming back to us for every animal.

Have you looked at the adoptable_js service call? It uses javascript to embed the adoptable list in your page and links to animalview to show the details. There's an example here of a customer using adoptable_js with animalview and online forms: http://greyhoundrescuewales.co.uk/asm-dynamic-upload/

bobintetley:
another reason for authentication is so that as a server operator you have the ability to keep unwanted clients from using the service. If you use the service cache that should alleviate DoS to some degree, but if you only want your Drupal instance accessing the data then it makes sense to put authentication on it.

bobintetley:
I don't see an issue with authentication on the single animal call as it keeps out unexpected clients and only means passing two extra parameters. I'm adding them to a new commit as json_adoptable_animal and xml_adoptable_animal instead to fit the existing name scheme.

Also, those calls should be using publish.get_animal_data as the source rather than animal.get_animal as the publish routine does extra bits of cleanup for bonded animals, sorting out the bio, etc.

Thanks for your contribution and for raising this!

PR #208, opened 2017-05-03
Added a JSON service to return a single animal

inghamn:
Updates #207

This is just a naive implementation.  One thing that jumps out at me is: why does json_adoptable_animals require authentication, and animal_view does not?

### DIFF
--- a/src/service.py
+++ b/src/service.py
@@ -279,6 +279,13 @@ def handler(post, path, remoteip, referer, querystring):
             else:
                 return set_cached_response(cache_key, "image/jpeg", 86400, 120, dbfs.get_string(dbo, mm[0]["MEDIANAME"]))
 
+    elif method == "animal_json":
+        if animalid == "" or utils.cint(animalid) == 0:
+            al.error("animal_json failed, %s is not an animalid" % str(animalid), "service.handler", dbo)
+            return ("text/plain", 0, "ERROR: Invalid animalid")
+        else:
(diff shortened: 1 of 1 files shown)


bobintetley:
animalview is not authenticated because it is a public visible call to return a page. The data owner has full control over the template and the data returned and nothing potentially sensitive is displayed.

the adoptable_animals calls are authenticated partly because they contain some private data such as details of the person who surrendered the animal. I don't see a problem with a call to retrieve one animal's data by ID but would have expected a caller to cache the complete adoptable set to avoid coming back to us for every animal.

Another reason for authentication is so that as a server operator you have the ability to keep unwanted clients from using the service. If you use the service cache that should alleviate DoS to some degree, but if you only want your Drupal instance accessing the adoptable data then it makes sense to put authentication on it and keep everyone else out.

Have you looked at the adoptable_js service call? It uses javascript to embed the adoptable list in your page and links to animalview to show the details. There's an example here of a customer using adoptable_js with animalview and online forms: http://greyhoundrescuewales.co.uk/asm-dynamic-upload/

inghamn:
Thank you for the merge.

adoptable_js is not a good match for our use case.  Our Shelter Manager is behind a firewall.  Members of the public aren't able to hit it with their browser.  All service requests are being done server to server.  In any case, we'd need full control over the HTML that gets inserted, so that it is incorporated into the Drupal theme.

Issue #248, opened 2017-06-29
Anisble playbook for installation and upgrading


inghamn:
We use Ansible for maintaining our servers, and it would be nice to have an Ansible playbook for installation and upgrading of ASM.   We typically include the anisble playbooks as part of the web application's repository.

inghamn:
I'm working through how to have Ansible write out sitedefs.py for different versions of ASM as new versions get released.

inghamn:
Ansible now writes out a templated sitedefs file

PR #249, opened 2017-06-30
Started work on Ansible playbook

inghamn:
Updates #248 

Here's my first pass at using Ansible to set up Shelter Manager.  All the installation stuff works, although I run into trouble when I'm trying to host Shelter Manager in a subdirectory.  (http://somehost.com/asm)

### DIFF
--- a/ansible/roles.yml
+++ b/ansible/roles.yml
@@ -1,0 +1,4 @@
+- src: City-of-Bloomington.linux
+- src: City-of-Bloomington.apache
+- src: City-of-Bloomington.mysql
+- src: City-of-Bloomington.wsgi
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,4 @@ cordova/asm/platforms
 .DS_Store
 /.project
 /.pydevproject
+/ansible/roles
--- a/ansible/README.md
+++ b/ansible/README.md
@@ -1,0 +1,174 @@
+# Ansible Playbooks
+
+Ansible is a configuration management system. For more information about ansible, see:
+
+http://www.ansible.com/how-ansible-works
+
+## Terminology
+
+The primary machine used to issue those commands is called the control machine. Hosts are the remote machines that you configure with ansible. For more details about terms, see here:
+
+http://docs.ansible.com/ansible/glossary.html
+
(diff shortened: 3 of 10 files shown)


ansible/roles.yml
```diff
@@ -0,0 +1,4 @@
+- src: City-of-Bloomington.linux
```
bobintetley: are these files missing from the commit? Also, naming seems a bit customer/branch specific? Did I miss something, I have no experience of ansible?

inghamn: The files are all there in the commit.  They are all in an "ansible" directory at the root of the project.  The idea is to be able to clone the project, then use ansible to deploy it to a server.  Admittedly, this ansible work is targetting an Ubuntu 16.04 server, but it's a start.

inghamn: This is the ansible work we did to deploy ASM on our servers back in June.  The commit is a little out of date now, as the ansible roles it depends on have changed.

For our web applications, we have found it useful to have deployment scripts included as part of the project. (As opposed to a seperate Github repository).  Is this a strategy that sounds like it would be useful for you?

bobintetley: there are 10 files in this commit and none of them are called City-of-Bloomington.anything ? It looks like this roles.yml file is referencing additional files not in the commit (or are they role names rather than files and I've misunderstood?). 

I'm fine including deployment-related stuff in the repo, Ansible looks very interesting and I'm sure there are similar automation products like puppet and chef that are popular too.  I just wanted to make sure that everything was there before merging and that the names were ok.

If we're good I'll merge it in so that further mods can be made to it. Apologies for taking so long to spot this pull request :)

inghamn: No worries, thanks for reviewing it.  You are correct the role dependencies are not included (all the City-of-Bloomington.*) directories.  They are pulled down using ansible's dependency management system, Galaxy.

```bash
ansible-galaxy install -r roles.yml
ansible-playbook asm_install.yml -i /path/to/your/inventory
```

inghamn:
We are looking forward to your next release.  One of the things we're going to be doing will be to migrate our installation off MySQL and onto Postgres.  This will be reflected in the Ansible playbook.  I'll be rewriting the role to use Postgres, instead of MySQL.

PR #356, opened 2018-02-06
Ansible script updated to v40

inghamn:
Ansible script now also sets up a daily cron script and
installs memcache.  This is still using MySQL for now.


inghamn:
Updated to add Postgres support.

PR #709, opened 2019-12-17
Fixed AppointmentFor wordkey returning the username.

### LINKED ISSUE #705: Clinic AppointmentFor token
The AppointmentFor wordkey is returning the username. It should resolve the real name of the user.

julian-poidevin:
Now resolves the real name of the user.

Fixes #705

### DIFF
--- a/changelog
+++ b/changelog
@@ -1,6 +1,6 @@
 43
 ================
-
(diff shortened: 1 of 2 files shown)


bobintetley:
Thanks for your contribution! This could be much simpler by using the call asm3.users.get_real_name instead. I've committed a change with that instead.

Issue #752, opened 2020-03-27
POT file generation may be inaccurate


julian-poidevin:
Hello,

I've been translating ASM in french for a few month now and I've always wondered why there was still some parts of the app not translated at all whereas it seemed to appear in po file.

Here's some weird things I noticed :

### 1 - English po translation files has more entries than the French file
Here's a comparison using Poedit software : 

| `en_GB.po`  | `fr.po`  | `asm.pot` |
|---|---|---|
|  ![image](https://user-images.githubusercontent.com/8756421/77801666-94eaf000-7079-11ea-9ae9-6020de4bb61c.png)  | ![image](https://user-images.githubusercontent.com/8756421/77801651-8a305b00-7079-11ea-8837-24130787d6da.png) | ![image](https://user-images.githubusercontent.com/8756421/77801751-c1067100-7079-11ea-96b5-fb5d45224648.png) |

After using the "Catalog ➡️ Update from POT file" menu in Poedit, here's the result for the `fr.po` file

| `fr.po`  |
|---|
| ![image](https://user-images.githubusercontent.com/8756421/77801868-fd39d180-7079-11ea-9c2c-18cb625f6ff3.png) | 

Using this method, some parts of the app are not translated : 
![image](https://user-images.githubusercontent.com/8756421/77802513-51918100-707b-11ea-8863-25ac5a0d9466.png)

### 2 - Updating to POT file corrupt good setup
When having a good setup with every part of the software translated, if updating the translation with `asm.po`, it corrupts it and we get parts non translated again.

### 3 - Regenerating asm.pot with an other method saved the problem.
I managed to fix this issue by using [this method on StackOverflow](https://stackoverflow.com/a/44705132)

When converting en_GB.po into a POT file and updating `fr.po` with it, this works fine and all software parts are being translated correctly.

Here's the app correctly translated : 
![image](https://user-images.githubusercontent.com/8756421/77802818-fe6bfe00-707b-11ea-9635-e17e6edc8219.png)

-----------

I'm of course using `make translation` before each compilation. Do you think there could be a problem with the `make pot` command ? 

Here's a ZIP file with OLD & NEW po and pot files, this might help you out : [ASM-PO-DEBUG.zip](https://github.com/bobintetley/asm3/files/4396053/ASM-PO-DEBUG.zip)

Thanks

bobintetley:
make pot is only supposed to update the pot file, it's upto the language translators tool to update their po file from it. 

[...]

PR #753, opened 2020-03-27
Update french translation

julian-poidevin:
Fixed POT file issue causing some part of the app not being translated


julian-poidevin:
Related to #752

PR #754, opened 2020-03-27
Fix return movement call not updated with new prototype

julian-poidevin:
Fixed return_movement calls not updated with new prototype
Added "username" to calls


bobintetley:
thanks for the save, I can't believe the unit tests passed with it in that state.

julian-poidevin:
No problem 😉

PR #788, opened 2020-04-20
Removed python3-sqlite dependency

julian-poidevin:
`python3-sqlite` is now part of `python3 `standard library and no longer needed as a dependency.

Please see Animal Shelter Manager 3 forum post called _"ASM3 package install failing because of python3-sqlite dependancy"_ for more details.


bobintetley:
Thanks

PR #808, opened 2020-04-29
Removed seconds from some time wordkeys when not relevant

julian-poidevin:
For some time related wordkeys, seconds are not relevant, here's a proposed change for the **clinic** wordkeys.

I don't know how much of an impact it could have for all ASM users so I'll understand if you don't accept this PR. Plus I'm not an experienced python/javascript developer so it may be wrong.

### DIFF
--- a/src/asm3/i18n.py
+++ b/src/asm3/i18n.py
@@ -6,8 +6,8 @@ import time
 # flake8: noqa - we have a lot of locales and this is convenient
 from asm3.locales import *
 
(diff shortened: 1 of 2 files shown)


bobintetley:
I think it makes sense. I'll accept the PR, but I hope you don't mind I'm going to change it a little bit. I would make i18n.time_format accept an optional format parameter and make it consistent with i18n.date_format.

Thanks for your work!

bobintetley:
also, I think you're right - seconds is not really necessary for anything in wordprocessor.py, humans just don't work in or need those seconds!

PR #837, opened 2020-05-27
Update quality_control.txt

kwinlr:
Un-commented 'AND adoption.istrial = 0'


bobintetley:
thanks Karen, updated the repository now with this change

PR #852, opened 2020-06-09
Added electronic signature option to asm3.conf example file

julian-poidevin:
The option does not appear in the example configuration file


julian-poidevin:
Is that PR not relevant ? Thanks

PR #881, opened 2020-07-19
Watermark Media

Connor9220:
This is the code for watermarking media with a .png file and the animals Name in a font.

Here is a sample image of the output.

sample

Also, I'm not sure why src/asm3/i18n.py was updated unless one of the build process update it. ??


julian-poidevin:
Loving it ! Good job, @Connor9220.
Yes, `i18n.py` is updated during the build.

Connor9220:
So, how do we deal with that.. just add that to the .gitignore file?

julian-poidevin:
I just discard the changes and don't include it to my pull request personally.
Anyway it's going to be recompiled to re-generated afterwards.

src/asm3/media.py
```diff
@@ -908,3 +925,68 @@ def scale_all_pdf(dbo):
...
 
+def watermark_with_transparency(imagedata, animalName):
+    """
+    Watermark the image with animalName and logo. 
+    """
+    try:
+        inputd = asm3.utils.bytesio(imagedata)
+        base_image = Image.open(inputd)
+
+        watermark = Image.open("/var/www/vhosts/asm.maarcadopt.org/watermark.png")
```
bobintetley: Ideally, the watermark image should be in the dbfs, so the user can upload watermark.png under Settings->Reports->Extra Images (no changes needed). 

Image data can then be loaded with:
```
asm3.dbfs.get_string_filepath("/reports/watermark.png")
```

src/asm3/media.py
```diff
@@ -908,3 +925,68 @@ def scale_all_pdf(dbo):
...
+        stroke = 3
+        transparent = Image.new('RGB', (width, height), (0,0,0,0))
+        transparent.paste(base_image, (0,0))
+        transparent.paste(watermark, position, mask=watermark)
+        draw = ImageDraw.Draw(transparent)
+
+        font_offset = 20
+
+        for fontsize in range(20, 180, 5):
+            font = ImageFont.truetype("/usr/share/fonts/truetype/freehand 575 bt.ttf", fontsize)
```
bobintetley: Watermark font path should be a sitedef/config item. It might also be wise to fall back to something available on most distros if not os.path.exists - eg: /usr/share/fonts/truetype/dejavu/DejaVuSans.ttf

src/asm3/media.py
Connor9220: Your welcome to make that change to the code. However, I did update the code in my 3rd and 4th commits to use sitedef to vs hard coding the files.

src/asm3/media.py
Connor9220: Again, I did make a change to use sitedef for the font file location, along with other configurable options, such as the font stroke, colors, offsets etc..    Fall Back is a good idea.

I was trying to figure out how to best handle it if they don't have a font or watermark file defined.  I.E. hide the button.

bobintetley: Sorry, I didn't see the other commits, just this one in the pull request. 

I'd have the backend check for watermark.png availability with asm3.dbfs.file_exists("watermark.png") and pass a boolean to the frontend to control the button visibility.

There are probably a few more edge cases to consider - what if an animal's name changes, or the shelter changes logo? If watermarking created a copy of the image rather than altering the original file, it would be easy to delete the watermarked image and generate a new one.

Connor9220: > Sorry, I didn't see the other commits, just this one in the pull request.
 
No problem.  I'm Kinda new to this, probably would have been better if I had only one pushed commit.  I was having some issues with merging my stuff down.

> I'd have the backend check for watermark.png availability with asm3.dbfs.file_exists("watermark.png") and pass a boolean to the frontend to control the button visibility.

So, Yea, checking for the watermark.png file is defined in sitedef or exists in the asm3.dbfs  was what I was thinking. I just wasn't sure how to pass that from the python to the media.js file.  I was looking for examples of how that was done.  I looked into how signaturetype, but wasn't understanding how it was being passed.  

> There are probably a few more edge cases to consider - what if an animal's name changes, or the shelter changes logo? If watermarking created a copy of the image rather than altering the original file, it would be easy to delete the watermarked image and generate a new one.

So, I had thought about this to some degree, for my rescue's use, it's not a huge dea., We're using S3 and have versioning turned on, so, I can go and manually pull the original photo if needed.  I didn't want to clog up the media area with both watermarked, and non-watermarked versions.  Maybe we add a db flag to indicate the photo is watermarked, and if both a watermarked and non-watermarked version exist, we only display the watermarked version, and give them the button to revert back if needed?

Connor9220:
Updated it to use asm3.dbfs.get_string_filepath(dbo, "/reports/watermark.png")

Also have it check for the file in the dbfs, if not found, fallback to the WATERMARK_FILE (Which is faster btw).

Issue #890, opened 2020-07-30
Email reports with no data


bobintetley:
The current behaviour is to not bother emailing a report if it has no data to display. Some users would like the report with no data so that they know that was the reason rather than a fault. 

Add a system option to email empty reports.

PR #896, opened 2020-08-19
Updated french translation

julian-poidevin:
**Monthly update of french translation**

Update for watermark feature (PR #881)
Update for Email reports with no data (#890)
And others


bobintetley:
Thanks!

PR #956, opened 2021-01-25
Hebrew translation fixes and RTL design fix

saggineumann:


bobintetley:
I had to make a few fixes - there were still RTL CSS declarations being applied in the javascript in common.js/bind_widgets that conflicted with some of the declarations in the CSS file. 

I've removed all javascript RTL CSS from common.js and made sure it is now only in the CSS file. Some of the selectors in the file were wrong (looking for class instead of id), I think I've got them all correct now, but let me know if I've made any mistakes. Future RTL fixes should only be applied to the asm-rtl.css file.

Also, the link tag was being appended to the head for every module load (when bind_widgets is run), which can happen many times before there is a page transition.

Thanks for your work, putting these into a separate CSS file was a great idea and definitely the way to go!

Commit is c854db06

saggineumann:
Thanks @bobintetley ! I'll test your changes soon and let you know

saggineumann:
@bobintetley Looks good!

PR #972, opened 2021-02-21
pp-cardcom

saggineumann:
Added support for Cardcom payment provider:
1. create a popup window in cardcom PoS style (either type in CC details or use an attached device to swipe card)
2. returned data also contains CC token info to be used without collecting CC information again. This data is stored in person extra ids.


saggineumann:
Added method to use token (where available) to make the charge

@bobintetley can you please review & merge?

src/asm3/configuration.py
```diff
@@ -354,7 +354,8 @@
     "WeightChangeLog": "Yes",
     "WeightChangeLogType": "4",
     "CardcomSuccessURL": "https://secure.cardcom.solutions/DealWasSuccessful.aspx",
-    "CardcomErrorURL": "https://secure.cardcom.solutions/DealWasUnSuccessful.aspx"
+    "CardcomErrorURL": "https://secure.cardcom.solutions/DealWasUnSuccessful.aspx",
+    "CardcomDocumentType": "3"
```
bobintetley, line 358: Unless the Cardcom URL changes per customer, these shoud be part of sitedefs.py. CardcomDocumentType looks like a constant that should be in PaymentProcessor subclass?

src/static/js/donations.js
```diff
@@ -488,6 +502,59 @@ $(function() {
                 return false;
             });
 
+            $("#button-cardcom-token").click(function() {
+                payment_processor_confirmation_dialog("cardcom");
+                return false;
+            });
+
+            const payment_processor_confirmation_dialog= async function(processor_name) {
+                let row = tableform.table_selected_row(donations.table);
```
bobintetley, line 511: function should at least have cardcom in the name. Might be nice to move it out at least to a higher level as a member of donations object or if there is going to be more cardcom UI stuff, to a separate file?

src/static/js/donations.js
```diff
@@ -354,8 +354,22 @@ $(function() {
                 s += '<li id="button-cardcom" class="processorbutton asm-menu-item"><a '
                         + ' target="_blank" href="#">' + html.icon("cardcom") + ' ' + _("Cardcom") + '</a></li>';
             }
+
+            if (controller.person) {
```
bobintetley, line 358: Other processors and functionality usually check sitedefs to make sure they have been configured, then pass a "hascardcom" boolean in the controller. That can be used to disable elements of the UI if they haven't been configured - this looks like a good spot to not add the cardcom button if there are no URLs configured in sitedefs or it has not been configured in options.

src/asm3/configuration.py
saggineumann: They're actually per customer - they're what Cardcom will redirect to on success or error. CardcomDocumentType is a default value that can be changed in the config. 3=receipt which may be used by pretty much everybody (that's why I set it as default) but other values can be set depending on the org's entity type (nonprofit/forprofit) and the whether they want Cardcom to produce a document or not

src/static/js/donations.js
saggineumann: I put it in the same level as the other buttons' functions and at first I thought this (just like the payment_processor_popup function) can actually be implemented with other payment providers, but you're right that the extra ID keys do contain Cardcom in their names... 

Perhaps we can just add a per-provider function to get the confirmation text (or just a switch/case if block by processor_name)?

src/static/js/donations.js
saggineumann: I added the same check I saw used with Stripe and Paypal - `if (config.str("CardcomUserName"))` (similiar to `if (config.str("StripeKey"))`). Should I replace it with something else?

For the token button I forgot to add the `config.str("CardcomUserName")` part but there are extra tests to check for another config property (`config.str("CardcomUseToken")`) and it also tests the specific person for extra ids and doesn't show if there aren't any.

What should I do?

src/asm3/configuration.py
bobintetley: ah, disregard, that's fine then. Same for the others, I'll do the merge and we can tweak from there. Thanks!

bobintetley (approved):

PR #984, opened 2021-03-19
Custom code injection options

saggineumann:
@bobintetley This PR adds the functionality to add custom javascript in html header or "footer" (before </body>).

I copied the html editor tags from reports.js to edit the js code in options.js, but it doesn't seem to work as nicely for some reason.

I'm also not sure if I shouldn't have added the custom js in other *header() functions in html.py.


bobintetley:
Discussed and replaced with JSInjection config item

PR #992, opened 2021-04-08
Add RescheduleDays to TestType lookup

### LINKED ISSUE #921: Reschedule date when performing tests
Similar to vaccinations, the test "Perform" button should have an optional reschedule date to create a copy of the test in the future. Some shelters perform heartworm tests every month and this would be useful to them.

wundrian:
This is a straight-forward clone of the vaccination flow. closes #921

@bobintetley one thing I didn't do (yet) is pre-filling the "Retest" field when creating a new test. I thought about adding a `onchange` handler to the type drop-down that would update the retest date. Would this be useful? Right now, a re-test would be just added when you "perform" a test.

### DIFF
--- a/src/asm3/dbupdate.py
+++ b/src/asm3/dbupdate.py
@@ -1527,6 +1527,7 @@ def sql_structure(dbo):
         fstr("TestName"),
         fstr("TestDescription", True),
         fint("DefaultCost", True),
+        fint("RescheduleDays", True),
         fint("IsRetired", True) ), False)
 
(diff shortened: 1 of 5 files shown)


bobintetley:
Thanks for this! Only bit missing is an update for existing databases in dbupdate, which I'll add. I think keeping this on the perform button is more consistent as it matches vaccination.

wundrian:
I'll keep dbupdate in mind for the next time, thanks!

PR #996, opened 2021-04-16
More heb translation

saggineumann:


bobintetley:
This looks like the code injection stuff again that we already have some code for rather than heb translation?

saggineumann:
Odd... I suppose this branch was based on the previous branch where I worked on the code injection stuff, but it doesn't contain any of the work I've done on the po file. I'll try to re-do it.

PR #1011, opened 2021-05-12
Added Re-Index button to online forms

SoulOfSet:
Been running into major re-factors of forms lately. Some had been setup with space in between indexes but even those get filled over time.

All this does is add a button when you are viewing the form editor that resets all the indexes (in respect to their current order) by starting at 10 and adding 10 for each field.

Example random display indexes
image

Dialog
image

After reload forced by js
image


bobintetley:
Thanks, good solid feature. Only thing I dislike is the reindex dialog code in common_tableform but I'll tweak that. Thanks again for contributing!

SoulOfSet:
Thanks. Yeah looking back, maybe late nights is not my prime for elegant wording.

PR #1032, opened 2021-07-17
add installments support in cardcom token pp

saggineumann:


src/asm3/paymentprocessor/cardcom.py
```diff
@@ -24,7 +24,7 @@ def getInvoiceItems(self, records):
             yield {"InvoiceLines%d.Quantity" % index: 1}
             yield {"InvoiceLines%d.Price" % index: price} 
 
-    def tokenCharge(self, payref, item_description=""):
+    def tokenCharge(self, payref, item_description = "", installments = 1):
```
bobintetley, line 27: Probably not a big deal as you're always going to pass installments, but your default type here is int, but you're passing str and it looks like it should be str for the JSON payload

bobintetley (approved):
Just a note on types, not a big deal I'll merge

PR #1033, opened 2021-07-21
Point to docker instructions in README

reutsharabani:
- add a reference to the docker scripts in the README
- add .idea/ to .gitignore

### DIFF
--- a/scripts/docker/README.md
+++ b/scripts/docker/README.md
@@ -13,11 +13,11 @@ Development environment for ASM3 with Docker. This stack creates two containers:
 ### Containers Up
 
 ```bash
-docker build -t asm3 ../../.
+docker-compose build
 docker-compose up -d
 ```
 
-Open [http://localhost:5000](http://localhost:5000) to view the running application
+Open [http://localhost:5000](http://localhost:8080) to view the running application
 
 ### Containers Down
 
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@ build
 node_modules
 tags
 .vscode
+.idea
 doc/manual/_build
 install/deb/sheltermanager3
 src/static/js/bundle
--- a/README.md
+++ b/README.md
@@ -71,6 +71,13 @@ In a development environment, you can start a test instance on port 5000 with:
 make test
 ```
 
+Docker Development Environment
+------------------------------
+
+You can run and test asm using a dockerized version.
+
+Find the resources at `asm3/scripts/docker/` and instructions at `asm3/scripts/docker/README.md`
+
 Logging
 -------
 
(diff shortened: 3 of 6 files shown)


SoulOfSet:
Were you able to actually get the docker container started and running? I remember when I was setting up my dev environment I had some issues. Not sure if everything in there is up-to-date. It could be that I was doing something incorrectly but iirc I just followed the instructions provided. I ended up just using Debian running under WSL on Windows.

reutsharabani:
@SoulOfSet I will try it today. I started implementing it and then I grep-ed the root directory and found it already exists. So I created the PR and didn't try it yet.

I'll update.

All I have accessible are two macs (one is M1) so I'll try and not be mac-specific if I do end up refreshing the files.

reutsharabani:
@SoulOfSet can you try the latest commit here? I also updated the README

SoulOfSet:
Sorry I didn't see the notification from being tagged. I can get the docker image to build now and it does come up but I'm curious on how you are using this? It looks like it's setting up Apache but the asm files from the package are never deployed to the www directory. Are you just using this as a dev environment using the make test command? The build time on this was also a little rough.  I think somewhere in the realm of 15-20 minutes? I just am not sure how this makes for a convenient to use development environment. Is that the goal, or are we trying to have a containerized deployable for production?

reutsharabani:
The startup time for me was minimal (a few seconds?) so that is interesting.

I added this since I don't have the tools to develop and test on my machine. But I do have docker facilities. So I use it as a possible development environment. If it works (I have no idea, I'm not a user) I can maybe work on other tickets using this way of development.

SoulOfSet:
To build the image or just do a docker-compose up? If you are able to access the web interface then I guess it's fine. I just couldn't get to it but maybe I'm not doing something right. In the future, it would probably be nice to get something setup to mount a volume with the code in it so docker compose really just sets up the environment and runs maybe the test target from the makefile. @bobintetley do you know if anyone is using docker for dev?

scripts/docker/README.md
```diff
@@ -13,11 +13,11 @@ Development environment for ASM3 with Docker. This stack creates two containers:
 ### Containers Up
 
 ```bash
-docker build -t asm3 ../../.
+docker-compose build
 docker-compose up -d
 ```
 
-Open [http://localhost:5000](http://localhost:5000) to view the running application
```
reutsharabani, line 20: change

reutsharabani:
For me when I run `docker-compose up` I can access the web interface. But if its not working for you its a problem.

Note that its mapping port 8080 to port 80 so its available on `localhost:80`.

So you should be able to access the web interface with `localhost:80`/`localhost`, or do something like `docker compose up -p 8080:1337` and then find the web interface on `localhost:1337`.

This is how I'm used to work with software across OSs. So I thought I'll give it a try.

PR #1123, opened 2022-04-18
Suppress warnings of optional CSV fields during import if not supplied

petervc:
This is my first ever pull request.  How many things did I get wrong?


bobintetley:
nothing, looks fine and corrects a huge oversight on our part. Thanks!

Issue #648, opened 2019-08-21
Allow find incident and find lost/found results to be configured


bobintetley:
The waiting list, find animal and find person screens can have their result columns configured, incidents and lost/found cannot and some users would like this.

PR #1125, opened 2022-04-21
Issue648

petervc:
Issue #648 - Allow find incident/lost/found results to be configured.  Minor column header changes.  Removed default sort column from lost/found results (postal code) as it used absolute column number.  Settings ==> Options ==> Find Animal/Person button renamed to include Incident configuration.  Lost and found configuration added to Settings ==> Options ==> Lost and Found.  Not sure if I18N will need to be addressed.


bobintetley:
great stuff, thank you!

PR #1127, opened 2022-04-27
Add CSV import support for declawed and special needs.

petervc:
Update docs to include new fields.


src/static/pages/manual/csvimportfields.html
```diff
@@ -271,6 +271,12 @@
 </dd>
 <dt>ANIMALMICROCHIPDATE</dt><dd><p>The date the microchip was implanted</p>
 </dd>
+</dd>
+<dt>ANIMALDECLAWED</dt><dd><p>Y/N to indicate yes/no</p>
+</dd>
+</dd>
```
petervc, line 279: Next commit removes extra </dd>

bobintetley: there's no need to edit any html files manually - it's generated from the markdown by sphinx

Issue #940, opened 2020-12-01
Shelter Events (fundraising etc)


bobintetley:
Requires a new table to store details of shelter events (open days, fund raising, etc)

ID
StartDateTime
EndDateTime
EventName
EventDescription (HTML)
(usual auditing fields)

This would be another heading under the ASM menu. Requires option to disable.

Events should be shown on the calendar. Future expansion might include using a HTML template and service call to generate a page of events

bobintetley:
Linking an event to a payment (ownerdonation) could also be useful

PR #1151, opened 2022-07-17
feature/sponsor-filtration

RomanSafovich:
* added new link to additional field
* added new types (sponsor, vet) to additional field
* can search for person that has the type of sponsor
* added 2 database tables and a field to existing database


src/static/js/header_edit_header.js
```diff
@@ -635,6 +635,9 @@ edit_header = {
         if (p.EXCLUDEFROMBULKEMAIL == 1) {
             flags.push(_("Exclude from bulk email"));
         }
+        if (p.ISSPONSOR == 1){
+            flags.push(_("Sponsor"));
+        }
         if (p.ADDITIONALFLAGS != null) {
             var stock = [ "aco", "adopter", "banned", "dangerous", "coordinator", "deceased", "donor", "driver", "excludefrombulkemail",
                 "fosterer", "homechecked", "homechecker", "member", "shelter", "retailer", "staff", "giftaid",
```
bobintetley, line 643: I think sponsor should be in the list of stock flags here.

src/asm3/dbupdate.py
```diff
@@ -242,6 +243,26 @@ def fclob(name, nullable = True):
...
+    sql += index("adoptionevent_StartDateTime", "adoptionevent", "StartDateTime")
+    sql += index("adoptionevent_EndDateTime", "adoptionevent", "EndDateTime")
+    sql += index("adoptionevent_EventName", "adoptionevent", "EventName")
+
+    sql += table("adoptioneventanimal", (
+        fid(),
+        fint("AdoptionEventID"),
+        fint("AnimalID"),
+        fint("WillReturnToFoster", True),
+        fdate("ArrivalDate")))
```
bobintetley, line 263: What is the purpose of WillReturnToFoster and ArrivalDate? The adoption event won't take the animal out of the care of the shelter, so presumably if a foster movement was already active you could infer whether it is going back to the fosterer? 

Is arrival date for multi-day events and tracking the animal arriving at the event?

src/asm3/dbupdate.py
```diff
@@ -242,6 +243,26 @@ def fclob(name, nullable = True):
     sql += index("adoption_ReturnedReasonID", "adoption", "ReturnedReasonID")
     sql += index("adoption_ReturnedByOwnerID", "adoption", "ReturnedByOwnerID")
     sql += index("adoption_TrialEndDate", "adoption", "TrialEndDate")
+    sql += index("adoption_AdoptionEventId", "adoption", "AdoptionEventID")
+
+    sql += table("adoptionevent", (
```
bobintetley, line 248: I think when we originally discussed this, I mentioned that adoption events sounded like a superset of #940 which was generic fundraising events.

I'd prefer more generic names - eg: just call the tables event and eventanimals

event could also have a type field so that adoption event could be set as the type (just a fixed list of integers no need for a lookup). Initially though I just add it as a column and not do anything with it since adoption events will be first.

This leaves it open for extending for other types of events and uses in future, such as the ability place an eventid  column on a payment (ownerdonation table) to track money received as part of an event, etc.

src/asm3/person.py
```diff
@@ -610,6 +611,7 @@ def get_person_find_advanced(dbo, criteria, username, includeStaff = False, incl
             elif flag == "giftaid": ss.ands.append("o.IsGiftAid=1")
             elif flag == "vet": ss.ands.append("o.IsVet=1")
             elif flag == "volunteer": ss.ands.append("o.IsVolunteer=1")
+            elif flag == "sponsor": ss.ands.append("o.IsSponsor=1")
```
bobintetley, line 614: minor thing but try and keep the order of the conditions alphabetical!

src/asm3/person.py
```diff
@@ -1013,7 +1017,8 @@ def bi(b):
         "IsRetailer":               retailer,
         "IsVet":                    vet,
         "IsGiftAid":                giftaid,
-        "AdditionalFlags":          flagstr
+        "AdditionalFlags":          flagstr,
+        "IsSponsor":                sponsor
```
bobintetley, line 1021: try and keep the Is* fields together in the list

bobintetley (changes requested):
Thanks, only a few minor bits really - just names and ordering. Nice job, thank you!

src/asm3/dbupdate.py
saggineumann: These fields are to be used to track the animals in the event - we saw that instead of a boolean you use you use a timestamp to track things (like given vaccination) so arrival date tracks which animal arrived at the event. WillReturnToFoster is used to help the event planner know whether the animal returns to same foster family when the event ends (and the animal wasn't adopted)

bobintetley: Ok, I understand ArrivalDate. Couldn't WillReturnToFoster be inferred from animal.ActiveMovementType=2 without needing the column at all? Other than the question over that I'm happy with everything else.

saggineumann: @bobintetley if I understand correctly, you suggest to check whether the ReturnDate is null or not on the active Movement (Type=2)? 

What we want to get to with this eventually is a method for the people in the field to easily manage the animals that are brought to the adoption event - we intend to show a list of the animals and allow single item and bulk edit in a similar fashion to the vaccinations list (per animal, or the general vaccinations list). One of the hardships they face is to figure out which animal goes back to its foster family or whether they have to find a new fosterer if the animal is not adopted. I suppose we can use the same mechanism and instead of using this column, set the ReturnDate on the active foster movement if the animal is not returning to its foster family, but in the case the user made a mistake, undoing the ReturnDate update might be harder to accomplish from the simple list UI.

What do you think?

bobintetley: I think that's fine. The user shouldn't need to handle returning the existing foster though - when you create a new foster or adoption movement, the backend code will automatically return any existing foster movement.

Ok, I'll do the merge and remove WillReturnToFoster.

saggineumann: @bobintetley so how would the user use the list of animals that show up on the event to know which ones need a new foster? The way they do it now is they print a list of animals, and use a pencil to mark which animals return to current foster and which ones don't.

bobintetley: If ActiveMovementType=2 then the animal has an active foster to return to. If ActiveMovementType=0 then the animal does not. 

Some users have added their own "Foster Needed" animal flag that they link an emblem to. It means the animal's name shows with a little icon to indicate that it is in need of a fosterer. They can also use the "Flag" mode of shelter view to identify animals needing a fosterer.

saggineumann: ActiveMovementType=0 indeed means that there's no active foster, but not all animals need to find a foster (e.g. they return to their original owner, or they stay at the shelter). So within the adoption event management, you think that we'd better use an animal flag to indicate which animal needs to find a foster?

bobintetley: Yes, since it will work irrespective of whether or not the animal is at an event.

PR #1160, opened 2022-08-01
feature/event-doctemplate-processor

RomanSafovich:
* added event tags to document template
* added sponsor and vet additional field to document template


bobintetley:
I don't see anything in this that wasn't in #1161, closing.

PR #1161, opened 2022-08-04
continue with document template

RomanSafovich:
* added functionality to document template that it will work with sponsor vet and person same as CURRENTOWNER. 

* added recursion with depth level 2 for the human additional field in case they want retrieve the data from the grandparent object


src/asm3/wordprocessor.py
```diff
@@ -153,6 +153,9 @@ def additional_field_tags(dbo, fields, prefix = ""):
             val = af["ANIMALNAME"]
         if af["FIELDTYPE"] in human_flags:
             val = af["OWNERNAME"]
+            human = asm3.additional.get_human_record(dbo, int(af["VALUE"]))[0]
```
bobintetley: Would prefer person_types to "human_flags" - they are additional field types, not flags!

As Saggi said, use asm3.person.get_person - no need to reinvent the wheel here and get_person resolves foreign keys for you

src/asm3/wordprocessor.py
```diff
@@ -153,6 +153,9 @@ def additional_field_tags(dbo, fields, prefix = ""):
...
+            for data in human:
+                tags[af["FIELDNAME"].upper() + data] = human[data]
```
bobintetley: This appears to be adding every single column in the owner row as a tag? This is unnecessary and will use a lot of memory.

Instead, add a function where you can pass a person record, tags collection and the prefix (eg: additional_field_person_tags(p, tags, prefix). This function then adds specific named fields to the tags from the person record, probably the same tags as we do for CURRENTOWNER, ORIGINALOWNER, etc.

.idea/asm3.iml
```diff
@@ -0,0 +1,12 @@
...
+  <component name="NewModuleRootManager">
+    <content url="file://$MODULE_DIR$" />
+    <orderEntry type="inheritedJdk" />
+    <orderEntry type="sourceFolder" forTests="false" />
+  </component>
+  <component name="PyDocumentationSettings">
+    <option name="format" value="PLAIN" />
+    <option name="myDocStringFormat" value="Plain" />
+  </component>
+</module>
```
bobintetley, line 12: did you accidentally add a load of stuff from your IDE?

src/asm3/wordprocessor.py
RomanSafovich: I did do that actually ... that loop was only for testing.
accidently wrong PR (even if I checked before)

bobintetley:
Sorry, this has gotten a bit messy and hard to follow. I've merged it and will tidy/commit in a few minutes.

PR #1186, opened 2022-10-24
event_new

RomanSafovich:
new page created /event_new


src/asm3/event.py
```diff
@@ -11,3 +19,39 @@ def get_event(dbo, eventid):
...
+        "StartDateTime": post.date("startdate"),
+        "EndDateTime": post.date("enddate"),
+        "EventName": post["eventname"],
+        "EventOwnerID": ownerid,
+        "EventAddress": post["address"],
+        "EventTown": post["town"],
+        "EventCounty": post["county"],
+        "EventPostCode": post["postcode"],
+        "EventCountry": post["country"]
+    }, user=username, generateID=False)
```
bobintetley, line 53: The new ID is the return value of dbo.insert - there's no need to specify the ID in the values list or call getid above, or set generateID=False.  Just do pid = dbo.insert("event", ...

src/asm3/event.py
```diff
@@ -11,3 +19,39 @@ def get_event(dbo, eventid):
...
+        raise asm3.utils.ASMValidationError(_("Event must have a start date.", l))
+
+    if post["enddate"].strip() == "":
+        raise asm3.utils.ASMValidationError(_("Event must have an end date.", l))
+
+    if post["address"].strip() == "":
+        raise asm3.utils.ASMValidationError(_("Event must have an address.", l))
+
+    if post.date("startdate") > post.date("enddate"):
+        raise asm3.utils.ASMValidationError(_("End date must be equal to or later than start date."))
```
bobintetley, line 40: Missed specifying the locale ,l

src/asm3/person.py
```diff
@@ -511,7 +511,7 @@ def get_investigation(dbo, personid, sort = ASCENDING):
         sql += "ORDER BY o.Date DESC"
     return dbo.query(sql, [personid])
 
-def get_person_find_simple(dbo, query, username="", classfilter="all", includeStaff = False, includeVolunteers = False, limit = 0, siteid = 0):
+def get_person_find_simple(dbo, query, username="", classfilter="all", includeStaff = False, includeVolunteers = False, limit = 0, siteid = 0, datatype="all"):
```
bobintetley, line 514: nothing really wrong with the implementation, but "datatype" is a bit of a confusing name. I would call it "typefilter"

src/static/js/header_additional.js
```diff
@@ -278,6 +278,22 @@ additional = {
         return valid;
     },
 
+    reset: function(){
+        //animal
+        $(".asm-animalchooser").animalchooser("clear");
+        if ($("#add_9.additional option[selected=selected]")[0])
```
bobintetley, line 284: what element is add_9 ? Can't see any changes to render method to add this. Are you referring to an additional field that might not be present in other databases?

src/asm3/person.py
bobintetley: would also rearrange the parameters so that it's next to classfilter in the list for readability, then do a quick search for other places that call get_person_find_simple to make sure they're unaffected

src/asm3/wordprocessor.py
```diff
@@ -1602,7 +1602,12 @@ def event_tags(dbo, e):
         "EVENTCREATEDBY":       e["CREATEDBY"],
         "EVENTCREATEDDATE":     python2display(l, e["CREATEDDATE"]),
         "EVENTLASTCHANGEDBY":   e["LASTCHANGEDBY"],
-        "EVENTLASTCHANGEDDATE": python2display(l, e["LASTCHANGEDDATE"])
+        "EVENTLASTCHANGEDDATE": python2display(l, e["LASTCHANGEDDATE"]),
+        "EVENTOWNERNAME":       e["EVENTOWNERNAME"],
+        "EVENTADDRESS":         e["EVENTADDRESS"],
+        "EVENTTOWN":            e["EVENTCOUNTY"],
+        "EVENTPOSTCODE":        e["EVENTPOSTCODE"],
+        "EVENTCOUNTRY":         e["EVENTCOUNTRY"]
```
bobintetley, line 1610: Not really an issue, but I would have the audit fields for created/lastchanged last in the list

src/static/js/event_new.js
```diff
@@ -0,0 +1,183 @@
...
+
+        render: function(){
+            return [
+                '<div id="dialog-similar" style="display: none" title="' + html.title(_("Similar Event")) + '">',
+                '<p><span class="ui-icon ui-icon-alert"></span>',
+                _("This event is very similar to another event on file, carry on creating this record?"),
+                '<br /><br />',
+                '<span class="similar-event"></span>',
+                '</p>',
+                '</div>',
```
bobintetley, line 17: Not sure that similar checks are really needed for events? I would have thought it unlikely that people would create duplicates.

src/static/js/event_new.js
```diff
@@ -0,0 +1,183 @@
...
+                }
+                finally{
+                    $("#asm-content button").button("enable");
+                }
+            };
+
+            // CONFIG
+            $("#countryrow").toggle( !config.bool("HideCountry") );
+            $("#statecounty").toggle( !config.bool("HideTownCounty") );
+
```
bobintetley, line 137: Country isn't enabled by default, and should be hidden by config option here

bobintetley (changes requested):
Thanks Roman, sorry for the delay getting to this.

src/static/js/event_new.js
saggineumann: @bobintetley it seems like the code does exactly that, no? the country is hidden by default (or by the config option setting)

src/static/js/event_new.js
saggineumann: @bobintetley we also think it's not likely for people to create dups and so haven't implemented it yet but just copied this part for consistency with the rest of the system, but we figured that it may be worth checking for dupes (by dates+location for instance) just to make it more friendly. in SOS Pets, there are weekly events, and if 2 people create the same event, it might cause a fuss

src/static/js/event_new.js
bobintetley: It does, clearly I'm blind!

saggineumann:
🤓

bobintetley:
see #940

PR #1207, opened 2022-12-18
event link in adoption movement type

RomanSafovich:
added event link check button in adoption movement type.
and also added it to move_adopt.

* when link to event is checked and movement date is chosen, it opens a dropdown with event dates within the next range: event start <= movement date <= event end.

* when movement date is changed, also the event dropdown will change to the correct range.


src/static/js/movements.js
```diff
@@ -932,6 +977,15 @@ $(function() {
             }
         },
 
+        event_dates: async function(){
+            let result = await common.ajax_post("movement", "mode=eventlink&movementdate=" + $("#movementdate").val());
+            let dates = jQuery.parseJSON(result);
+            $.each(dates, function(i, v){
+                $("#eventlinkdates").append("<option value='" + v.ID + "'>" + format.date(v.STARTDATETIME) + " - " + format.date(v.ENDDATETIME) +
+                " " + v.EVENTNAME + " " + v.EVENTADDRESS + ", " + v.EVENTTOWN + ", " + v.EVENTCOUNTY + ", " + v.EVENTCOUNTRY + "</option>");
```
bobintetley: Not seeing #eventlinkdates get cleared, so every time this runs it will append new dates?

src/static/js/movements.js
```diff
@@ -49,6 +49,8 @@ $(function() {
                     { json_field: "ISTRIAL", post_field: "trial", label: _("Trial Adoption"), tooltip: _("Is this a trial adoption?"), type: "check" },
                     { json_field: "TRIALENDDATE", post_field: "trialenddate", label: _("Trial ends on"), tooltip: _("The date the trial adoption is over"), type: "date" },
                     { json_field: "COMMENTS", post_field: "comments", label: _("Comments"), type: "textarea" },
+                    { json_field: "EVENTLINK", post_field: "eventlink", label: _("Link to event"), type: "check", hideif: function(){return !common.has_permission("lem");}},
```
bobintetley: hideif should also check a config switch to check whether events are enabled and have them off by default (add to configuration.py default map as EnableEvents: No). You can set EnableEvents to Yes in your configuration table manually while working on the functionality and we can merge it into master without anyone seeing it until you add the option to Settings->Options->Remove (if this makes sense)

src/static/js/move_adopt.js
```diff
@@ -80,6 +82,43 @@ $(function() {
...
+                    $("#event").closest("tr").fadeIn();
+                    move_adopt.event_dates();
+                }
+                else
+                    $("#event").closest("tr").fadeOut();
+
+            });
+            //callback when movementdate is changed
+            $("#movementdate").change(function(){
+                $("#event").empty();
```
bobintetley, line 109: Why not hide the event fields until movement date is completed?

saggineumann: @bobintetley we thought about this  but it would be odd - the page starts with a value (today) in the movement date field, but the value can be deleted (which would make the entire movement invalid), and it would be weird if the event would hide and show depending on whether there’s a movement date value. What do you think?

src/static/js/movements.js
RomanSafovich: I did clear the #eventlinkdates twice in movements.js (I've changed the id to #event). first time in that method and second time when #eventlink is unchecked (it will put 0 in the database).
I think this one was an old commit.

src/static/js/movements.js
RomanSafovich: I added this functionality, and fixed a bug that I found in my code, hope it works well :-)
I changed it to DisableEvents: Yes in the default map and also added it already to settings.

PR #1216, opened 2023-01-13
find event pages

saggineumann:
Find event form + results, part of the adoption event feature.


src/static/js/event_find_results.js
```diff
@@ -0,0 +1,214 @@
...
+                        if (v.FIELDTYPE == additional.YESNO) { 
+                            rv = v.VALUE == "1" ? _("Yes") : _("No");
+                        }
+                        else if (v.FIELDTYPE == additional.MONEY) {
+                            rv = format.currency(v.VALUE);
+                        }
+                        else if (v.FIELDTYPE == additional.ANIMAL_LOOKUP) {
+                            rv = '<a href="animal?id=' + v.VALUE + '">' + v.ANIMALNAME + '</a>';
+                        }
+                        else if ([additional.PERSON_LOOKUP, additional.VET, additional.SPONSOR].includes(v.FIELDTYPE)) {
```
saggineumann, line 189: @bobintetley you might want to change other *_results.js pages to use this instead of only additional.PERSON_LOOKUP

bobintetley: I think I'd add a function to additional to check if it's a person type rather than copying the same test everywhere. Makes it easy to add another in future then.

PR #1217, opened 2023-01-18
Feature/event page

saggineumann:
@RomanSafovich fyi

@bobintetley adding new event page which is not completed yet - we need to add the animals tab to it - coming up in the next couple of PRs


bobintetley:
Sorry, I've cocked up a bit here - I didn't commit the merge changes back to the branch first so it looks like it hasn't been merged, but it has!

PR #1228, opened 2023-02-11
Feature/event animals tab

saggineumann:
@bobintetley - I need your help with 2 issues I couldn't figure out:

1. When using the dropdown filter to change the table view, it seems like the filter is added to the route url, but this isn't used in the get request (unless I hit refresh) - what am I doing wrong there?
2. After adding an animal (or bulk add), I'd like to refresh the controller and reload the table - it would be easier than trying to build it in the UI as the query is more complex than just adding the animal and/or eventanimal to the table. How do I go about this? (or any other suggestions?)

### DIFF
--- a/src/static/js/event_animals.js
+++ b/src/static/js/event_animals.js
@@ -310,7 +310,7 @@ $(function(){
         },
 
         routes: {
-            "event_animals": function() { common.module_loadandstart("event_animals", "event_animals?id=" + this.qs.id); }
+            "event_animals": function() { common.module_loadandstart("event_animals", "event_animals?" + this.rawqs); }
         }
     };
 
--- a/src/static/js/event_animals.js
+++ b/src/static/js/event_animals.js
@@ -108,7 +108,7 @@ $(function(){
                             await tableform.show_okcancel_dialog("#dialog-addanimal", _("Add"), { notzero: [ "addanimal" ] });
                             let a = $("#addanimal").animalchooser("get_selected");
                             await common.ajax_post("event_animals", "mode=create&eventid=" + controller.event.ID + "&animalid=" + a.ID);
-                            //TODO: refresh list
+                            common.route_reload();
                         }
                     },
                     { id: "addanimals", text: _("Add from List"), icon: "litter", tooltip: _("Add animals to this event from animal list"), enabled: "always", perm: "cea",
@@ -117,7 +117,7 @@ $(function(){
                             await tableform.show_okcancel_dialog("#dialog-addbulkanimal", _("Add"), { notzero: [ "addanimals" ] });
                             let animals = $("#addanimals").val();
                             await common.ajax_post("event_animals", "mode=createbulk&eventid=" + controller.event.ID + "&animals=" + animals);
-                            //TODO: refresh list
+                            common.route_reload();
                         }
                     },
                     { id: "removeanimal", text: _("Remove"), icon: "delete", tooltip: _(""), enabled: "multi", perm: "cea", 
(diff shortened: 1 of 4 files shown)


src/static/js/event_animals.js
```diff
@@ -0,0 +1,320 @@
...
+                    },
+                    { id: "filter", type: "dropdownfilter", 
+                     options: [ "all|" + _("All"), "arrived|" + _("Arrived"), 
+                        "noshow|" + _("No show"), "neednewfoster|" + _("Need new foster"), 
+                        "dontneednewfoster|" + _("Don't need new foster"),
+                        "adopted|" + _("Adopted"), 
+                        "notadopted|" + _("Not adopted") ],
+                     click: function(selval) {
+                        common.route(controller.name + "?id=" + controller.event.ID + "&filter=" + selval);
+                        common.route_reload();
```
bobintetley, line 173: You don't need to do route_reload here, just setting a new route with the updated filter as your first line does is all you need. I think this is why your filter isn't working.

src/static/js/event_animals.js
```diff
@@ -0,0 +1,320 @@
...
+            };
+
+            const buttons = [
+                    { id: "addanimal", text: _("Add Animal"), icon: "animal-add", tooltip: _("Add animal to this event"), enabled: "always", perm: "cea",
+                        click: async function() { 
+                            $("#addanimal").animalchooser("clear");
+                            await tableform.show_okcancel_dialog("#dialog-addanimal", _("Add"), { notzero: [ "addanimal" ] });
+                            let a = $("#addanimal").animalchooser("get_selected");
+                            await common.ajax_post("event_animals", "mode=create&eventid=" + controller.event.ID + "&animalid=" + a.ID);
+                            //TODO: refresh list
```
bobintetley, line 111: This is where you should do route_reload() if you want to get the data from the backend again and reload the screen

src/static/js/event_animals.js
saggineumann: I tried without route_reload and it didn't work... adding route_reload didn't help either :)

src/static/js/event_animals.js
saggineumann: silly me, I found route_reload() but since it didn't work with the previous issue, I didn't try it here.

saggineumann: this did the trick!

src/static/js/event_animals.js
saggineumann: when I drop route_reload, I see a request being made with the following payload:
image


while the browser url shows as:
image

If I then hit refresh, the payload shows everything that's in the url param list and the filtered list is returned correctly

src/static/js/event_animals.js
```diff
@@ -0,0 +1,319 @@
...
+            var dates_range = "";
+            if(format.date(e.STARTDATETIME) == format.date(e.ENDDATETIME))
+                    dates_range = format.date(e.STARTDATETIME);
+                else
+                    dates_range = format.date(e.STARTDATETIME) + " - " + format.date(e.ENDDATETIME);
+            return dates_range + " " + e.EVENTNAME + " " + [e.EVENTADDRESS, e.EVENTTOWN, e.EVENTCOUNTY, e.EVENTCOUNTRY].filter(Boolean).join(", ");
+        },
+
+        routes: {
+            "event_animals": function() { common.module_loadandstart("event_animals", "event_animals?id=" + this.qs.id); }
```
bobintetley, line 313: I think this is your issue for offset not working - your only route for this module assumes a single id parameter. Change this to:

common.module_loadandstart("event_animals", "event_animals?" + this.rawqs);

You'll be able to pass any parameters then.

PR #1234, opened 2023-02-24
Feature/event menu

RomanSafovich:
adding event to asm menu


src/static/js/header.js
```diff
@@ -324,6 +324,12 @@ header = {
             if (asm.locale != "en_GB") {
                 $(".taggb").hide();
             }
+            if (config.bool("DisableEvents")){
+                $(".event").hide();
+            }
+            if (!common.has_permission("ae") && !common.has_permission("ve")){
+                $(".event").hide();
+            }
```
bobintetley, line 332: You don't need to do this - that's why the first element of menu_structure items is a permission.

src/asm3/html.py
```diff
@@ -559,6 +559,9 @@ def menu_structure(l, publisherlist, reports, mailmerges):
             ( asm3.users.ADD_LOG, "", "", "log_new?mode=person", "asm-icon-log", _("Add a log entry", l) ),
             ( asm3.users.VIEW_PERSON, "", "", "person_lookingfor", "asm-icon-animal-find", _("Person looking for report", l) ),
             ( asm3.users.VIEW_STAFF_ROTA, "", "tagrota", "staff_rota", "asm-icon-rota", _("Staff rota", l) ),
+            ("", "", "event", "--cat", "asm-icon-calendar", _("Event", l)),
+            (asm3.users.ADD_EVENT, "alt+shift+e", "addevent", "event_new", "asm-icon-event-add", _("Add event", l)),
+            (asm3.users.VIEW_EVENT, "", "viewevent", "event_find", "asm-icon-event-find", _("Find event", l)),
```
bobintetley, line 564: these 3 items should all have the same class so you can disable it in header.js and it should start with tag - eg: tagevent

PR #1235, opened 2023-02-26
Feature/heb translation

saggineumann:


bobintetley:
I can't really tell why they're conflicting, but weren't these changes to he.po and asm.pot in #1236 anyway so this one can be ignored?

PR #1274, opened 2023-05-04
Update quality_control.rep

kwinlr:
Updated check for email in wrong fields and corrected drivers license duplication check.

### DIFF
--- a/reports/quality_control.rep
+++ b/reports/quality_control.rep
@@ -1005,10 +1005,12 @@ UNION SELECT
  '<font color = "Red">Email Address in Wrong Field (Possible).  Alerts Only Once.</font>' AS issue
 FROM owner o
 WHERE date_trunc('day', o.LASTCHANGEDDATE) = current_date -1
-AND (o.hometelephone LIKE '%@%' 
-OR o.mobiletelephone LIKE '%@%'
-OR o.worktelephone LIKE '%@%' 
-OR o.mobiletelephone2 LIKE '%@%')
+  AND (o.hometelephone LIKE '%@%' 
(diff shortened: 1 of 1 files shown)


bobintetley:
Could I make a few suggestions Karen?

1. Instead of having lots of CUSTOMIZE HERE comments, you could use SM's CONST token instead to define variables at the top that the user can set so they don't have to edit the actual code further down. For example, at the top you could put:

```
$CONST PT_ADOPTIONFEE=2$ -- Adoption fee payment type ID from donationtype table
$CONST SP_DOG=1$ -- Dog species ID
$CONST SP_CAT=2$ -- Cat species ID
$CONST AT_NONSHELTER=40$ -- Non-shelter animaltype ID
... etc
```

Then later in your code when you want to use the values, just put the token eg: $PT_ADOPTIONFEE$

2. It's a little thing, but changelogs are usually in reverse date order so that you can see the latest change first.

3. We version all reports for auto-updates. By asking people to modify the report as it is installed, they will never get any of your improvements. Instead, ask people to clone their own version of the report after installing - this is what we suggest for users customising any reports so that report updates don't remove customisations.

Thanks!

kwinlr:
That is a great suggestion.  I'll work on that.   If if I add comments to
the top that they must CLONE before making customization, that should
prevent people from getting all their changes overwritten.

Hey, while you are looking at it... would you mind looking at the very last
SQL that checks for duplicate idnumbers?   There has to be a better way to
write that...   Could you suggest an alternative?

bobintetley:
it's a little bit clunky, but it's not a big query and if it works I wouldn't worry about it :)

PR #1294, opened 2023-06-03
Add the ability to collect payments within the move_adopt process (co…

saggineumann:
…nfigurable)

### DIFF
--- a/src/asm3/configuration.py
+++ b/src/asm3/configuration.py
@@ -253,6 +253,7 @@ DEFAULTS = {
     "MovementNumberOverride": "No",
     "MovementPersonOnlyReserves": "Yes",
     "MultiSiteEnabled": "No", 
+    "MoveAdoptDonationsEnabled": "No",
     "JSWindowPrint": "Yes",
     "OnlineFormSpamHoneyTrap": "Yes",
     "OnlineFormSpamUACheck": "No",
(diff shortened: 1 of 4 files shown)


bobintetley:
Ok, I think I get it, it takes you to a version of the payments screen for the movement you just added instead of showing the list of document templates to generate? What's the use case for doing this?

saggineumann:
Exactly and then you can proceed to the docs. It simplifies the adoption
flow, as there’s no way to charge the customer in the non-self-serve flow
other than to switch to the animal/owner page after getting to move_docgen
(correct me if I'm wrong)

bobintetley:
There's a payments widget on move_adopt for adding the payments? Is this so that you can request payment by email or something?

saggineumann:
Exactly. To accept the payments. I thought of doing it in move_adopt first,
but I couldn’t figure out how to do it in the payment widget because the
payments don’t exist yet. Do you have a better way to do it? I should’ve
consulted with you first but I was on  a roll this weekend 😅

bobintetley:
Ok, I understand, and you can't have a tickbox for requesting payment on move->adopt because you have all that crazy popup stuff for CardCom. I'll merge for now but might have a further think/mess with it a bit before it goes into production.

bobintetley:
I take it back, this is actually quite a good/elegant solution for the problem. Made a few minor tweaks but pretty much left as is and added a config switch to Options->Movements. It will go into production with Monday's update. Thanks!

saggineumann:
Thank you 🥹. I thought it would work the same regardless of payment
gateway (email dialog, cardcom popup etc). My original idea was to
optionally break the payments completely to a payments page, but then
figured that the only method to create a multi item payment is to use the
move_adopt payments.

Issue #1536, opened 2024-09-09
Quick Reports option


bobintetley:
Like quicklinks, this is a user setting that lets the user choose from a multi-select dropdown of their favourite reports.

This will add a new button/dropdown menu at the beginning of the quicklinks line, labeled "Reports" with the favourited reports in, in the order they were chosen.

PR #1593, opened 2024-12-18
Quick reports #1536

adamdspencer77:
Seem to have got the required behaviour about sorted. Ready to discuss work done so far before moving on.


src/main.py
```diff
@@ -2716,6 +2717,15 @@ def post_all(self, o):
         email = post["email"]
         signature = post["signature"]
         quicklinks = post["quicklinksid"]
+        rawfavouritereports = post["favouritereportsid"]
```
bobintetley, line 2720: minor thing, but the name of the feature was quickreports rather than favourite reports, even though they equate to the same thing. I would call this variable just quickreports

src/main.py
```diff
@@ -2716,6 +2717,15 @@ def post_all(self, o):
...
+        favouritereports = ""
```
bobintetley, line 2721: This is for the string that we're going to store in the config table, I'd called it quickreportscfg. I would make it a list instead, that you append to in your loop so we can use join to add the commas. quickreportscfg = []

src/main.py
```diff
@@ -2716,6 +2717,15 @@ def post_all(self, o):
...
+        for reportid in rawfavouritereports.split(","):
+            for report in asm3.reports.get_reports(o.dbo):
```
bobintetley, line 2723: this is going to open a database connection and execute the get_reports query for every single iteration of the outer loop - call it once outside the loop and assign it to a variable to iterate here instead so you aren't re-reading it from the database every time

src/main.py
```diff
@@ -2716,6 +2717,15 @@ def post_all(self, o):
...
+                if report.ID == int(reportid):
+                    reporttitle = report.TITLE
+            if favouritereports != "":
+                favouritereports += ","
```
bobintetley, line 2727: this line and the if statement before it can be removed, and replaced later with ",".join(quickreportscfg) to stitch them together

src/main.py
```diff
@@ -2716,6 +2717,15 @@ def post_all(self, o):
...
+            favouritereports += reportid + "=" + reporttitle
```
bobintetley, line 2728: concatenation! do something like quickreportscfg.append( f"{reportid}={reporttitle}" ) instead

src/main.py
```diff
@@ -2728,6 +2738,7 @@ def post_all(self, o):
             asm3.configuration.cset(o.dbo, "%s_QuicklinksID" % o.user, "")
         else:
             asm3.configuration.cset(o.dbo, "%s_QuicklinksID" % o.user, quicklinks)
+        asm3.configuration.cset(o.dbo, "%s_FavouriteReportsID" % o.user, favouritereports)
```
bobintetley, line 2741: here's where you do ",".join(quickreportscfg)

src/static/js/change_user_settings.js
```diff
@@ -190,6 +199,13 @@ $(function() {
                 $("#quicklinksid").find("option[value='" + common.trim(v + "']")).attr("selected", "selected");
             });
             $("#quicklinksid").change();
+
+            let fr = config.str(asm.user + "_FavouriteReportsID").split(",");
+            $.each(fr, function(i, v) {
+                $("#favouritereportsid").find("option[value='" + common.trim(v + "']")).attr("selected", "selected");
```
bobintetley, line 205: isn't v here in ID=Title format? So you'll need to throw away the =Title part when finding the value to select

src/static/js/change_user_settings.js
```diff
@@ -79,6 +87,7 @@ $(function() {
                         options: '<option value="" data-style="background-image: url(static/images/flags/' + config.str("Locale") + '.png)">' + _("(use system)") + '</option>' + 
                             this.two_pair_options(controller.locales, true) },
                     { post_field: "quicklinksid", label: _("Quicklinks"), type: "selectmulti", options: change_user_settings.quicklink_options() },
+                    { post_field: "favouritereportsid", label: _("Favourite Reports"), type: "selectmulti", options: { displayfield: "TITLE", rows: controller.reports} },
```
bobintetley, line 90: again, really minor thing, prefer the name quickreportsid and Quick Reports!

src/static/js/header.js
```diff
@@ -353,6 +353,19 @@ header = {
...
+        if (favouritereports) {
+            console.log(favouritereports);
+            s += "<a><span class='asm-icon asm-icon-report'></span> <select>";
+            for ( let a = 0; a < favouritereports.length; a++ ) {
+                //get_field(rows, id, field)
+                s += "<option value='" + favouritereports[a].split("=")[0] + "'>" + favouritereports[a].split("=")[1] + "</option>";
+                //s += "href='report_criteria?id=" + favouritereports[0] + "&target=report'>";
+            }
+            s += "</select></a>";
+        }
```
bobintetley, line 367: Rather than using a select, use an asmmenu widget from common_widgets.js - the HTML will be something like:

```
<div id="asm-menu-quickreports" class="asm-menu-icon">Quick Reports</div>
<div id="asm-menu-quickreports-body" class="asm-menu-body">

<ul class="asm-menu-list">
<li class="asm-menu-item"><a href="report?id=1">Report 1</a></li>
</ul>

</div>
```

This is what header.js uses to make the menus at the top.

src/main.py
adamdspencer77: :) I took that approach originally but couldn't get the join to work and was keen to move on.

adamdspencer77: I renamed quickreportscfg and thought while I was renaming fields I'd change all the variable names I'd created that included Favourite to Quick. I found some other bugs (possibly ones you've put further down this list) that I don't think were related to me renaming the variables. Anyway, sorted out the bugs, renamed the fields and made an array and got the join working, I think I was confusing javascript joins with python joins :)

bobintetley: Yeah, they're the opposite way round. Javascript join is a method of arrays, so you'd do [ ].join(",") - but Python join is a method of String, so you do ",".join([])

src/main.py
adamdspencer77: Sorted

src/main.py
adamdspencer77: I did see this one coming :)

src/main.py
adamdspencer77: Already done :)

src/static/js/change_user_settings.js
adamdspencer77: Already sorted.

src/static/js/change_user_settings.js
adamdspencer77: Already done.

Issue #1544, opened 2024-09-28
Invoice number on costs


bobintetley:
A few people have mentioned something like this. They'd like costs to have an associated person (ie. vet or other org) and invoice number so that it's easy to tell whether they've been billed for something and what invoice it was associated with. The costpaid field also exists already for the shelter to indicate when they've paid it.

animalcost.OwnerID INTEGER
animalcost.InvoiceNumber VARCHAR

bobintetley:
The code that anonymises person records should consider cost records within the period linking to it too.

PR #1595, opened 2024-12-20
Add invoice number and payee to costs #1544

adamdspencer77:
Invoice numbers and person records can be associated with animal costs.


src/asm3/dbupdate.py
```diff
@@ -6312,3 +6312,8 @@ def update_34902(dbo: Database) -> None:
         dbo.ddl_add_table_column("IsRetired", dbo.type_integer, True)
     ])
     dbo.execute_dbupdate( dbo.ddl_add_table("lkmediaflags", fields) )
+
+def update_34903(dbo: Database) -> None:
+    # Add extra fields to facilitate invoice tracking to animalcost table
+    add_column(dbo, "animalcost", "OwnerID", dbo.type_integer)
+    add_column(dbo, "animalcost", "InvoiceNumber", dbo.type_shorttext)
```
bobintetley, line 6319: Need to add indexes for both these columns. Also, as well as writing a database update function, you need to add these columns to the definition of the animalcost table above in the sql_structure function (and the indexes again).

This is because there are two paths, one for setting up a brand new database, and one for updating existing databases.

Issue #1372, opened 2023-11-30
Extra good with fields


bobintetley:
Some more "Good with" fields that would make sense.

IsCrateTrained
IsGoodWithElderly
IsGoodTraveller
IsGoodOnLead

EnergyLevel 1-5

This makes sense in that people might be looking for a less energetic animal (although in many cases could be inferred by age)

bobintetley:
There is an option to hide the "good with" fields. Also, good with fields do tie into websites via adopable_js, so that needs checking.

bobintetley:
There is an option to hide the "good with" fields. Also, good with fields do tie into websites via adopable_js, so that needs checking.

PR #1599, opened 2025-01-02
 Extra good with fields #1372 

adamdspencer77:
Added IsCrateTrained, IsGoodWithElderly, IsGoodTraveller, IsGoodOnLead and EnergyLevel 1-5 to animal notes. Publisher and wordprocessor updated accordingly.


bobintetley:
Pull request is the wrong way round, changes should be merged from ft-1372 into master

PR #1600, opened 2025-01-02
Additional "good with" fields #1372

adamdspencer77:
Added IsCrateTrained, IsGoodWithElderly, IsGoodTraveller, IsGoodOnLead, EnergyLevel (1-5) to animal notes. Updated adoptable_js filters for new fields. Added wordkeys to animal document generator.


src/asm3/animal.py
```diff
@@ -2934,7 +2949,12 @@ def insert_animal_from_form(dbo: Database, post: PostedData, username: str) -> i
         "Archived":         0,
         "ActiveMovementID": 0,
         "HasActiveReserve": 0,
-        "MostRecentEntryDate": dbo.today()
+        "MostRecentEntryDate": dbo.today(),
+        "IsCrateTrained": cratetrained,
+        "IsGoodWithElderly":   goodwithelderly,
+        "IsGoodTraveller": goodtraveller,
+        "IsGoodOnLead": goodonlead,
+        "EnergyLevel": energylevel
```
bobintetley, line 2957: little thing, but I would try and put these with the other good with fields in the list in both insert_animal_from_form and update_animal_from_form

src/asm3/dbupdate.py
```diff
@@ -369,7 +370,12 @@ def fclob(name: str, nullable: bool = True) -> str:
         fint("TotalDaysOnShelter", True),
         fstr("AgeGroupActiveMovement", True),
         fint("DailyBoardingCost", True),
-        fstr("AnimalAge", True) ))
+        fstr("AnimalAge", True),
+        fint("IsCrateTrained"),
+        fint("IsGoodWithElderly"),
+        fint("IsGoodTraveller"),
+        fint("IsGoodOnLead"),
+        fint("EnergyLevel") ))
```
bobintetley, line 378: similarly, put these near the other good with fields in the list rather than at the end, also they should have the True argument to indicate they're nullable

src/asm3/dbupdate.py
```diff
@@ -6323,3 +6329,11 @@ def update_34903(dbo: Database) -> None:
     add_index(dbo, "animalcost_OwnerID", "animalcost", "OwnerID")
     add_column(dbo, "animalcost", "InvoiceNumber", dbo.type_shorttext)
     add_index(dbo, "animalcost_InvoiceNumber", "animalcost", "InvoiceNumber")
+
+def update_34904(dbo: Database) -> None:
+    # Add extra fields to facilitate invoice tracking to animalcost table
```
bobintetley, line 6334: copied comment from previous dbupdate

src/media/internet/animalviewadoptable/body.html
```diff
@@ -1,6 +1,6 @@
 <div id="asm3-adoptables" />
 <script>
-asm3_adoptable_filters = "sex breed agegroup size species";
+asm3_adoptable_filters = "sex breed agegroup size species goodwith";
```
bobintetley, line 3: this is making the goodwith filter on by default for all new databases, on reflection, this is probably ok

bobintetley:
Just a couple of very minor things, nice work :)

src/asm3/animal.py
adamdspencer77: Done. Noticed update_animals_from_form, updating it...

adamdspencer77: Updated code but will need to test.

src/asm3/dbupdate.py
adamdspencer77: Done. I didn't include the True argument because the existing IsGoodWith's don't have them. Should I add it to them too?

src/asm3/animal.py
adamdspencer77: Removed changes to update_animals_from_form

Issue #1142, opened 2022-06-19
Media flags


bobintetley:
Some users would like to be able to add their own classification/type to media records (eg: "Adoption packet", or "Driver's licence"). At the moment, there is the free type media notes field, but it would make reporting on different types of media easier.

Rather than it being a single dropdown, I think we should add a media.MediaFlags column and have an lkmediaflags table like the other flags. This means that multiple flags can then be assigned to a piece of media. A multiselect dropdown filter on the media tab allows you to select only media that has all the chosen flags for filtering.

I don't think we need any defaults either as everything we might make as a default for reporting we can infer from other fields, or the proposed MediaSource column #1295

bobintetley:
This would allow the media tab to filter by type with a toolbar dropdown so that people could much more easily find what they are looking for.

bobintetley:
It should include some defaults, such as Photo, Application Form, Adoption Paperwork, Medical Paperwork. The default is to leave this field blank/0 on existing media records. Photo could probably be chosen for all jpgs.

bobintetley:
Makes sense to do this at the same time as #1295

Tetzor:
Would it be possible to also add an option to select a default flag filter.

PR #1605, opened 2025-01-06
Add media flags #1142

adamdspencer77:
Media may now be flagged using user defined flags.

### DIFF
--- a/src/static/js/media.js
+++ b/src/static/js/media.js
@@ -122,7 +122,7 @@ $(function() {
                         $.each(m.MEDIAFLAGS.split("|"), function(count, flag) {
                             $.each(controller.flags, function(count, flagdata) {
                                 if (flag == flagdata.ID) {
-                                    h.push("<span class=asm-media-flag>" + flagdata.FLAG + "</span>");
+                                    h.push("<span class=asm-media-flag>" + flag + "</span>");
                                 }
                             });
                             
@@ -1216,18 +1216,15 @@ $(function() {
             $("#mediaflagsfilter").change(function() {
                 let flagfilters = $("#mediaflagsfilter").val();
                 if (flagfilters.length > 0) {
-                    var newmediarows = [];
+                    let newmediarows = [];
                     $.each(controller.media, function(mediacount, media) {
-                        var include = true;
-                        console.log("media.MEDIANOTES = " + media.MEDIANOTES);
-                        //console.log('media.MEDIAFLAGS' + " = " + media.MEDIAFLAGS);
-                        //console.log('media.MEDIAFLAGS.split("|")' + " = " + media.MEDIAFLAGS.split("|"));
-                        $.each(flagfilters, function(filtercount, flag) {
-                            console.log("flag = " + flag);
-                            var flagfound = false;
-                            var mediaflags = media.MEDIAFLAGS.split("|");
-                                $.each(mediaflags, function(flagcount, filter) {
-                                    if ( flagcount < mediaflags.length - 1) {
+                        let include = true;
+                        $.each(flagfilters, function(filtercount, filter) {
+                            let flagfound = false;
+                            let mediaflags = media.MEDIAFLAGS.split("|");
+                                $.each(mediaflags, function(flagcount, flag) {
+                                    if ( flagcount < mediaflags.length) {
+                                        if (!flag) {return false;}
                                         if (flag == filter) {
                                             flagfound = true;
                                             return false;
--- a/src/static/js/media.js
+++ b/src/static/js/media.js
@@ -29,11 +29,12 @@ $(function() {
                 close_on_ok: true,
                 columns: 1,
                 width: 550,
+                focusfirst: false,
                 fields: [
-                    { json_field: "MEDIANOTES", post_field: "medianotes", label: _("Notes"), type: "textarea" },
                     { json_field: "RETAINUNTIL", post_field: "retainuntil", label: _("Retain Until"), type: "date",
                         callout: _("Automatically remove this media item on this date") },
-                    { json_field: "MEDIAFLAGS", post_field: "mediaflags", label: _("Flags"), type: "selectmulti" }
+                    { json_field: "MEDIAFLAGS", post_field: "mediaflags", label: _("Flags"), type: "selectmulti" },
+                    { json_field: "MEDIANOTES", post_field: "medianotes", label: _("Notes"), type: "textarea" }
                 ]
             };
 
@@ -42,10 +43,8 @@ $(function() {
                 idcolumn: "ID",
                 truncatelink: 50, // Only use first 50 chars of MEDIANOTES for edit link
                 edit: async function(row) {
-                    console.log(dialog);
                     tableform.fields_populate_from_json(dialog.fields, row);
                     html.media_flag_options(controller.flags, $("#mediaflags"));
-                    //$("#mediaflags").html("<option>Arse</option><option>Boobies</option>");
                     await tableform.dialog_show_edit(dialog, row);
                     tableform.fields_update_row(dialog.fields, row);
                     await tableform.fields_post(dialog.fields, "mode=update&mediaid=" + row.ID, "media");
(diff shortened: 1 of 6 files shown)


src/static/js/media.js
```diff
@@ -42,12 +44,14 @@ $(function() {
                 truncatelink: 50, // Only use first 50 chars of MEDIANOTES for edit link
                 edit: async function(row) {
                     tableform.fields_populate_from_json(dialog.fields, row);
+                    html.media_flag_options(controller.flags, $("#mediaflags"));
```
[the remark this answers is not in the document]
adamdspencer77: Moved to sync.

src/static/js/media.js
```diff
@@ -1181,7 +1213,37 @@ $(function() {
...
                 return false;
             });
-
+            
+            $("#mediaflagsfilter").change(function() {
+                let flagfilters = $("#mediaflagsfilter").val();
+                if (flagfilters.length > 0) {
+                    let mediarows = controller.media;
+                    var newmediarows = [];
+                    $.each(mediarows, function(mediacount, media) {
```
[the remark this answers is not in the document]
adamdspencer77: Done.

src/static/js/media.js
```diff
@@ -135,6 +153,16 @@ $(function() {
                         h.push(tableform.table_render_edit_link(m.ID, format.date(m.DATE)));
                         h.push("<br/>");
                         h.push(media.render_mods(m));
+                        h.push("<br/>");
+                        $.each(m.MEDIAFLAGS.split(","), function(count, flag) {
```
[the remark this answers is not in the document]
adamdspencer77: Hopefully sorted. Broke the table filter which I ended up rewriting a lot of the code for.

src/static/js/media.js
```diff
@@ -113,6 +116,20 @@ $(function() {
...
                         return h.join("");
                     }},
+
+                    { field: "MEDIAFLAGS", classes: "mode-table", display: _("Flags"), formatter: function(m) {
+                        let h = [];
+                        $.each(m.MEDIAFLAGS.split("|"), function(count, flag) {
+                            $.each(controller.flags, function(count, flagdata) {
+                                if (flag == flagdata.ID) {
+                                    h.push("<span class=asm-media-flag>" + flagdata.FLAG + "</span>");
+                                }
```
bobintetley, line 126: I don't think this needs the inner iteration of controller.flags - you can do the h.push line with flag instead flagdata.FLAG. 

```
$.each(m.MEDIAFLAGS.split("|"), function(count, flag) {
    if (flag) { h.push("<span class=asm-media-flag>" + flag + "</span>"); }
});
```

It also means that flags will still appear on media items even if they were deleted from the lkmediaflag table. This would be consistent with the behaviour of the animal/person flags, which have code that deletes them from animals and people when they are removed from the lookup data screen.

src/static/js/media.js
```diff
@@ -135,6 +152,16 @@ $(function() {
                         h.push(tableform.table_render_edit_link(m.ID, format.date(m.DATE)));
                         h.push("<br/>");
                         h.push(media.render_mods(m));
+                        h.push("<br/>");
+                        $.each(m.MEDIAFLAGS.split("|"), function(count, flag) {
+                            $.each(controller.flags, function(count, flagdata) {
+                                if (flag == flagdata.ID) {
+                                    h.push("<span class='asm-media-flag asm-media-flag-thumb'>" + flagdata.FLAG + "</span>");
+                                }
```
bobintetley, line 160: same again, you don't need the inner iteration.

src/static/js/media.js
```diff
@@ -1181,7 +1212,46 @@ $(function() {
...
+                        var include = true;
+                        console.log("media.MEDIANOTES = " + media.MEDIANOTES);
+                        //console.log('media.MEDIAFLAGS' + " = " + media.MEDIAFLAGS);
+                        //console.log('media.MEDIAFLAGS.split("|")' + " = " + media.MEDIAFLAGS.split("|"));
+                        $.each(flagfilters, function(filtercount, flag) {
+                            console.log("flag = " + flag);
+                            var flagfound = false;
+                            var mediaflags = media.MEDIAFLAGS.split("|");
+                                $.each(mediaflags, function(flagcount, filter) {
+                                    if ( flagcount < mediaflags.length - 1) {
```
bobintetley, line 1230: I presume this is to ignore the blank value after the trailing pipe. I think it would be more readable and simpler to just do nothing if the flag is an empty string. ```if (!filter) { return; } ``` instead would short circuit the function and take out a level of nested ifs.

This is just for info though, because post-merge we can replace the whole lot with a call to array_overlap_all :

```
let newmediarows = [];
$.each(controller.media, function(mediacount, media) {
    if (common.array_overlap_all(flagfilters, media.MEDIAFLAGS.split("|"))) {
        newmediarows.push(media);
    }
});
```

src/static/js/media.js
```diff
@@ -1181,7 +1212,46 @@ $(function() {
                 $("#dialog-copyperson").dialog("open");
                 return false;
             });
-
+            
+            $("#mediaflagsfilter").change(function() {
+                let flagfilters = $("#mediaflagsfilter").val();
+                if (flagfilters.length > 0) {
+                    var newmediarows = [];
```
bobintetley, line 1219: try to avoid using var, it can have unexpected behaviours in nested functions due to how var gets "hoisted" to the outer functions - always use let unless you explicitly need the behaviour of var or you're writing code for ancient versions of Internet Explorer. https://www.w3schools.com/js/js_hoisting.asp

bobintetley (commented):
Few more comments. Sorry if this seems a bit picky, they're very minor things, your code for this one is a good standard.

src/static/js/media.js
adamdspencer77: Done.

src/static/js/media.js
adamdspencer77: Righto. I know that this is just for info but I thought I'd implement it to help me understand it. Is `return;` handled differently that `return false;` or is the false optional?

bobintetley: ```return;``` is the same as ```return undefined;``` 

JQuery's each will stop iterating if gets a return value of false, but it will keep iterating if it gets undefined.

adamdspencer77: OK, I think I've sorted it. I swapped a couple of variable names around too that were a little misleading. I included the false as the return value, I assume that this will then skip over the next if statement but it would still still run through that if I left it as undefined?

adamdspencer77: `$("#mediaflagsfilter").change(function() {
                let flagfilters = $("#mediaflagsfilter").val();
                if (flagfilters.length > 0) {
                    var newmediarows = [];
                    $.each(controller.media, function(mediacount, media) {
                        var include = true;
                        $.each(flagfilters, function(filtercount, filter) {
                            var flagfound = false;
                            var mediaflags = media.MEDIAFLAGS.split("|");
                                $.each(mediaflags, function(flagcount, flag) {
                                    if ( flagcount < mediaflags.length) {
                                        if (!flag) {return false;}
                                        if (flag == filter) {
                                            flagfound = true;
                                            return false;
                                        }
                                    }
                                });
                                if (!flagfound) {
                                    include = false;
                                }
                        });
                        if (include) {
                            newmediarows.push(media);
                        }
                    });
                    media.table.rows = newmediarows;
                } else {
                    media.table.rows = controller.media;
                }
                tableform.table_update(media.table);
                if (media.icon_mode_active) {
                    media.mode_icon();
                }
                return false;
            });`

adamdspencer77: Tried to paste in the code here but it looked terrible. I'll send it on Slack.

src/static/js/media.js
adamdspencer77: Ah, I thought I needed to use var if a nested function needed access, I didn't think that a let could be accessed from within the $.each but I see that's not the case :) I've changed my vars to lets and all seems well :)

Issue #1591, opened 2024-12-17
Import "good with" fields for existing animals


bobintetley:
Request from a user who is writing their own interface and using the service API CSV import mechanism to pass data. They want to be able to update the "good with" fields via CSV import. At the moment they are only imported on new animals, but ignored for existing.

bobintetley:
The fields that can be imported for existing animals are documented in the manual, which will need updating as a result of this change.

Issue #1285, opened 2023-05-24
csvimport: Time information


bobintetley:
Some of the date fields that we import should be able to include time info, such as LOGDATE. Currently, they don't. 

Look at which is best/easiest - whether to have a TIME field like LOGTIME and a new gkt method in csvimport.py, or to have gkd recognise time info included with the dates.

Export Animals as CSV should include time info when exporting for these fields.

bobintetley:
separate LOGTIME field is better because csvimport feeds what looks like HTTP POST data to the insert functions and they expect a separate time field.

bobintetley:
Fields requiring time include:

animal:

ANIMALENTRYTIME

incidents:

INCIDENTDATE/INCIDENTTIME
INCIDENTCOMPLETEDDATE/INCIDENTCOMPLETEDTIME
DISPATCHDATE/DISPATCHTIME

log:

LOGDATE/LOGTIME

PR #1613, opened 2025-01-14
CSV time import and new goodwith fields #1285 #1591

adamdspencer77:
Added
- ANIMALENTRYTIME
- INCIDENTTIME
- INCIDENTCOMPLETEDTIME
- DISPATCHTIME
- LOGTIME
as CSV import options (#1285).

Also all "Good with" animal fields can be imported via CSV for both new and existing animals (#1591).


src/asm3/csvimport.py
```diff
@@ -29,8 +29,8 @@
     "ANIMALCOMMENTS", "ANIMALDESCRIPTION", "ANIMALMARKINGS", "ANIMALNEUTERED", "ANIMALNEUTEREDDATE", "ANIMALMICROCHIP", "ANIMALMICROCHIPDATE", 
     "ANIMALENTRYDATE", "ANIMALENTRYCATEGORY", "ANIMALENTRYTYPE", "ANIMALFLAGS", "ANIMALWARNING",
     "ANIMALREASONFORENTRY", "ANIMALHIDDENDETAILS", "ANIMALNOTFORADOPTION", "ANIMALNONSHELTER", "ANIMALTRANSFER",
-    "ANIMALGOODWITHCATS", "ANIMALGOODWITHDOGS", "ANIMALGOODWITHKIDS", 
-    "ANIMALHOUSETRAINED", "ANIMALHEALTHPROBLEMS", "ANIMALIMAGE",
+    "ANIMALGOODWITHCATS", "ANIMALGOODWITHDOGS", "ANIMALGOODWITHKIDS", "ANIMALGOODWITHELDERLY", "ANIMALGOODONLEAD",
+    "ANIMALHOUSETRAINED", "ANIMALCRATETRAINED", "ANIMALENERGYLEVEL", "ANIMALHEALTHPROBLEMS", "ANIMALIMAGE",
     "COSTDATE", "COSTTYPE", "COSTAMOUNT", "COSTDESCRIPTION",
     "VACCINATIONTYPE", "VACCINATIONDUEDATE", "VACCINATIONGIVENDATE", "VACCINATIONEXPIRESDATE", "VACCINATIONRABIESTAG",
```
bobintetley, line 35: I think ANIMALENTRYTIME, LOGTIME, INCIDENTTIME etc need adding to this list of valid import fields

src/asm3/animal.py
```diff
@@ -4135,6 +4135,16 @@ def merge_bool(dictfield, fieldname):
...
     merge("pickupaddress", "PICKUPADDRESS")
+    merge("housetrained", "ISHOUSETRAINED")
+    merge("cratetrained", "ISCRATETRAINED")
+    merge("goodwithcats", "ISGOODWITHCATS")
+    merge("goodwithdogs", "ISGOODWITHDOGS")
+    merge("goodwithkids", "ISGOODWITHCHILDREN")
+    merge("goodwithelderly", "ISGOODWITHELDERLY")
+    merge("energylevel", "ENERGYLEVEL")
+    merge("goodonlead", "ISGOODONLEAD")
+    merge("goodtraveller", "ISGOODTRAVELLER")
```
bobintetley, line 4146: All of these fields are int - they should be calling merge_int rather than merge. This is one of those times where SQLite will be forgiving because all of its types are strings behind the scenes, but it will fail on Postgres or MySQL.

src/asm3/csvimport.py
adamdspencer77: Done.

src/asm3/animal.py
adamdspencer77: Yeah, I started with merge_int but it does odd things when passed a 0.

src/asm3/csvimport.py
```diff
@@ -461,16 +461,22 @@ def csvimport(dbo: Database, csvdata: bytes, encoding: str = "utf-8-sig", user:
...
-            a["goodwithkids"] = gkynu(row, "ANIMALGOODWITHKIDS")
+            a["housetrained"] = int(gkynu(row, "ANIMALHOUSETRAINED"))
+            a["cratetrained"] = int(gkynu(row, "ANIMALCRATETRAINED"))
+            a["goodwithcats"] = int(gkynu(row, "ANIMALGOODWITHCATS"))
+            a["goodwithdogs"] = int(gkynu(row, "ANIMALGOODWITHDOGS"))
+            a["goodwithkids"] = int(gkynu(row, "ANIMALGOODWITHKIDS"))
+            a["goodwithelderly"] = int(gkynu(row, "ANIMALGOODWITHELDERLY"))
+            a["goodonlead"] = int(gkynu(row, "ANIMALGOODONLEAD"))
+            a["goodtraveller"] = int(gkynu(row, "ANIMALGOODTRAVELLER"))
+            a["energylevel"] = int(row["ANIMALENERGYLEVEL"])
```
bobintetley, line 472: These should not have int wrapped around them. energylevel should be using gks rather than raw row["ANIMALENERGYLEVEL"] because the latter will fail when that column isn't present in the CSV

adamdspencer77: Done.

src/asm3/animal.py
adamdspencer77: Sorted.

Issue #1629, opened 2025-01-20
Mobile UI: image upload


bobintetley:
Add "take photo" and "upload from the gallery" buttons to the new image gallery widget in the mobile UI. It should therefore support animals and incidents. 

I think the best place for the buttons might be to the left of the thumbnails.

PR #1631, opened 2025-01-22
Image uploads from animal and incident records on mobile UI #1629

adamdspencer77:


src/static/js/mobile2.js
```diff
@@ -8,6 +8,80 @@ $(document).ready(function() {
...
+            reader.addEventListener("load", function() {
+                let formdata = "animalid=" + animalid + "&type=" + uploadtype + "&filename=" + encodeURIComponent(file.name) + "&filedata=" + encodeURIComponent(reader.result);
+                let targeturl =  "mobile_photo_upload";
+                $.ajax({
+                    method: "POST",
+                    url: targeturl,
+                    data: formdata,
+                    dataType: "text",
+                    error: function(e) {
+                        //console.log("Error");
```
bobintetley, line 22: there's a mobile.show_error function that pops up a dialog - might be nice to tell the user if something went wrong!

src/static/js/mobile2.js
```diff
@@ -8,6 +8,80 @@ $(document).ready(function() {
...
+                    },
+                    success: function(mid) {
+                        let newthumbnail = '<div class="media-thumb" style="position: relative; border-style: solid; border-width: 1px; ' +
```
bobintetley, line 25: a single function to render media thumbnails would make sense instead of copy/pasting this html and styling in multiple places - the only thing it needs passing is the media id and date string as far as I can tell?

src/static/js/mobile2.js
```diff
@@ -8,6 +8,80 @@ $(document).ready(function() {
...
+            reader.addEventListener("load", function() {
+                let formdata = "incidentid=" + incidentid + "&type=" + uploadtype + "&filename=" + encodeURIComponent(file.name) + "&filedata=" + encodeURIComponent(reader.result);
+                let targeturl =  "mobile_photo_upload";
+                $.ajax({
+                    method: "POST",
+                    url: targeturl,
+                    data: formdata,
+                    dataType: "text",
+                    error: function(e) {
+                        //console.log("Error");
```
bobintetley, line 59: same as above, tell the user that an error happened

src/static/js/mobile2.js
```diff
@@ -8,6 +8,80 @@ $(document).ready(function() {
...
+                    },
+                    success: function(mid) {
+                        let newthumbnail = '<div class="media-thumb" style="position: relative; border-style: solid; border-width: 1px; ' +
```
bobintetley, line 62: same as above, put this in a shared function

src/static/js/mobile2.js
```diff
@@ -573,6 +647,39 @@ $(document).ready(function() {
...
+            head += '<div style="position: relative; display: inline-block; height: 100px; width: 50px; margin-right: 10px; margin-bottom: 10px;">' +
+                    '<div style="position: absolute; padding: 5px;">' +
+                    '<button id="' + id + '-media-add-camera" class="btn btn-primary mt-1" style="display: block;"><i class="bi-camera"></i></button>' +
+                    '<button id="' + id + '-media-add-gallery" class="btn btn-secondary mt-1" style="display: block;"><i class="bi-card-image"></i></button>' +
+                    '</div></div>' + 
+                    '<input id="' + id + '-image-slider-upload-camera" type="file" capture="environment" accept="image/*" style="display: none;">' +
+                    '<input id="' + id + '-image-slider-upload-file" type="file" accept="image/*" multiple="multiple" style="display: none;">';
+            $.each(rows, function(i, row) {
+                if (row.MEDIAMIMETYPE != 'image/jpeg') { return; }
+                head += '<div class="media-thumb" style="position: relative; border-style: solid; border-width: 1px; ' +
```
bobintetley, line 669: call to new shared function to render a media thumbnail

src/static/js/mobile2.js
adamdspencer77: Done.

src/static/js/mobile2.js
adamdspencer77: Done.

src/static/js/mobile2.js
adamdspencer77: Done.

src/static/js/mobile2.js
adamdspencer77: Done.

src/static/js/mobile2.js
```diff
@@ -8,6 +8,81 @@ $(document).ready(function() {
 
         post_handler: "mobile2",
 
+        media_slider_thumbnail: function(mid, id, date="", notes="") {
```
bobintetley, line 11: would have called it render_image_thumbnail, also why is it right at the top of the file? Should probably be down below with render_image_slider

adamdspencer77: Just put it at the top for ease of finding during development. Renamed and rearranged :)

Issue #1370, opened 2023-11-29
Unsubscribe link


bobintetley:
Have an option when sending mail merges from the mail merge screen to include an unsubscribe link at the bottom of the email. 

This will link to a service API call method "unsubscribe" for the account and person id. It will set the excludefrombulkemail flag on that person record and link them to a static page saying they've been unsubscribed.

jliu1016:
Hello, was interested in this project as someone who fosters for a rescue that still pretty much uses just pen and paper. I figured I might also try tackle some issues since I really appreciate the work being done here.  If I wanted to work on this issue do I just add a comment here and get it assigned to me? What is the best way to coordinate here?

bobintetley:
Hi, that's great. Just a comment that you'd like to work on it is fine, however it won't necessarily stop someone else looking at it. The best thing to do is crack on and send a PR.

jliu1016:
Seems like I don't have permission to push a new branch?

bobintetley:
I think this is the way github works - you have to fork the branch in your own account, and then submit a pull request against our master branch so we can review your code. Thanks!

PR #1640, opened 2025-01-27
add unsubscribe service and static unsubscribe page #1370

jliu1016:
Add a service endpoint to unsubscribe from emails by setting the flag on the owner to be false for ExcludeFromBulkEmail and redirecting to a static unsubscribe page.

This pr does not include the work to add the UI changes to the merge email page, nor the necessary changes to _send_bulk_email to include an unsubscribe link.


bobintetley:
Hi, thanks for doing this, hope you don't mind but I made a few tweaks to use flag functions in person instead.

Issue #1121, opened 2022-04-18
Payment Processor: Square


bobintetley:
Quite a few of our customers are using Square to take payments and would like integration with SM. They have a checkout system and their dev API seems very similar to Stripe. It should be straightforward to write a handler for them.

https://developer.squareup.com/docs/checkout-api-overview

KieranDBARC:
Would you also be able to look at Zettle please

bobintetley:
https://squareup.com/gb/en/online-checkout

bobintetley:
For all payment services, we need a way to generate a checkout page, and a way to have that processor call an HTTP endpoint to tell us when a payment is complete. See src/asm3/paymentprocessors for the stripe/paypal implementations. They both have endpoints name pp_paypal and pp_stripe in main.py that passes through to the payment processor.

PR #1642, opened 2025-01-29
Added support for Square payments #1121

adamdspencer77:


src/static/js/options.js
```diff
@@ -1520,6 +1520,21 @@ $(function() {
...
+                '<hr/>',
+                '<p class="centered"><img height="25px" src="static/images/ui/logo_square_100.png" /></p>',
+                '<table>',
+                '<tr><td><label for="squareaccesstoken">' + _("Square Access Token") + '</label></td>',
+                '<td><input data="SquareAccessToken" id="squareaccesstoken" type="text" class="asm-textbox asm-doubletextbox asm-mask" /></td></tr>',
+                '<tr><td><label for="squarelocationid">' + _("Square Location ID") + '</label></td>',
+                '<td><input data="SquareLocationID" id="squarelocationid" type="text" class="asm-textbox asm-doubletextbox" /></td></tr>',
+                '</table>',
+                '<p class="centered">',
+                    _("In your Square account, enable Instant Payment Notifications with a URL of {0}")
```
bobintetley, line 1533: I doubt Square will call it Instant Payment Notification - that's a PayPal thing. Do they call it a webhook or something? Presume you've got a Square account where you can look up what they call it!

unittest/test_paymentprocessor.py
```diff
@@ -66,34 +67,131 @@ def tearDown(self):
...
+        }
+        post = asm3.utils.PostedData(data, "en")
+        self.recid = asm3.financial.insert_donation_from_form(base.get_dbo(), "test", post)
+
+    def tearDown(self):
+        asm3.financial.delete_donation(base.get_dbo(), "test", self.dueid)
+        asm3.financial.delete_donation(base.get_dbo(), "test", self.recid)
+        asm3.person.delete_person(base.get_dbo(), "test", self.personid)
+    
+    def test_square_ipn_success(self):
```
bobintetley, line 189: not an ipn! I'd call this method test_square

unittest/test_paymentprocessor.py
```diff
@@ -7,23 +8,23 @@
 import asm3.paymentprocessor.paypal
```
bobintetley, line 8: should be an import here for your square paymentprocessor, surprised it didn't complain about that!

src/main.py
```diff
@@ -6040,6 +6042,52 @@ def post_all(self, o):
...
+    def post_all(self, o):
+        asm3.al.debug(o.data, "main.pp_square")
+        try:
+            j = asm3.utils.json_parse(o.data)
+
+            #if "client_reference_id" not in j["data"]["object"]:
+            #    asm3.al.error("client_reference_id missing, this is not an ASM requested payment", "main.pp_stripe")
+            #    return # OK 200, this payment notification is not for us
+            client_reference_id = j["data"]["object"]["payment"]["note"]
+            dbname = client_reference_id[-2:client_reference_id.find("-")]
```
bobintetley, line 6065: when you have webhooks on a stripe account, they will notify us of all payments to that account - even ones that we didn't initate. This dbname code looks like it could fail with an exception if the payment note is blank.  

I don't know how Square work here, but in this scenario we would return a 500 status code. Do we need to be returning a 200 even for payments that we can't handle in order to stop them repeatedly sending the notification to us?

Lastly, client_reference_id is the name of a specific stripe field, if you're pulling out the note from your blob of data here, call it note for clarity!

src/asm3/paymentprocessor/square.py
```diff
@@ -0,0 +1,118 @@
...
+            </html>""" % (result.body["payment_link"]["url"],)
+            return s
+        elif result.is_error():
+            #print(result.errors)
+            print("Failure")
+
+    def receive(self, rawdata: str) -> None:
+        """ 
+        Method to be called by the provider via an endpoint on receipt of payment.
+        validate: Whether or not to skip validation of the IPN - useful for testing
```
bobintetley, line 98: comment seems to be wrong, there's no validate parameter here?

src/asm3/paymentprocessor/square.py
```diff
@@ -0,0 +1,118 @@
...
+        if zp != "": return zp
+
+        item_description = item_description or ", ".join(paymenttypes)
+        client_reference_id = "%s-%s" % (self.dbo.database, payref) # prefix database to payref
+        currency = asm3.configuration.currency_code(self.dbo)
+
+        #asm3.al.debug("create square session: api_key=%s, client_reference_id=%s, " \
+        #    "description=%s, amount=%s, currency=%s" % (api_key, client_reference_id, 
+        #    item_description, totalamount + totalvat, currency), "square.checkoutPage", self.dbo)
+
```
bobintetley, line 49: this seems like a useful log message to have, not sure why it's commented out

src/asm3/paymentprocessor/square.py
```diff
@@ -0,0 +1,118 @@
...
+        client = Client(
+        bearer_auth_credentials = BearerAuthCredentials(
+            access_token = asm3.configuration.square_access_token(self.dbo)
+        ),
+        environment='sandbox')
+
+        result = client.locations.list_locations()
+
+        locationid = 'LQQNMQC474MTG'
```
bobintetley, line 61: this seems to be hard coded and not using your config?

src/asm3/paymentprocessor/square.py
```diff
@@ -0,0 +1,118 @@
...
+            totalamount += r.DONATION
+            if r.VATAMOUNT > 0: totalvat += r.VATAMOUNT
+            #if r.VATRATE > 0: vatrate = r.VATRATE
+            paymenttypes.append(r.DONATIONNAME)
+
+        zp = self._checkForZeroPaymentPage(payref, totalamount)
+        if zp != "": return zp
+
+        item_description = item_description or ", ".join(paymenttypes)
+        client_reference_id = "%s-%s" % (self.dbo.database, payref) # prefix database to payref
```
bobintetley, line 43: makes more sense to call it payment note since that's where it's going?

src/asm3/paymentprocessor/square.py
```diff
@@ -0,0 +1,118 @@
...
+        #    item_description, totalamount + totalvat, currency), "square.checkoutPage", self.dbo)
+
+        from square.http.auth.o_auth_2 import BearerAuthCredentials
+        from square.client import Client
+
+        client = Client(
+        bearer_auth_credentials = BearerAuthCredentials(
+            access_token = asm3.configuration.square_access_token(self.dbo)
+        ),
+        environment='sandbox')
```
bobintetley, line 57: this should be a sitedef for switching between sandbox and production in different environments

src/asm3/paymentprocessor/square.py
```diff
@@ -0,0 +1,118 @@
...
+
+        result = client.locations.list_locations()
```
bobintetley, line 59: doesn't appear to be used for anything?

src/asm3/paymentprocessor/square.py
```diff
@@ -0,0 +1,118 @@
...
+
+        locationid = 'LQQNMQC474MTG'
+        
+        result = client.checkout.create_payment_link(
+            body = {
+                "idempotency_key": asm3.utils.uuid_b64(),
```
bobintetley, line 65: Seeing the code now, I'm guessing that the purpose of this key is so that if we request the same payment again, Square will continue to use the same session at their end. This could be the payref to accomplish this rather than a new ID every time

src/asm3/paymentprocessor/square.py
```diff
@@ -0,0 +1,118 @@
...
+        Marks all payments in payref received.
+        trxid: Transaction ID from the payment service (for ChequeNumber column)
+        received (int): The gross amount received
+        vat (int): Any vat/tax amount
+        fee (int): The transaction fee aount charged
+        rawdata (str): The raw data from the payment service.
+        The fee is only applied to the first payment if there are multiple payments in the payref.
+        It is expected that received, vat and fee are all integer currency amounts in whole pence.
+        If there are licenses waiting to renew on this payment, handles calling out to that functionality too.
+        """
```
bobintetley, line 117: if you look at the paypal one, there are a number of checks done in this function - that the payref is valid, and that it hasn't already been received, etc. That functionality should be duplicated here. I think you're checking whether the payment has a complete status in your pp_square endpoint, but that should be done in here as well.

src/static/js/options.js
adamdspencer77: Done.

unittest/test_paymentprocessor.py
adamdspencer77: Done.

unittest/test_paymentprocessor.py
adamdspencer77: Done.

src/main.py
adamdspencer77: Done.

src/asm3/paymentprocessor/square.py
adamdspencer77: Done.

src/asm3/paymentprocessor/square.py
adamdspencer77: Done.

src/asm3/paymentprocessor/square.py
adamdspencer77: Commented it out to simplify until I'd got output. Uncommented.

src/asm3/paymentprocessor/square.py
adamdspencer77: Yep. Left over from testing. Removed.

src/asm3/paymentprocessor/square.py
adamdspencer77: Done.

src/asm3/paymentprocessor/square.py
adamdspencer77: Sorted.

src/asm3/paymentprocessor/square.py
adamdspencer77: Think I've sorted it.

src/asm3/paymentprocessor/square.py
adamdspencer77: Had to compromise some of the desired functionality. It is possible that multiple events may be triggered by Square payments are processed even if they have already been received. Have commented out check for valid payref as will be triggered by hyphen in my localdb setup, can be uncommented when merged with updated base code.

PR #1651, opened 2025-02-04
Added citation numbers to citations. Made citation numbers searchable.

adamdspencer77:


src/asm3/search.py
```diff
@@ -198,6 +202,7 @@ def cp(p):
         acsort = "LASTCHANGEDDATE"
         lasort = "LASTCHANGEDDATE"
         lisort = "ISSUEDATE"
+        cisort = "LASTCHANGEDDATE"
```
bobintetley, line 205: Should this be CITATIONDATE?

src/static/js/citations.js
```diff
@@ -16,6 +16,7 @@ $(function() {
                 columns: 1,
                 width: 550,
                 fields: [
+                    { json_field: "CITATIONNUMBER", post_field: "citationnumber", label: _("Citation Number"), type: "text" },
```
bobintetley, line 19: Make this field mandatory and default it to zero padded ID. Since we won't have ID available at this stage, pass the next citation ID to the UI in the controller object as nextid: dbo.get_id("ownercitation"). Then, the code that saves a new citation can increment controller.nextid with each save.

src/asm3/search.py
adamdspencer77: That fits with how the vouchers are handled but the comment above that section said last changed. I've changed it to CITATIONDATE date.

bobintetley: yeah, I didn't expand the code, but I'm kind of wondering now why licences and vouchers aren't using LASTCHANGEDDATE here when they both have that column. So yeah, you were probably right in the first place!

src/static/js/citations.js
adamdspencer77: Done.

adamdspencer77:
Still not happy with the new dbupdate function. Going to try to tweak the unit tests to make sure that all is well.

Issue #1661, opened 2025-02-07
Auto-populate Jurisdiction field in add incident screen


Tetzor:
Currently the address fields populates automatically when creating a new incident from the selected caller or suspect record (depending which is selected first). It's been requested the Jurisidiction field also auto-populates from the selected record.

PR #1678, opened 2025-02-28
Set juridiction from caller/suspect when adding an incident #1661

adamdspencer77:


src/static/js/incident_new.js
```diff
@@ -64,6 +64,7 @@ $(function() {
                 return true;
             };
             const add_incident = async function(mode) {
+                add_incident.jurisdictionupdated = false;
```
bobintetley, line 67: this probably should have been incident_new rather than add_incident

src/static/js/incident_new.js
```diff
@@ -106,6 +107,11 @@ $(function() {
                     $("#dispatchcounty").val(rec.OWNERCOUNTY);
                     $("#dispatchpostcode").val(rec.OWNERPOSTCODE);
                 }
+                // Default jurisdiction to the caller jurisdiction if it's not set
+                if (!incident_new.jurisdictionupdated) {
+                    $("#jurisdiction").val(rec.JURISDICTIONID);
```
bobintetley, line 112: I'm not sure this is necessary with a variable to track updated, etc. Simpler to prepend a blank option to the jurisdiction box if it doesn't have one already and only overwrite it if the blank option is picked

src/static/js/incident_new.js
adamdspencer77: Doesn't matter any more :)

src/static/js/incident_new.js
adamdspencer77: Done.

Issue #1643, opened 2025-01-30
4 digit padding XXXX and NNNN codes


bobintetley:
Quite a few people have asked if they can make NNNN pad to 4 digits as that is more reflective of the volume of animals that larger shelters have. Same for XXXX

In addition to this, add a new set of codes PPP and PPPP for animal of that species for the year

bobintetley:
there aren't any unit tests for calc_shelter_code and it would be a good candidate for testing of each type of token where possible

PR #1681, opened 2025-02-28
4 digit padding for sheltercodes, PPP codes plus unit tests #1643

adamdspencer77:


src/asm3/animal.py
```diff
@@ -1940,6 +1961,14 @@ def substitute_tokens(fmt: str, year: int, month: int, tyear: int, ever: int, da
             "DateBroughtIn <= ? AND " \
             "AnimalTypeID = ?", ( beginningofyear, endofyear, animaltypeid))
         highesttyear += 1
+    
+    # If our code uses P, calculate the highest code seen for this species this year
+    if codeformat.find("P") != -1 or shortformat.find("P") != -1:
+        highesttyear = dbo.query_int("SELECT MAX(YearCodeID) FROM animal WHERE " \
+            "DateBroughtIn >= ? AND " \
```
bobintetley, line 1968: YearCodeID is for type based counting only, this should be COUNT(ID), also it needs a new variable assigning as this is going to overwrite highesttyear which is for NNN codes. maybe highestsyear since P is for species?

adamdspencer77: Done.

PR #1700, opened 2025-03-17
Use new tax rate lookups for payments and clinic

adamdspencer77:


bobintetley:
Couple of observations:

Lookup labels on screen should be "Tax Rate" - only show either the lookup or the rate field. Show the lookup when it's a new appointment or payment, show the rate box/number when editing.

In lookups.py/get_tax_rates (or stock.py if it's there), add the rate to the name in brackets, so they are displayed by the UI as eg: "Tax Free (0%)"

src/asm3/lookups.py
```diff
@@ -1355,7 +1356,7 @@ def get_urgency_name(dbo: Database, uid: int) -> str:
     return dbo.query_string("SELECT Urgency FROM lkurgency WHERE ID = ?", [uid])
 
 def get_tax_rates(dbo: Database) -> Results:
-    return dbo.query("SELECT * FROM lktaxrate ORDER BY TaxRateName")
+    return dbo.query("SELECT ID, TaxRateName || ' (' || TaxRate || '%)' AS TaxRateName, TaxRate FROM lktaxrate ORDER BY TaxRateName")
```
bobintetley, line 1359: this should be using dbo.sql_concat as || isn't supported by all database engines (older MySQL)

src/static/js/common_widgets.js
```diff
@@ -906,8 +917,26 @@ $.widget("asm.createpayment", {
...
+        let vatrates = [];
+        let defaulttaxrateval = "";
+        $.each(controller.taxrates, function(taxratecount, taxrate) {
+            let optval = taxrate.ID + "|" + taxrate.TAXRATE;
+            if (taxrate.ID == config.integer("AFDefaultTaxRate")) {
+                defaulttaxrateval = optval;
+            }
+            vatrates.push({ID: optval, TAXRATENAME: taxrate.TAXRATENAME});
+        });
+        let vatrateoptions = html.list_to_options(vatrates, "ID", "TAXRATENAME" );
```
bobintetley, line 931: This loop and the two variables vatrates and defaulttaxrateeval seems to be superfluous to me?  Think it could be written as:

```
$("#pm-vatratechoice").html( html.list_to_options(controller.taxrates, "ID", "TAXRATENAME") );
$("#pm-vatratechoice").val(config.integer("AFDefaultTaxRate"));
```

src/static/js/common_widgets.js
```diff
@@ -1322,7 +1351,10 @@ $.widget("asm.payments", {
             '<td class="vat centered nowrap">',
             '<input id="vat{i}" data="vat{i}" type="checkbox" class="asm-checkbox" />',
             '<span id="vatboxes{i}" style="display: none">',
-            '<input id="vatrate{i}" data="vatrate{i}" class="rightalign asm-textbox asm-halftextbox asm-numberbox" value="0" /> %',
+            '<select id="vatratechoice{i}" data="destaccount{i}" class="asm-selectbox asm-halfselectbox">',
+            html.list_to_options(this.options.controller.taxrates, "ID", "TAXRATENAME"),
+            '</select>',
+            '<input id="vatrate{i}" data="vatrate{i}" class="rightalign asm-textbox asm-halftextbox asm-numberbox" value="0" style="display: none;" /> %',
```
bobintetley, line 1357: the % sign between the lookup and amount looks weird now, I think it can be removed

src/static/js/common_widgets.js
```diff
@@ -1361,9 +1393,27 @@ $.widget("asm.payments", {
...
             self.update_totals(); 
         });
+        $("#vatratechoice" + i).change(function() {
+            let taxrate = 0.0;
+            $.each(self.options.controller.taxrates, function(trcount, tr) {
+                if (tr.ID == $("#vatratechoice" + i).val()) {
+                    taxrate = tr.TAXRATE;
+                    return false;
+                }
+            });
```
bobintetley, line 1403: you can use common.get_field to read TAXRATE from controller.taxrates without having to do this loop

src/static/js/common_widgets.js
adamdspencer77: Done.

src/static/js/common_widgets.js
adamdspencer77: Done.

PR #1724, opened 2025-04-02
Signature widget marks records as dirty

adamdspencer77:


bobintetley:
I think this approach is going to fail when used on mobile devices because common_validate.js won't be loaded.

Rather than creating a custom event, I think all you need to do is make sure that the clear and sign button has a unique class name. 

You can then trap onclick on that class name in common_validate.js/bind_dirty to call dirtychange.

In checking this, I found that bind_dirty was horrendously inefficient and was probably causing all the issues you saw with slowdown on the options screen. I've rewritten it to use delegate events now and included the signature widget button so that this is no longer an issue :)

PR #1737, opened 2025-04-16
Added signing link option to "Sign" toolbar dropdown on the media tab

adamdspencer77:


src/static/js/media.js
```diff
@@ -784,6 +790,23 @@ $(function() {
...
             });
 
+            $("#dialog-signlink").dialog({
+                autoOpen: false,
+                width: 550,
+                modal: true,
+                dialogClass: "dialogshadow",
+                show: dlgfx.edit_show,
+                hide: dlgfx.edit_hide
+            });
```
bobintetley, line 800: future reference - you don't need to do this when you're using show_okcancel_dialog

PR #1738, opened 2025-04-16
Added Microchip Status dropdowns

adamdspencer77:


src/static/js/animal.js
```diff
@@ -873,6 +885,7 @@ $(function() {
             if (config.bool("DontShowSize")) { $("#sizerow").hide(); }
             if (config.bool("DontShowWeight")) { $("#kilosrow, #poundsrow").hide(); }
             if (config.bool("DontShowMicrochip")) { $("#microchiprow, #microchiprow2").hide(); }
+            if (config.bool("DontShowMicrochipStatus")) { $("#microchipstatus, #microchipstatus2").hide(); }
```
bobintetley, line 888: DontShowMicrochipStatus should also be set to "Yes" by default in the dict at the top of configuration.py

PR #1740, opened 2025-04-17
Can update additional animal fields in bulk

adamdspencer77:


src/static/js/animal_bulk.js
```diff
@@ -148,6 +148,32 @@ $(function() {
...
+                    choosetypes.push(v);
+                }
+                else if (v.ID == 0) {
+                    v.MOVEMENTTYPE = _("Reservation");
+                    choosetypes.push(v);
+                }
+                else if (v.ID !=8 && v.ID != 9 && v.ID != 10 && v.ID != 11 && v.ID != 12) {
+                    choosetypes.push(v);
+                }
+            });
```
bobintetley, line 164: I don't think this choosetypes block is actually doing anything here? Was it copied from render_movement where it should live?

adamdspencer77: I think you're right. Removed.

PR #1749, opened 2025-05-01
Online forms: email address confirmation

adamdspencer77:


bobintetley:
Add the missing style for input[type='email'] to the screen section for non-mobile in src/media/onlineform/head.html

src/asm3/onlineform.py
```diff
@@ -230,7 +231,13 @@ def get_onlineform_html(dbo: Database, formid: int, completedocument: bool = Tru
         elif f.FIELDTYPE == FIELDTYPE_NUMBER:
             h.append('<input class="asm-onlineform-number" type="text" id="%s" name="%s" %s %s />' % ( fid, cname, autocomplete, requiredtext))
         elif f.FIELDTYPE == FIELDTYPE_EMAIL:
-            h.append('<input class="asm-onlineform-email" type="email" id="%s" name="%s" %s %s />' % ( fid, cname, autocomplete, requiredtext))
+            h.append(
+                '<input class="asm-onlineform-email" type="email" id="%s" name="%s" %s %s />' % ( fid, cname, autocomplete, requiredtext) + 
+                '</td></tr><tr class=" class="asm-onlineform-tr"><td class="asm-onlineform-td">' + 
+                '<label for="%s">Confirm email</label>' % (fid + "verify",) + # To do - check that this non-translatable label is acceptable - Adam.
```
bobintetley, line 237: I think this would be better if the user could influence the confirm email label. We could repurpose the LOOKUPS column to contain the user's confirm email text (and default it to a translatable string for "Confirm email").

Should be straightforward to modify onlineform.js to show #lookups for emailaddress fields and label it as the "Confirm label"

adamdspencer77: Hopefully done.

PR #1754, opened 2025-05-07
Added medical types and custom treatment timings

adamdspencer77:


src/asm3/medical.py
```diff
@@ -1252,28 +1307,41 @@ def insert_profile_from_form(dbo: Database, username: str, post: PostedData) ->
...
-        totalnumberoftreatments = 0
+    customtimingrule = post["customtiming"].strip()
+
+    # Remove trailing comma if it exists
+    if len(customtimingrule) > 0:
+        while customtimingrule[-1] == ",":
+            customtimingrule = customtimingrule[:-1]
+    
+    if post.integer("singlemulti") == 2:
+        totalnumberoftreatments = len(customtimingrule.split(","))
```
bobintetley, line 1318: I'd make some constants at the top and use those here to make this code a bit more readable: 

Eg: TREATMENT_SINGLE = 0, TREATMENT_MULTI = 1, TREATMENT_CUSTOM = 2 

```if post.integer("singlemulti") == TREATMENT_CUSTOM```

File: src/asm3/medical.py [hunk 2]
```diff
@@ -1252,28 +1307,41 @@ def insert_profile_from_form(dbo: Database, username: str, post: PostedData) ->
...
+    else:
+        totalnumberoftreatments = post.integer("totalnumberoftreatments")
+
+        treatmentrule = post.integer("treatmentrule")
+        singlemulti = post.integer("singlemulti")
+        if singlemulti == 0:
```
bobintetley, line 1324: ```if singlemulti == TREATMENT_SINGLE```, etc

src/static/js/medical.js
```diff
@@ -154,8 +173,13 @@ $(function() {
                     { field: "DOSAGE", display: _("Dosage") },
                     { field: "STARTDATE", display: _("Started"), formatter: tableform.format_date },
                     { field: "NAMEDSTATUS", display: _("Status"), formatter: function(row) {
-                        return row.NAMEDSTATUS + ", " + row.NAMEDFREQUENCY + " " + html.icon("right") + " " + row.NAMEDNUMBEROFTREATMENTS +
-                            " (" + row.TREATMENTNUMBER + "/" + row.TOTALTREATMENTS + ")<br/>" +
+                        let status = "";
+                        if (row.CUSTOMTIMINGRULE != "") {
+                            status = row.NAMEDSTATUS + ", " + _("Custom timing");
```
bobintetley, line 178: Would it make sense to output the custom timing rule value here?

src/static/js/medical.js
```diff
@@ -771,6 +861,44 @@ $(function() {
...
+                customtiming = customtiming.slice(0, -1); //Remove trailing commas
+            }
+            $("#customtiming").val(customtiming);
+            if (customtiming.length == 0) {
+                problem = _("No custom rules found");
+            } else {
+                $.each(customtiming.split(","), function(i, v) {
+                    if (v.includes("=")) {
+                        let label = v.split("=")[0].trim();
+                        let value = v.split("=")[1].trim();
```
bobintetley, line 877: These 2 lines can be written as a destructure:

```let [label, value] = v.split("=");```

File: src/static/js/medical.js [hunk 5]
```diff
@@ -771,6 +861,44 @@ $(function() {
...
+                        if (label == "") {
+                            problem = _("Missing label");
+                            return false;
+                        }
+                        if (value == "") {
+                            problem = _("Missing value");
+                            return false;
+                        } else if (!common.is_integer(value)) {
+                            problem = _("Value '{0}' is not an integer".replace("{0}", value));
```
bobintetley, line 886: One of the closing brackets is in the wrong place - should be before .replace

src/static/js/medical.js
```diff
@@ -771,6 +861,44 @@ $(function() {
...
+                            return false;
+                        }
+                    } else {
+                        let value = v.trim()
+                        if (!common.is_integer(value)) {
+                            problem = _("Value '{0}' is not an integer".replace("{0}", value));
```
bobintetley, line 892: Same again!

src/static/js/medicalprofile.js
```diff
@@ -49,36 +50,71 @@ $(function() {
...
+                edit: function(row) {
                     tableform.fields_populate_from_json(dialog.fields, row);
-                    $("#singlemulti").select("value", (row.TOTALNUMBEROFTREATMENTS == 1 ? 0 : 1));
+                    $("#singlemulti").prop("disabled", false);
+                    if (row.CUSTOMTIMINGRULE) {
+                        $("#singlemulti").val(2);
+                    } else if (row.TOTALNUMBEROFTREATMENTS == 1) {
+                        $("#singlemulti").val(0);
+                    } else {
+                        $("#singlemulti").val(1);
```
bobintetley, line 74: I think we should have the constants in here for readability too. Add them to medicalprofile.js and medical.js as members of the medical and medicalprofile objects. You can then add:

```
TREATMENT_SINGLE: 0,
TREATMENT_MULTI: 1,
TREATMENT_CUSTOM: 2
```

Then the line above would be ```$("#singlemulti").val(medicalprofile.TREATMENT_MULTI);``` - due to the vagaries of javascript, you can use integers and strings interchangeably as == comparisons do type conversion. 

```2 == "2"``` is true in javascript, ```2 === "2"``` is false

File: src/static/js/medicalprofile.js [hunk 8]
```diff
@@ -203,6 +293,44 @@ $(function() {
...
+            }
+            $("#customtiming").val(customtiming);
+            if (customtiming.length == 0) {
+                problem = _("No custom rules found");
+            } else {
+                $.each(customtiming.split(","), function(i, v) {
+                    if (v.includes("=")) {
+                        let label = v.split("=")[0].trim();
+                        let value = v.split("=")[1].trim();
+                        if (label == "") {
```
bobintetley, line 310: destructure again, ``` let [label, value] = v.split("="); ```

src/asm3/medical.py
adamdspencer77: Done.

src/static/js/medical.js
adamdspencer77: Done. Do you think it needs a label?

src/static/js/medical.js
adamdspencer77: Done. But turned 2 lines to 3 after adding the trims. Is there a way of integrating the trims into the destructure?

`let [label, value] = v.split("=");
label = label.trim();
value = value.trim();`

src/static/js/medical.js
adamdspencer77: Done.

src/static/js/medical.js
adamdspencer77: Done.

src/static/js/medicalprofile.js
adamdspencer77: Done.

src/asm3/dbupdate.py
```diff
@@ -2591,6 +2604,34 @@ def user(tid: int, username: str, realname: str, password: str, superuser: bool)
     sql += lookup1("lksloglink", "LinkType", 4, _("Waiting List", l))
     sql += lookup1("lksloglink", "LinkType", 5, _("Movement", l))
     sql += lookup1("lksloglink", "LinkType", 6, _("Incident", l))
+    sql += medicaltype(1, _("Worm treatment", l), 0)
+    sql += medicaltype(2, _("Flea treatment", l), 0)
+    sql += medicaltype(3, _("Other parasite treatment", l), 0)
```
bobintetley, line 2609: can we make categories work? 

Parasite: Wormer
Parasite: Flea
Parasite: Other

Surgery: C-Section
Surgery: Other

hmmm

src/asm3/medical.py
```diff
@@ -1004,12 +1047,28 @@ def insert_regimen_from_form(dbo: Database, username: str, post: PostedData) ->
...
-            while created < totalnumberoftreatments:
-                reqdate = insert_treatments(dbo, username, nregid, reqdate, isstart = False)
-                created += 1
+            if post["customtiming"] != "":
+                startdate = post.date("startdate")
+                customtiming = post["customtiming"].strip()
+                if customtiming[-1] == ",":
+                    customtiming = customtiming[:-1]
+                for a in customtiming.split(","):
+                    if "=" in a:
```
bobintetley, line 1056: Kind of surprised this works, but if str supports __contains__ I guess it will. I think everywhere else, we would have used if a.find("=") != -1

src/asm3/medical.py
```diff
@@ -1004,12 +1047,28 @@ def insert_regimen_from_form(dbo: Database, username: str, post: PostedData) ->
...
+                        label = a.split("=")[0].strip()
+                        value = int(a.split("=")[1].strip())
```
bobintetley, line 1058: it's safer here to use asm3.utils.cint rather than int() - cint will return 0 if the value given isn't an integer, where int() will raise an exception

src/static/js/medical.js
```diff
@@ -617,6 +667,27 @@ $(function() {
 
         },
 
+        change_medicaltype: function() {
+            let forcesingletx = false;
+            let mtid = $("#medicaltype").val();
+            $.each(controller.medicaltypes, function(i, mt) {
+                if (mt.ID == mtid) {
```
bobintetley, line 674: This whole loop can be replaced with ```let forcesingletx = common.get_field(controller.medicaltypes, mtid, "FORCESINGLEUSE");```

src/static/js/medical.js
```diff
@@ -628,21 +699,31 @@ $(function() {
                 $("#treatmentrule").select("value", "0");
                 $("#treatmentrule").select("disable");
                 $("#totalnumberoftreatments").val("1");
-
                 $("#timingrulerow").fadeOut();
                 $("#treatmentrulerow").fadeOut();
-            }
-            else {
+                $("#customtimingrow").fadeOut();
+            } else if ($("#singlemulti").val() == 1) {
```
bobintetley, line 705: didn't you add constants so the 1 could be medical.MULTIPLE_TREATMENTS ?

src/static/js/medicalprofile.js
```diff
@@ -123,18 +161,59 @@ $(function() {
...
                 log.error(err, err);
                 tableform.dialog_enable_buttons();   
             }
         },
 
+        change_medicaltype: function() {
+            let forcesingletx = false;
+            let mtid = $("#medicaltype").val();
+            $.each(controller.medicaltypes, function(i, mt) {
+                if (mt.ID == mtid) {
```
bobintetley, line 200: this loop can be replaced too with common.get_field

src/static/js/medicalprofile.js
```diff
@@ -147,8 +226,8 @@ $(function() {
                 $("#totalnumberoftreatments").val("1");
                 $("#timingrulerow").fadeOut();
                 $("#treatmentrulerow").fadeOut();
-            }
-            else {
+                $("#customtimingrow").fadeOut();
+            } else if ($("#singlemulti").val() == 1) {
```
bobintetley, line 230: same again, compare to medicalprofile.TREATMENT_MULTI or whatever the constant is

src/asm3/medical.py
```diff
@@ -909,7 +943,7 @@ def update_medical_treatments(dbo: Database, username: str, amid: int) -> None:
         ldg = dbo.query_date("SELECT DateGiven FROM animalmedicaltreatment WHERE AnimalMedicalID=? ORDER BY DateGiven DESC", [amid])
         insert_treatments(dbo, username, amid, ldg, False)
 
-def insert_treatments(dbo: Database, username: str, amid: int, requireddate: datetime, isstart: bool = True) -> datetime:
+def insert_treatments(dbo: Database, username: str, amid: int, requireddate: datetime, isstart: bool = True, customlabel: str = "") -> datetime:
     """
```
bobintetley, line 947: I think there's functionality missing from update_medical_treatments as when I tested I'm only getting the first treatment record being created, but none after that. I think if the option is on to create all treatments up front then it works.

src/asm3/medical.py
adamdspencer77: Done.

src/asm3/medical.py
adamdspencer77: Done.

src/static/js/medical.js
adamdspencer77: Yep and done.

src/static/js/medicalprofile.js
adamdspencer77: Done.

src/static/js/common.js
```diff
@@ -300,6 +300,17 @@ const common = {
         return false;
     },
 
+    /** 
+     * Returns true if v is an integer
+     */
+    is_integer: function(v) {
+        if (!$.isNumeric(v) || Math.floor(v) != v) {
```
bobintetley, line 307: couldn't this be written as a one-liner?

```return $.isNumeric(v) && Math.floor(v) == v;```

File: src/asm3/dbupdate.py [hunk 18]
```diff
@@ -2591,6 +2604,34 @@ def user(tid: int, username: str, realname: str, password: str, superuser: bool)
...
+    sql += medicaltype(7, _("Antiviral", l), 0)
+    sql += medicaltype(8, _("Pain relief", l), 0)
+    sql += medicaltype(9, _("Euthanasia", l), 1)
+    sql += medicaltype(10, _("Anesthesia", l), 1)
+    sql += medicaltype(11, _("Medicated bath", l), 0)
+    sql += medicaltype(12, _("Examination", l), 1)
+    sql += medicaltype(13, _("Surgery: Sterilization", l), 1)
+    sql += medicaltype(14, _("Surgery: C-Section", l), 1)
+    sql += medicaltype(15, _("X-Ray / Scan", l), 1)
+    sql += medicaltype(16, _("Amputation", l), 1)
```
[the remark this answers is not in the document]
adamdspencer77: Yep. Done.

src/asm3/dbupdate.py
```diff
@@ -2591,6 +2604,34 @@ def user(tid: int, username: str, realname: str, password: str, superuser: bool)
...
+    sql += medicaltype(17, _("Other", l), 0)
+    sql += medicaltype(18, _("Microchip", l), 1)
+    sql += medicaltype(19, _("Tattoo", l), 1)
```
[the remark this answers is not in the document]
adamdspencer77: Done.

src/static/js/common.js
adamdspencer77: Yes and done.

src/asm3/medical.py
```diff
@@ -281,6 +288,32 @@ def get_batch_for_vaccination_types(dbo: Database) -> Results:
         "AND DateOfVaccination Is Not Null AND DateOfVaccination >= ? " \
         "ORDER BY animalvaccination.ID DESC, VaccinationID", [ dbo.today(offset=-31) ], distincton="ID")
 
+def get_medical_types(dbo: Database, animalid: int) -> Results:
```
bobintetley, line 291: Bit of clarity in the name needed, get_medical_types_animal maybe? Otherwise, I'd think this was just going to return the lookup of medical types.

src/asm3/medical.py
```diff
@@ -1015,6 +1058,21 @@ def insert_regimen_from_form(dbo: Database, username: str, post: PostedData) ->
         # just create the first treatment(s).
         update_medical_treatments(dbo, username, nregid)
 
+    if post["customtiming"] != "":
+        startdate = post.date("startdate")
```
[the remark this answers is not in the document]
adamdspencer77: Hopefully done.

src/asm3/medical.py
adamdspencer77: Done.

PR #1755, opened 2025-05-08
Citation additional fields

adamdspencer77:


src/static/js/header_additional.js
```diff
@@ -223,7 +223,7 @@ additional = {
         // return additional.map((item) => ({ ...item, VALUE: (row.hasOwnProperty(item.FIELDNAME.toUpperCase()) ? row[item.FIELDNAME.toUpperCase()] : "") }));
         $.each(additional, function(i, a) {
             let fieldname = a.FIELDNAME.toUpperCase();
-            a.VALUE = "";
+            a.VALUE = a.DEFAULTVALUE;
```
bobintetley, line 226: I don't think this is right here - it's going to set the default value when you're editing a citation if the additional field is blank. 

Instead, I think the citations screen should be calling additional.reset_defaults(selector) in the onload callback of dialog_show_add

src/static/js/citations.js
```diff
@@ -23,21 +23,33 @@ $(function() {
...
                 idcolumn: "ID",
                 edit: async function(row) {
-                    await tableform.dialog_show_edit(dialog, row);
-                    tableform.fields_update_row(dialog.fields, row);
-                    row.CITATIONNAME = common.get_field(controller.citationtypes, row.CITATIONTYPEID, "CITATIONNAME");
-                    row.OWNERNAME = $("#person").personchooser("get_selected").OWNERNAME;
-                    await tableform.fields_post(dialog.fields, "mode=update&citationid=" + row.ID, "citations");
-                    tableform.table_update(table);
-                    tableform.dialog_close();
+                    await tableform.dialog_show_edit(dialog, row, {
```
bobintetley, line 35: You don't need to await since your handler is an onchange event

src/static/js/citations.js
```diff
@@ -80,31 +92,45 @@ $(function() {
...
+                                    row.OWNERNAME = $("#person").personchooser("get_selected").OWNERNAME;
+                                    controller.rows.push(row);
+                                    
+                                    tableform.table_update(table);
+                                    tableform.dialog_close();
+                                    controller.nextid++;
+                                    $("#citationnumber").val(format.padleft(controller.nextid, 6));
+                                });
+                            },
+                            onload: function() {
```
bobintetley, line 118: minor thing but I'd put onload function above onadd for readability

src/static/js/citations.js
```diff
@@ -80,31 +92,45 @@ $(function() {
...
+                    click: function() { 
+                        tableform.dialog_show_add(dialog, {
+                            onadd: function() {
+                                var incid = "";
+                                if (controller.incident) { incid = controller.incident.ACID; }
+                                
+                                let afpost = additional.additional_fields_post(controller.additional, 19);
+                                
+                                tableform.fields_post(dialog.fields, "mode=create&incident=" + incid + afpost, "citations").then(function(response) {
+                                    var row = {};
```
bobintetley, line 104: Use let over var to avoid unnecessary hoisting

src/static/js/citations.js
```diff
@@ -80,31 +92,45 @@ $(function() {
...
-                        row.OWNERNAME = $("#person").personchooser("get_selected").OWNERNAME;
-                        controller.rows.push(row);
-                        tableform.table_update(table);
-                        tableform.dialog_close();
-                        controller.nextid++;
-                        $("#citationnumber").val(format.padleft(controller.nextid, 6));
+                    click: function() { 
+                        tableform.dialog_show_add(dialog, {
+                            onadd: function() {
+                                var incid = "";
```
bobintetley, line 98: Use let over var to avoid unnecessary hoisting

src/static/js/citations.js
adamdspencer77: Done.

src/static/js/citations.js
adamdspencer77: Done.

src/static/js/citations.js
adamdspencer77: Done.

src/static/js/citations.js
adamdspencer77: Done.

PR #1756, opened 2025-05-09
templatedocument, templatehtml - saving to them writes to deletion table

adamdspencer77:
Working but html templates didn't fit nicely with the existing tools. The templates look to be assigned a new ID each time they're updated and the name must be unique. The restore button "winds back" templates as long as, for document templates - a template with the same ID, for html templates - a template with that name.

### DIFF
--- a/src/asm3/dbms/base.py
+++ b/src/asm3/dbms/base.py
@@ -916,7 +916,6 @@ class Database(object):
         function that Writes an INSERT query for a result row
         """
         fields = []
-        donefields = False
         values = []
         for k in sorted(r.keys()):
             if not donefields:
@@ -929,7 +928,6 @@ class Database(object):
         """
         function that Writes an UPDATE query for a result row
         """
-        donefields = False # Don't know what the purpose of donefields is, code based on row_to_insert_sql(), left in as looked important - Adam.
         cdata = []
         rid = 0
         for k in sorted(r.keys()):
--- a/src/asm3/dbms/base.py
+++ b/src/asm3/dbms/base.py
@@ -918,10 +918,8 @@ class Database(object):
         fields = []
         values = []
         for k in sorted(r.keys()):
-            if not donefields:
-                fields.append(k)
+            fields.append(k)
             values.append(self.sql_value(r[k]))
-        donefields = True
         return "INSERT INTO %s (%s) VALUES (%s);\n" % (table, ",".join(fields), ",".join(values))
     
     def row_to_update_sql(self, table: str, r: ResultRow, uniquecol = "ID", escapeCR: str = "") -> str:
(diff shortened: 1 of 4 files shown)


src/asm3/template.py
```diff
@@ -31,6 +31,8 @@ def get_html_template_names(dbo: Database) -> List[str]:
 
 def update_html_template(dbo: Database, username: str, name: str, head: str, body: str, foot: str, builtin: bool = False) -> None:
     """ Creates/updates an HTML publishing template """
+    row = dbo.first_row(dbo.query("SELECT * FROM templatehtml WHERE Name = ?", (name,)))
+    sql = dbo.row_to_update_sql("templatehtml", row, "NAME")
     dbo.execute("DELETE FROM templatehtml WHERE Name = ?", [name])
     htid = dbo.insert("templatehtml", {
         "Name":     name,
```
bobintetley, line 38: Wouldn't it make life much easier if we just used the same ID again? We can use it to indicate whether it's new or not and if it isn't new, delete it first and handle the write to deletion.

```
htid = dbo.query_int("SELECT ID FROM templatehtml WHERE Name=?", [name])
if htid == 0: 
    htid = dbo.get_id("templatehtml")
else:
    dbo.delete(templatehtml, htid, username)

dbo.insert("templatehtml", {
    "ID":        htid,
   "Name":   name,
... etc ...
}, generateID=False) 
```

You don't need to manually call insert_deletion then as dbo.delete does it for you.

src/asm3/dbms/base.py
```diff
@@ -924,6 +924,24 @@ def row_to_insert_sql(self, table: str, r: ResultRow, escapeCR: str = "") -> str
             values.append(self.sql_value(r[k]))
         donefields = True
         return "INSERT INTO %s (%s) VALUES (%s);\n" % (table, ",".join(fields), ",".join(values))
+    
+    def row_to_update_sql(self, table: str, r: ResultRow, uniquecol = "ID", escapeCR: str = "") -> str:
+        """
+        function that Writes an UPDATE query for a result row
+        """
+        donefields = False # Don't know what the purpose of donefields is, code based on row_to_insert_sql(), left in as looked important - Adam.
```
bobintetley, line 932: Doesn't look like it's doing anything to me. I guess it was copied from some function that output a header row originally and wanted to do something only on the first row. It's redundant both here and in  row_to_insert_sql

src/asm3/template.py
```diff
@@ -148,10 +153,15 @@ def rename_document_template(dbo: Database, username: str, dtid: int, newname: s
 
 def update_document_template_content(dbo: Database, username: str, dtid: int, content: bytes) -> None:
     """ Changes the content of a template """
+    row = dbo.first_row(dbo.query("SELECT * FROM templatedocument WHERE ID = ?", (dtid,)))
+    sql = dbo.row_to_update_sql("templatedocument", row)
+
```
bobintetley, line 158: this is good, works well for this bit :)

src/asm3/dbms/base.py
adamdspencer77: Removed.

PR #1757, opened 2025-05-09
Transport "View on map"

adamdspencer77:


src/static/js/transport.js
```diff
@@ -249,6 +249,33 @@ $(function() {
...
+                },
+                { id: "mapview", text: _("Map"), icon: "map", enabled: "one", perm: "atr",
+                    click: function() {
+                        let row = tableform.table_selected_row(table);
+                        let fromlocation = "";
+                        let tolocation = "";
+                        if (row.PICKUPPOSTCODE == "") {
+                            fromlocation = row.PICKUPADDRESS.replace(" ", "%20").replace("\n", "+");
+                        }
+                        if (row.DROPOFFPOSTCODE == "") {
```
bobintetley, line 261: I'd write as if (!row.DROPOFFPOSTCODE)

src/static/js/transport.js
```diff
@@ -249,6 +249,33 @@ $(function() {
...
+                            tolocation = row.DROPOFFADDRESS.replace(" ", "%20").replace("\n", "+");
+                        }
+                        if (row.PICKUPTOWN != "") {
```
bobintetley, line 264: and if (row.PICKUPTOWN) {

src/static/js/transport.js
```diff
@@ -249,6 +249,33 @@ $(function() {
...
+                            fromlocation += "+" + row.PICKUPTOWN;
+                        }
+                        if (row.DROPOFFTOWN != "") {
+                            tolocation += "+" + row.DROPOFFTOWN;
+                        }
+                        if (row.PICKUPPOSTCODE != "") {
+                            fromlocation += "+" + row.PICKUPPOSTCODE.replace(" ", "%20");
```
bobintetley, line 271: I'd group the pickup/fromlocation building conditions together, rather than mixing them with tolocation ones since they don't seem to affect each other. Makes it a bit clearer to read then I think

src/static/js/transport.js
adamdspencer77: Done.

src/static/js/transport.js
adamdspencer77: Done.

src/static/js/transport.js
adamdspencer77: Done.

adamdspencer77:
Did some more testing and found some issues. Also used encodeURI instead of all the individual replaces.

PR #1768, opened 2025-05-14
Bulk change animals: use correct widget

adamdspencer77:


src/asm3/additional.py
```diff
@@ -338,6 +338,35 @@ def insert_additional(dbo: Database, linktype: int, linkid: int, additionalfield
     except Exception as err:
         asm3.al.error("Failed saving additional field: %s" % err, "additional.insert_additional", dbo, sys.exc_info())
 
+def save_bulk_values_for_link(dbo: Database, post: PostedData, username: str, linkid: int, linktype: str = "animal", setdefaults: bool = False) -> None:
```
bobintetley, line 341: rather than making a whole new identical method without the delete, add a new parameter to save_values_for_link of removeallforlink: bool = True, and test it before deleting all additional fields for the link given.

You can just have your caller set that to false then!

src/asm3/animal.py
```diff
@@ -3427,10 +3427,10 @@ def update_animals_from_form(dbo: Database, username: str, post: PostedData) ->
                 move_dict["reservationstatus"] = asm3.configuration.default_reservation_status(dbo)
                 move_dict["reservationdate"] = post["movementdate"]
             asm3.movement.insert_movement_from_form(dbo, username, asm3.utils.PostedData(move_dict, dbo.locale))
-    
     if post.integer("additionalfield") != -1:
         for animalid in post.integer_list("animals"):
-            asm3.additional.insert_additional(dbo, asm3.additional.ANIMAL, animalid, post.integer("additionalfield"), post["additionalvalue"])
+            #asm3.additional.insert_additional(dbo, asm3.additional.ANIMAL, animalid, post.integer("additionalfield"), post["additionalvalue"])
```
bobintetley, line 3432: I'd take this comment out to avoid later confusion :)

src/static/js/animal_bulk.js
```diff
@@ -193,6 +195,7 @@ $(function() {
                 $("#button-update").button("disable");
                 header.show_loading(_("Updating..."));
                 let formdata = "mode=update&" + $("input, select, textarea").toPOST();
+                console.log(formdata);
```
bobintetley, line 198: debug code that should be removed :)

src/asm3/additional.py
adamdspencer77: Done.

src/asm3/animal.py
adamdspencer77: Done.

src/static/js/animal_bulk.js
adamdspencer77: Done.

PR #1771, opened 2025-05-19
Report PDF export, allow landscape option

adamdspencer77:


src/asm3/utils.py
```diff
@@ -1589,10 +1589,12 @@ def html_to_text(htmldata: str) -> str:
     p.feed(htmldata)
     return "".join(p.s)
 
-def html_to_pdf(dbo: Database, htmldata: str) -> bytes:
+def html_to_pdf(dbo: Database, htmldata: str, orientation: str = "portrait") -> bytes:
```
bobintetley, line 1592: We shouldn't be modifying this method signature.

We should be prepending the HTML comment/directive to the start of the content in the caller that is passing htmldata to this function. 

It should have been done in main.py/report_export_pdf

src/main.py
```diff
@@ -7069,7 +7069,10 @@ def content(self, o):
         disp = asm3.configuration.pdf_inline(dbo) and "inline" or "attachment"
         self.content_type("application/pdf")
         self.content_disposition(disp, "report.pdf")
-        return asm3.utils.html_to_pdf(dbo, asm3.reports.execute(dbo, crid, o.user, p))
+        orientation = "portrait"
```
bobintetley, line 7072: execute the report here, then prepend any directive as needed before the call to html_to_pdf

src/asm3/utils.py
adamdspencer77: Hopefully done.

src/main.py
adamdspencer77: Done.

PR #1772, opened 2025-05-23
Product images

adamdspencer77:


src/asm3/configuration.py
```diff
@@ -435,6 +435,7 @@
     "WatermarkYOffset": "10",
     "WeightChangeLog": "Yes",
     "WeightChangeLogType": "4",
+    "ProductImageScale": "200x200",
```
bobintetley, line 438: This dict should be in alphabetical order so we can find things later

src/asm3/dbupdate.py
```diff
@@ -1721,7 +1721,7 @@ def fclob(name: str, nullable: bool = True) -> str:
         fstr("Barcode", True),
         fint("Cost", True),
         fint("UnitPrice", True),
-        fdate("CreatedDate") ), False)
+        fdate("CreatedDate"), ), False)
```
bobintetley, line 1724: not sure why this was added?

src/main.py
```diff
@@ -6799,7 +6799,8 @@ def controller(self, o):
             "stockusagetypes": asm3.lookups.get_stock_usage_types(dbo),
             "units": asm3.lookups.get_unit_types(dbo),
             "yesno": asm3.lookups.get_yesno(dbo),
-            "rows": products
+            "rows": products,
+            "imagedimesions": asm3.configuration.DEFAULTS["ProductImageScale"]
```
bobintetley, line 6803: returning the default here means we will never be able to override it. Add a product_image_scale method to configuration.py that works as other methods in there to return the set value or default

src/static/js/product.js
```diff
@@ -6,6 +6,8 @@ $(function() {
 
     const product = {
 
+        LINKTYPEID: 7,
```
bobintetley, line 9: this should be MEDIA_LINKTYPEID for clarity

src/static/js/product.js
```diff
@@ -56,11 +69,27 @@ $(function() {
...
+                        $("#removeimage").hide();
+                    }
                     tableform.dialog_show_edit(dialog, row, {
-                        onchange: function() {
+                        onchange: async function() {
+                            let selectedfile = $("#imageinput")[0].files[0];
+                            if (selectedfile) {
+                                row.MEDIAID = await product.attach_image(selectedfile, "imageinput", row.ID);
+                            } else {
+                                row.MEDIAID = "";
```
bobintetley, line 87: everywhere else uses a 0 for a foreign key that isn't linked to anything. Using empty string means inconsistent types. It still works fine for falsy tests

src/asm3/dbupdate.py
adamdspencer77: I think I originally added image fields to stock movements then removed them and using product images only. I think this is left over from that.

adamdspencer77: Removed rogue comma.

src/main.py
adamdspencer77: Done.

src/main.py
```diff
@@ -6815,11 +6816,23 @@ def post_move(self, o):
     def post_update(self, o):
         self.check(asm3.users.CHANGE_STOCKLEVEL)
         asm3.stock.update_product_from_form(o.dbo, o.post, o.user)
+        if not o.post["mediaid"]:
```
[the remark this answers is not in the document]
adamdspencer77: Done.

PR #1774, opened 2025-05-29
Online forms: phone number type field

adamdspencer77:


src/static/js/onlineform_extra.js
```diff
@@ -243,6 +253,22 @@ $(document).ready(function() {
         return rv;
     };
 
+    const validate_phone = function() {
+        let rv = true;
+        $(".asm-onlineform-phone").each(function() {
+            let v = $(this).val();
+            if (v) {
+                if (v.replace(/\D/g, '').length < 6) {
```
bobintetley, line 261: tricksy, I like this. Assume \D is "not digit" so it's removing everything that isn't a number and checking the length of what's left :)

src/asm3/onlineform.py
```diff
@@ -354,6 +356,20 @@ def get_onlineform_html(dbo: Database, formid: int, completedocument: bool = Tru
...
+            if dbo.locale == "en":
+                placeholder = "NNN NNN-NNNN"
+            elif dbo.locale == "en_AU":
+                placeholder = "NN NNNN NNNN"
+            elif dbo.locale == "en_CA":
+                placeholder = "NNN NNN-NNNN"
+            elif dbo.locale == "fr_CA":
+                placeholder = "NNN NNN-NNNN"
+            elif dbo.locale == "en_GB":
+                placeholder = "NNNN NNNNNNN"
```
bobintetley, line 370: I'm not keen on the placeholders (a presentation thing) being in the python while the rules are in the javascript. For consistency and saving of pain when we have to update it or add a new one, the placeholders should be in the javascript too.

src/static/js/onlineform_extra.js
```diff
@@ -3,6 +3,16 @@
...
+const PHONE_RULES = [
+    { locale: "en", prefix: "", length: 10, elements: 3, extract: /^(\d{3})(\d{3})(\d{4})$/, display: "({1}) {2}-{3}" },
+    { locale: "en_AU", prefix: "04", length: 10, elements: 3, extract: /^(\d{4})(\d{3})(\d{3})$/, display: "{1} {2} {3}" },
+    { locale: "en_AU", prefix: "", length: 10, elements: 3, extract: /^(\d{2})(\d{4})(\d{4})$/, display: "{1} {2} {3}" },
+    { locale: "en_CA", prefix: "", length: 10, elements: 3, extract: /^(\d{3})(\d{3})(\d{4})$/, display: "({1}) {2}-{3}" },
+    { locale: "fr_CA", prefix: "", length: 10, elements: 3, extract: /^(\d{3})(\d{3})(\d{4})$/, display: "({1}) {2}-{3}" },
+    { locale: "en_GB", prefix: "011", length: 11, elements: 2, extract: /^(\d{4})(\d{7})$/, display: "{1} {2}" },
+    { locale: "en_GB", prefix: "", length: 11, elements: 2, extract: /^(\d{5})(\d{6})$/, display: "{1} {2}" }
+];
+
```
bobintetley, line 15: Add a placeholder property for each locale to these objects. You can then iterate all the .asm-onlineform-phone elements, read their locale and set their placeholder when the script loads.

src/asm3/onlineform.py
adamdspencer77: Done.

src/static/js/onlineform_extra.js
adamdspencer77: :D I copied it from some other code of yours.

PR #1776, opened 2025-05-29
Species specific additional fields

adamdspencer77:


src/static/js/header_additional.js
```diff
@@ -599,7 +599,9 @@ additional = {
                 var d = String(t.attr("data-post"));
                 // mandatory additional fields have a post attribute prefixed with a.1
                 if (d.indexOf("a.1") != -1) {
-                    if (common.trim(t.val()) == "") {
+                    if ($("#species").val() && !t.attr("data-speciesids").split(",").includes($("#species").val()) ) {
+                    }
```
bobintetley, line 603: This doesn't look to be doing anything except I guess making sure that a species specific field can't be blank? I don't like that it's looking for #species anywhere in the DOM too, those kinds of behaviours should be in the caller I think.

src/static/js/animal_new.js
```diff
@@ -578,6 +580,18 @@ $(function() {
                 $("#animalname").val(response); 
             });
 
+            $("#species").change(function() {
+                let speciesid = $("#species").val();
+                $.each($(".additional"), function(i, af) {
+                    if ($(af).attr("data-speciesids").split(",").includes(speciesid)) {
+                        $(af).closest("tr").show();
```
bobintetley, line 587: this is ok for now, but closest("tr") is going to fail one day when we want to get rid of using tables for layouts and switch to bootstrap. I will add a separate ticket for refactoring and having tableform output some kind of form field/row class with all form field rows to use instead

PR #1789, opened 2025-06-17
Checkout and request payment, copy link to clipboard

adamdspencer77:


src/asm3/movement.py
```diff
@@ -1097,9 +1097,7 @@ def send_adoption_checkout(dbo: Database, username: str, post: PostedData) -> No
     }
     asm3.cachedisk.put(key, dbo.name(), co, 86400 * 2) # persist for 2 days
     # Send the email to the adopter
-    body = post["body"]
-    url = "%s?account=%s&method=checkout_adoption&token=%s" % (SERVICE_URL, dbo.name(), key)
-    body = asm3.utils.replace_url_token(body, url, asm3.i18n._("Adoption Checkout", l))
```
bobintetley, line 1102: I think removing these two lines is going to break where the move_adopt endpoint calls the checkout - that endpoint will need to be changed to handle the URL. It's different in that the user doesn't get a tinymce/email dialog - they just pick the template and the URL is generated and added to the email automatically - the replace_url_token function call was to switch $URL in the template for the real URL

The functionality of switching $URL for the real URL probably needs to stay at the backend because there are people using this with existing email templates that expect it to keep working

src/asm3/utils.py
```diff
@@ -1228,6 +1228,15 @@ def url(method, params):
             asm3.al.debug("strip invalid url '%s'" % l, "utils.fix_relative_document_uris", dbo)
     return s
 
+def fix_tinymce_uris(dbo: Database, s: str) -> str:
```
bobintetley: utils functions should have a Python docstring so we know what it does without having to read it!

src/asm3/utils.py
```diff
@@ -1228,6 +1228,15 @@ def url(method, params):
...
+    o = ""
+    for chunk in s.split("<"):
+        if chunk != '':
+            snippet = chunk.split(">")[0]
+            snippet = snippet.replace('&amp;', '&')
+            o += "<" + snippet + ">" + chunk.split(">")[1]
```
bobintetley: This is a cheeky little function, like the use of split instead of regex groups. I have an irrational hatred of string concatenation though - make o a list and append to it instead. This pattern is less CPU intensive than doing lots of concatenation in a loop because you aren't creating a new immutable string for every iteration of the loop.

```
o = []
for ....
    do stuff
    o.append( value )
return "".join(o)
```

src/static/js/movements.js
```diff
@@ -576,6 +576,7 @@ $(function() {
                             header.show_error(_("No adoption fee has been set for this animal."));
                             return;
                         }
+                        let paymentlink = await common.ajax_post("movement", "mode=checkouturl&id=" + row.ID + "&animalid=" + row.ANIMALID + "&personid=" + row.OWNERID);
```
bobintetley, line 579: Something is going to be needed in move_adopt.js as well

src/asm3/movement.py
adamdspencer77: Done.

src/asm3/utils.py
adamdspencer77: Done. Also removed the dbo argument, wasn't used inside the function.

src/asm3/utils.py
adamdspencer77: Done.

src/asm3/movement.py
```diff
@@ -1097,9 +1097,10 @@ def send_adoption_checkout(dbo: Database, username: str, post: PostedData) -> No
     }
     asm3.cachedisk.put(key, dbo.name(), co, 86400 * 2) # persist for 2 days
     # Send the email to the adopter
-    body = post["body"]
     url = "%s?account=%s&method=checkout_adoption&token=%s" % (SERVICE_URL, dbo.name(), key)
+    body = post["body"]
     body = asm3.utils.replace_url_token(body, url, asm3.i18n._("Adoption Checkout", l))
```
bobintetley: Putting these lines back causes problems now for code paths other than move_adopt. This is because replace_url_token appends the URL if it doesn't find a $URL token. 

If you change movement.send_adoption_checkout to accept a boolean parameter for "substutite_url" then the callers in main.py can indicate whether or not they've already dealt with the URL and if they have, don't call replace_url_token

adamdspencer77: Hopefully done.

src/main.py
```diff
@@ -5288,7 +5291,7 @@ def post_create(self, o):
                 "subject":      tokens["SUBJECT"] or _("Adoption Checkout", l),
                 "body":         tokens["BODY"]
             }
-            asm3.movement.send_adoption_checkout(dbo, o.user, asm3.utils.PostedData(d, dbo.locale))
+            asm3.movement.send_adoption_checkout(dbo, o.user, asm3.utils.PostedData(d, dbo.locale), True)
```
bobintetley: I always name optional arguments in function calls. This way, if you need to add more arguments to the function in future, or change the order of optional arguments, you don't have to go and find every caller and update it. 

It also means that you can read what the optional arguments are when you're looking at the caller without having to go check the function.

So True here should be ```substitute_url=True```

PR #1799, opened 2025-06-20
Clinic: autocomplete from previous invoice items

adamdspencer77:


src/asm3/dbupdates/35011.py
```diff
@@ -0,0 +1,4 @@
+from asm3.dbupdate import execute
+
+execute(dbo, "CREATE TABLE lkclinicinvoiceitems (ID %(int)s NOT NULL PRIMARY KEY, ClinicInvoiceItemName %(short)s, ClinicInvoiceItemDescription %(long)s, DefaultCost %(int)s, IsRetired %(int)s)" % { "int": dbo.type_integer, "short": dbo.type_shorttext, "long": dbo.type_longtext })
+
```
bobintetley: While this is probably ok, safer to use the dbo.ddl functions for  #consistency, eg:

```
fields = ",".join([
    dbo.ddl_add_table_column("ID", dbo.type_integer, False, pk=True),
    dbo.ddl_add_table_column("ClinicInvoiceItemName", dbo.type_shorttext, False),
    dbo.ddl_add_table_column("ClinicInvoiceItemDescription", dbo.type_longtext, True),
    dbo.ddl_add_table_column("DefaultCost", dbo.type_integer, False),
    dbo.ddl_add_table_column("IsRetired", dbo.type_integer, False)
])
execute(dbo, dbo.ddl_add_table("lkclinicinvoiceitems", fields) )
```

src/static/js/clinic_invoice.js
```diff
@@ -10,14 +10,14 @@ $(function() {
             const dialog = {
                 add_title: _("Add Invoice Item"),
                 edit_title: _("Edit Invoice Item"),
+                width: '550px',
                 edit_perm: 'ccl',
                 close_on_ok: false,
                 fields: [
-                    { json_field: "DESCRIPTION", post_field: "description", label: _("Description"), type: "textarea", validation: "notblank" },
+                    { json_field: "DESCRIPTION", post_field: "description", label: _("Description"), type: "autotext", validation: "notblank", autocomplete: "on", doublesize: true },
```
bobintetley: The autocomplete attribute is not doing what you'd expect here. It is setting the autocomplete HTML5 attribute on the input field that will be created. The HTML5 attribute controls whether the browser will try and pick things you've entered before in its own dropdown, so this is probably not what you want and should be removed! See https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/autocomplete

src/static/js/clinic_invoice.js
```diff
@@ -45,8 +45,19 @@ $(function() {
                     click: async function() { 
                         try {
                             await tableform.dialog_show_add(dialog);
+                            $.each($(".recentinvoiceitem"), function(i, v) {
+
+                                console.log('Inserting recent invoice');
```
bobintetley: shouldn't be here :)

src/static/js/clinic_invoice.js
```diff
@@ -45,8 +45,19 @@ $(function() {
...
+                                let response = tableform.fields_post(dialog.fields, "mode=create&appointmentid=" + controller.appointmentid, "clinic_invoice");
+                                let row = {};
+                                row.ID = response;
+                                console.log(dialog.fields);
+                                tableform.fields_update_row(dialog.fields, row);
+                                controller.rows.push(row);
+                            });
+                            console.log('Inserting manual invoice');
```
bobintetley: or here

src/static/js/clinic_invoice.js
```diff
@@ -96,9 +107,25 @@ $(function() {
...
+                });
+            });
         },
 
         sync: function() {
+            let sourcelist = [];
+            clinic_invoice.invoiceitemsdict = {};
+            $.each(controller.invoiceitems, function(i, v) {
+                sourcelist.push(v.CLINICINVOICEITEMNAME);
+                clinic_invoice.invoiceitemsdict[v.CLINICINVOICEITEMNAME] = v.DEFAULTCOST;
```
bobintetley, line 119: I can't see invoiceitemsdict being declared above, so it is being implicitly created here. It looks like you're building invoiceitemsdict as a lookup so that it's easy to look up the cost from the item name - but the code above isn't actually using it, it seems to be looping controller.invoiceitems again!

src/static/js/clinic_invoice.js
```diff
@@ -96,9 +107,25 @@ $(function() {
             tableform.dialog_bind(this.dialog);
             tableform.buttons_bind(this.buttons);
             tableform.table_bind(this.table, this.buttons);
+
+            $("#description").on("change", function() {
+                $.each(controller.invoiceitems, function(i, v) {
+                    if (v.CLINICINVOICEITEMNAME == $("#description").val()) {
```
bobintetley: If this used your clinic_invoice.invoiceitemsdict object (that you create below in sync), then it would be a one-liner instead of a loop wouldn't it?

src/asm3/dbupdates/35011.py
adamdspencer77: Done.

src/static/js/clinic_invoice.js
adamdspencer77: Done.

src/static/js/clinic_invoice.js
adamdspencer77: Removed.

src/static/js/clinic_invoice.js
adamdspencer77: Removed.

src/static/js/clinic_invoice.js
adamdspencer77: Sorted.

src/static/js/clinic_invoice.js
adamdspencer77: Hopefully sorted. Put in an if that avoids messing with the price if the item isn't in the dictionary.

PR #1809, opened 2025-06-27
Animal Links

adamdspencer77:


src/asm3/users.py
```diff
@@ -25,6 +25,7 @@
 MERGE_ANIMAL                    = "ma"
 RESERVESPONSOR_UNIT             = "rsu"
 VIEW_TIMELINE                   = "vti"
+VIEW_ANIMAL_LINKS               = "valk"
```
bobintetley: Don't think this is used or necessary

src/asm3/animal.py
```diff
@@ -2347,6 +2347,78 @@ def delete_animal_entry(dbo: Database, username: str, aeid: int) -> int:
...
+    sql += "UNION SELECT 'AFP' AS TYPE, " \
+        "aff.FieldLabel AS TYPEDISPLAY, o.LastChangedDate AS DDATE, o.ID AS LINKID, " \
+        "o.OwnerName AS LINKDISPLAY, " \
+        "o.OwnerAddress AS FIELD2, " \
+        "CASE WHEN o.IsDeceased=1 THEN 'D' ELSE '' END AS DMOD " \
+        "FROM additional af " \
+        "INNER JOIN additionalfield aff ON aff.ID = af.AdditionalFieldID " \
+        "INNER JOIN owner o ON o.ID = af.LinkID " \
+        "WHERE af.Value = '%d' AND aff.FieldType = %s AND aff.LinkType IN (%s) " % \
+            ( int(aid), asm3.additional.ANIMAL_LOOKUP, asm3.additional.clause_for_linktype("person") )
```
bobintetley, line 2389: I think we should have a union query in here for additional fields linked from waiting list records as that's something people are likely to do.

src/asm3/animal.py
```diff
@@ -2347,6 +2347,78 @@ def delete_animal_entry(dbo: Database, username: str, aeid: int) -> int:
...
+        "'%s' AS TYPEDISPLAY, " \
+        "al.LastChangedDate AS DDATE, " \
+        "al.ID AS LINKID, " \
+        "al.AcceptanceNumber AS LINKDISPLAY, " \
+        "%s AS FIELD2, " \
+        "CASE WHEN a.DeceasedDate Is Not Null THEN 'D' ELSE '' END AS DMOD " \
+        "FROM animallitter al " \
+        "INNER JOIN animal a ON al.ParentAnimalID = a.ID " \
+        "WHERE a.ID != %s " % \
+            ( _("Parent to litter"), litterextra, int(aid) )
```
bobintetley, line 2428: I just thought of somewhere else animals can be linked that isn't visible anywhere else and that's when they are part of events, eg: for adoption events (they will have an entry in the eventanimal table). There should be a union query in here for those as well.

src/asm3/users.py
adamdspencer77: Removed.

src/asm3/animal.py
adamdspencer77: Done.

src/asm3/animal.py
adamdspencer77: Events already have their own slider.

PR #1814, opened 2025-07-01
Person - Details (Couple) additional field link

adamdspencer77:


src/asm3/dbupdates/35012.py
```diff
@@ -0,0 +1,4 @@
+from asm3.dbupdate import execute
+
+execute(dbo, "INSERT INTO lksfieldlink (LinkType) VALUES (?)", [_("Person - Couple", dbo.locale),])
```
bobintetley: How is this query working? It's not specifying an ID! It should be adding this linktype for ID 31. Also, you only need to use the dangling comma in tuples containing one item - it's not necessary with lists

src/asm3/additional.py
```diff
@@ -50,7 +50,7 @@
 INCIDENT_IN = "16, 17, 18, 20"
 LOSTANIMAL_IN = "9, 10"
 MOVEMENT_IN = '22, 23, 24, 25, 26, 27, 28, 29, 30'
-PERSON_IN = "1, 7, 8"
+PERSON_IN = "1, 7, 8, 31"
```
bobintetley, line 54: There also needs to be a matching constant above this in additional.py for PERSON_COUPLE = 31

src/static/js/person.js
```diff
@@ -609,6 +613,13 @@ $(function() {
             // If a popup warning has been set, display it
             person.show_popup_warning();
 
+            // If couple show additional couple fields
+            /*if ($("#ownertype").val() == 3) {
+                $(".additionalcouples").closest("tr").show();
+            } else {
+                $(".additionalcouples").closest("tr").hide();
+            }*/
```
bobintetley: Probably should have been removed?

src/asm3/dbupdates/35012.py
adamdspencer77: Done.

src/static/js/person.js
adamdspencer77: Removed.

src/asm3/additional.py
adamdspencer77: Done.

PR #1820, opened 2025-07-14
Online form import button - validate that the file is .html or .json

adamdspencer77:


src/static/js/onlineforms.js
```diff
@@ -289,6 +290,14 @@ $(function() {
             tableform.buttons_bind(this.buttons);
             tableform.table_bind(this.table, this.buttons);
             this.load_person_flags();
+            $("#filechooser").change(function() {
+                let importfile = $("#filechooser").val();
+                if (['html', 'json'].includes(importfile.split(".").slice(-1)[0].toLowerCase()) == false) {
```
bobintetley: This is very tricksy but I think might break if you feed it a file with no extension? A simple 

```
!importfile.endsWith(".html") && !importfile.endsWith(".json")
```

would have done the job!

src/static/js/onlineforms.js
```diff
@@ -289,6 +290,14 @@ $(function() {
...
+                    $("#dialog-import .ui-state-error").show();
+                    $("#filechooser").val('');
+                }
+            });
+            $("#dialog-import .ui-state-error").hide();
```
bobintetley: I think it would also be nice to validate the actual content in the backend, so have the post_import function of the onlineforms endpoint in main.py verify that the content contains <html before trying to parse it as such

src/static/js/onlineforms.js
adamdspencer77: Done. I snuck in  a couple of .toLowerCases().

src/static/js/onlineforms.js

src/static/js/onlineforms.js
```diff
@@ -257,9 +257,10 @@ $(function() {
         render_import: function() {
             return [
                 '<div id="dialog-import" style="display: none" title="' + html.title(_("Import from file")) + '">',
+                html.error('File must be html or json format'),
```
bobintetley: You can give the error box an id attribute in the second argument, eg: 

```html.error("File must be html or json format", "formaterror"),```

File: src/static/js/onlineforms.js [hunk 4]
```diff
@@ -289,6 +290,14 @@ $(function() {
             tableform.buttons_bind(this.buttons);
             tableform.table_bind(this.table, this.buttons);
             this.load_person_flags();
+            $("#filechooser").change(function() {
+                let importfile = $("#filechooser").val();
+                if ( !importfile.toLowerCase().endsWith(".html") && !importfile.toLowerCase().endsWith(".json") ) {
+                    $("#dialog-import .ui-state-error").show();
```
bobintetley: If you gave it an id above, you can use ```$("#formaterror").show();``` for legibility and it means if we stop using jquery UI in the future, we aren't dependent on JQUI specific classes

src/static/js/onlineforms.js
```diff
@@ -289,6 +290,14 @@ $(function() {
...
+                    $("#filechooser").val('');
+                }
+            });
+            $("#dialog-import .ui-state-error").hide();
```
bobintetley: and here

src/static/js/onlineforms.js

src/static/js/onlineforms.js

src/static/js/onlineforms.js

PR #1824, opened 2025-07-23
Batch send donation receipts

adamdspencer77:


src/static/js/receipt_bulk.js
```diff
@@ -0,0 +1,119 @@
...
+                { type: "raw", markup: '<center>'},
+                { type: "raw",  markup: tableform.render_select({
+                        'id': 'paymenttemplate',
+                        'justwidget': true,
+                        'options': { displayfield: "NAME", valuefield: "ID", rows: controller.templates },
+                        'tooltip': _("Template")
+                    })
+                },
+                { type: "raw", markup: _("from") + ' '},
+                { type: "raw", markup: tableform.render_date({'id': "fromdate",'justwidget': true, 'halfsize': true, value: format.date(controller.fromdate), tooltip: _("From date")}) },
```
bobintetley: very minor thing, but why are the keys quoted in this javascript object?

src/static/js/receipt_bulk.js
```diff
@@ -0,0 +1,119 @@
...
+                { type: "raw", markup: ' ' + _('to') + ' ' },
+                { id: "todate", type: "raw", markup: tableform.render_date({'id': "todate",'justwidget': true, 'halfsize': true, value: format.date(controller.todate), tooltip: _("To date")}) },
```
bobintetley: more quoted keys

src/static/js/receipt_bulk.js
```diff
@@ -0,0 +1,119 @@
...
+                        common.route("receipt_bulk?fromdate=" + $("#fromdate").val() + "&todate=" + $("#todate").val() + "&paymentmethod=" + $("#paymentmethod").val() + "&templateid=" + $("#paymenttemplate").val());
+                    }
+                },
+                { id: "send", text: _("Send"), icon: "email", enabled: "multi", 
+                    click: async function() {
+                        $("#button-send").button('disable');
+                        let ids = tableform.table_ids(table);
+                        await common.ajax_post("receipt_bulk", "mode=send&ids=" + ids + "&tid=" + $("#paymenttemplate").val() + "&username=" + asm.user);
+                        $("#button-send").button('enable');
+                        setTimeout(() => { header.show_info(_('Receipts emailed')); }, 850);
```
bobintetley: was the setTimeout necessary here? Why 850ms?

src/static/js/receipt_bulk.js
```diff
@@ -0,0 +1,119 @@
...
+                    } 
+                },
+                { type: "raw", markup: '<br>'},
+                { type: "raw", markup: '<div style="display: inline-block;vertical-align: top;">' + tableform.render_selectmulti({
+                        'id': 'paymentmethod',
+                        'justwidget': true,
+                        'options': { displayfield: "PAYMENTNAME", valuefield: "ID", rows: controller.paymentmethods },
```
bobintetley: quoted keys...

src/main.py
```diff
@@ -6978,6 +6978,82 @@ def post_vesignup(self, o):
...
+            if p:
+                pid = p['ID']
+                to = p['EMAILADDRESS']
+                if to in emailaddresses:
+                    emailaddresses[to].append(donationid)
+                else:
+                    emailaddresses[to] = [donationid]
+        count = len(emailaddresses.keys())
+        if count > asm3.configuration.mail_merge_max_emails(dbo):
+            raise asm3.utils.ASMError(f"{count} exceeds configured limit of {asm3.configuration.mail_merge_max_emails(dbo)} emails via mail merge")
```
bobintetley, line 7039: mail_merge_max_emails only applies to open source users - we need to do the checks for sheltermanager customers as well.

You can copy paste the code straight from mail merge that does the same job:

```
        if asm3.utils.is_smcom_smtp(dbo):
            asm3.smcom.check_bulk_email(dbo, count)
        elif len(rows) > asm3.configuration.mail_merge_max_emails(dbo):
            raise asm3.utils.ASMError(f"{count} exceeds configured limit of {asm3.configuration.mail_merge_max_emails(dbo)} emails via mail merge")
```

note that check_bulk_email does the work of adding to the total emails sent for the period, so there's nothing else to do.

src/main.py
```diff
@@ -6978,6 +6978,82 @@ def post_vesignup(self, o):
...
+        else:
+            for emailaddress in emailaddresses.keys():
```
bobintetley: there's no need for an else here - calling raise ends execution of the function, so the rest of the code doesn't need to be indented or in a block

src/asm3/financial.py
```diff
@@ -474,6 +480,29 @@ def get_donations_due_two_dates(dbo: Database, start: datetime, end: datetime) -
         "WHERE od.DateDue >= ? AND od.DateDue <= ? AND od.Date Is Null " \
         "ORDER BY od.DateDue DESC", (start, end))
 
+def get_donations_paid_two_dates(dbo: Database, start: datetime, end: datetime, paymentmethods: list) -> Results:
```
bobintetley: rather than using a typehint of list, you can use List[type] to say what we expect in the list. In this case it's a list of ints so ```List[int]```

src/asm3/financial.py
```diff
@@ -474,6 +480,29 @@ def get_donations_due_two_dates(dbo: Database, start: datetime, end: datetime) -
...
+    ql = [
+            get_donation_query(dbo),
+            'WHERE od.Date >= ? AND od.Date <= ?',
+            'AND ('
+        ]
+    pl = []
+    for pm in paymentmethods:
+        pl.append('DonationPaymentID = ?')
+        pmv.append(pm)
+    ql.append(' OR '.join(pl))
```
bobintetley: An IN clause would be easier here, so that the whole function is written something like:

```
methodids = ",".join([ str(x) for x in paymentmethods ])
sql = f"{get_donation_query(dbo)} WHERE od.DATE >= ? AND od.Date <= ? " \
    "AND DonationPaymentID IN ({methodids}) ORDER BY od.Date DESC"
return dbo.query(sql, [start, end])
```

src/static/js/receipt_bulk.js
adamdspencer77: No idea :) Removed quotes.

src/static/js/receipt_bulk.js
adamdspencer77: Unquoted.

src/static/js/receipt_bulk.js
adamdspencer77: It was from when I just disabled the button while it was busy. It was so fast that you couldn't tell anything had happened so I made it wait a while before re-enabling the button. Not necessary now the show_info is hooked up. Removed.

src/static/js/receipt_bulk.js
adamdspencer77: Unquoted.

src/main.py
adamdspencer77: Changed.

src/main.py
adamdspencer77: Hopefully sorted.

src/asm3/financial.py
adamdspencer77: Changed.

src/asm3/financial.py
adamdspencer77: Done.

PR #1825, opened 2025-07-23
SmartTag: new HTTP API

adamdspencer77:


src/asm3/publishers/smarttaghttp.py
```diff
@@ -0,0 +1,184 @@
...
+        firstname = an["CURRENTOWNERFORENAMES"].split(" ")[0]
+        
+        phones = []
+        for phone in (an["CURRENTOWNERHOMETELEPHONE"], an["CURRENTOWNERWORKTELEPHONE"], an["CURRENTOWNERMOBILETELEPHONE"]):
+            if phone:
+                phones.append(
+                    {
+                        "extension": "",
+                        "number": phone,
+                        "type": "UNKNOWN"
```
bobintetley: does this type correspond to one of work, home or cell? Could we set this? 

Looking at the code below, it doesn't look like this list phones[] is being used as the two phone fields are being set manually from CURRENTOWNERMOBILEPHONE etc?

src/asm3/publishers/smarttaghttp.py
```diff
@@ -0,0 +1,184 @@
...
+        elif species.find("Cat") != -1: species = "Feline"
+        elif species.find("Bird") != -1: species = "Avian"
+        elif species.find("Horse") != -1: species = "Equine"
+        elif species.find("Reptile") != -1: species = "Reptilian"
+        else: species = "Other"
+
+        address1 = an["CURRENTOWNERADDRESS"].split("\n")[0]
+        address2 = ""
+        if "\n" in an["CURRENTOWNERADDRESS"]:
+            address2 = an["CURRENTOWNERADDRESS"].split("\n")[1]
```
bobintetley: .split("\n", 1)[1] would not eat the 3rd line if there was one

src/asm3/publishers/smarttaghttp.py
adamdspencer77: That bit of code was a red herring. It is from the AVID publisher but not actually used. I've removed it.

PR #1828, opened 2025-07-31
animallocation: write entries for death and exit

adamdspencer77:


src/asm3/animal.py
```diff
@@ -3847,8 +3851,169 @@ def update_diary_linkinfo(dbo: Database, animalid: int, a: ResultRow = None, dia
...
+    entrydate = asm3.i18n.remove_time(animaldata["DATEBROUGHTIN"])
+    deceaseddate = asm3.i18n.remove_time(animaldata["DECEASEDDATE"])
+    animalname = animaldata["ANIMALNAME"]
+    sheltercode = animaldata["SHELTERCODE"]
+    releventmovementtypes = ['1', '3', '4', '5', '6', '7', '13']
+    if not asm3.configuration.foster_on_shelter(dbo):
+        releventmovementtypes.append('2')
+        releventmovementtypes.append('12')
+    if not asm3.configuration.trial_on_shelter(dbo):
+        releventmovementtypes.append('11')
```
bobintetley: 11 is a fake movement type that will never be set in the adoption table. Trial adoptions are still movement type 1, so you don't need this block

src/asm3/animal.py
```diff
@@ -3847,8 +3851,169 @@ def update_diary_linkinfo(dbo: Database, animalid: int, a: ResultRow = None, dia
...
+    ## Grab all data relevent to animal movement/death
+    animaldata = dbo.query("SELECT ShelterCode, AnimalName, DateBroughtIn, DeceasedDate FROM animal WHERE ID = ?", (animalid,))[0]
+    entrydate = asm3.i18n.remove_time(animaldata["DATEBROUGHTIN"])
+    deceaseddate = asm3.i18n.remove_time(animaldata["DECEASEDDATE"])
+    animalname = animaldata["ANIMALNAME"]
+    sheltercode = animaldata["SHELTERCODE"]
+    releventmovementtypes = ['1', '3', '4', '5', '6', '7', '13']
+    if not asm3.configuration.foster_on_shelter(dbo):
+        releventmovementtypes.append('2')
+        releventmovementtypes.append('12')
```
bobintetley: 12 is a fake movement type, permanent fosters are still always ID 2

src/asm3/animal.py
```diff
@@ -3847,8 +3851,169 @@ def update_diary_linkinfo(dbo: Database, animalid: int, a: ResultRow = None, dia
...
         dbo.execute("UPDATE diary SET LinkInfo = ? WHERE LinkType = ? AND LinkID = ?", (diaryloc, asm3.diary.ANIMAL, animalid))
 
+def update_animallocation(dbo: Database, animalid: int, username: str):
+    ## Grab all data relevent to animal movement/death
+    animaldata = dbo.query("SELECT ShelterCode, AnimalName, DateBroughtIn, DeceasedDate FROM animal WHERE ID = ?", (animalid,))[0]
+    entrydate = asm3.i18n.remove_time(animaldata["DATEBROUGHTIN"])
+    deceaseddate = asm3.i18n.remove_time(animaldata["DECEASEDDATE"])
+    animalname = animaldata["ANIMALNAME"]
+    sheltercode = animaldata["SHELTERCODE"]
+    releventmovementtypes = ['1', '3', '4', '5', '6', '7', '13']
```
bobintetley: "relevant" but I would shorten it to just movetypes or something - also, 13 will never be set on a record so can be left out of here

src/asm3/animal.py
```diff
@@ -3847,8 +3851,169 @@ def update_diary_linkinfo(dbo: Database, animalid: int, a: ResultRow = None, dia
...
+    movementrows = dbo.query(query, (animalid,))
+    animallocations = dbo.query("SELECT ID, Date, MovementID, IsDeath, ToLocationID, ToUnit, PrevAnimalLocationID FROM animallocation WHERE AnimalID = ? ORDER BY DATE desc", (animalid,))
+    ## Detect and remove impossible internal movements
+    indates = [entrydate]
+    outdates = []
+    if movementrows:
+        for row in movementrows:
+            outdates.append(asm3.i18n.remove_time(row["MOVEMENTDATE"]))
+            if row["RETURNDATE"]:
+                indates.append(asm3.i18n.remove_time(row["RETURNDATE"]))
```
bobintetley: the ReturnDate and MovementDate columns don't have a time component, so removing it is unnecessary

src/asm3/animal.py
```diff
@@ -3847,8 +3851,169 @@ def update_diary_linkinfo(dbo: Database, animalid: int, a: ResultRow = None, dia
...
+    if deceaseddate:
+        outdates.append(deceaseddate)
+    validrows = []
+    indate = indates[0]
+    indatecount = 0
+    outdatecount = 0
+    while indatecount < len(indates):
```
bobintetley: this could be written as a range for loop instead: ``` for indatecount in range(len(indates)):```

src/asm3/animal.py
```diff
@@ -3847,8 +3851,169 @@ def update_diary_linkinfo(dbo: Database, animalid: int, a: ResultRow = None, dia
...
+        outdate = None
+        for outdate in outdates[outdatecount:]:
+            if outdate >= indate:
+                outdatecount = outdatecount + 1
+                break
+        indatecount = indatecount + 1
+        for row in animallocations:
+            if outdate:
+                if row["MOVEMENTID"] == 0 and row["ISDEATH"] == 0 and asm3.i18n.remove_time(row["DATE"]) >= indate and asm3.i18n.remove_time(row["DATE"]) <= outdate:
```
bobintetley: access result rows via . instead of [""] - it makes it obvious that they're fields from result rows then, eg:

```if row.MOVEMENTID == 0 and row.ISDEATH == 0 and asm3.i18n.remove_time(row.DATE) ... ```

File: src/asm3/animal.py [hunk 7]
```diff
@@ -3847,8 +3851,169 @@ def update_diary_linkinfo(dbo: Database, animalid: int, a: ResultRow = None, dia
...
+                    if row not in validrows:
+                        validrows.append(row)
+            elif row["MOVEMENTID"] == 0 and row["ISDEATH"] == 0 and asm3.i18n.remove_time(row["DATE"]) >= indate:
+                    if row not in validrows:
+                        validrows.append(row)
+    for row in animallocations.copy():
```
bobintetley, line 3909: comment what this is doing, eg:

```# Iterate our list of valid rows and remove the ones that don't match, reparenting invalid PrevAnimalLocationID values as necessary```

File: src/asm3/animal.py [hunk 8]
```diff
@@ -3847,8 +3851,169 @@ def update_diary_linkinfo(dbo: Database, animalid: int, a: ResultRow = None, dia
...
+                (rowprevlocationid, animalid, rowid)
+            )
+    
+    ## Detect and delete obsolete external movements
+    obsoletemovements = []
+    for row in animallocations:
+        if row["MOVEMENTID"]:
+            movementfound = False
+            for movement in movementrows:
+                if row["MOVEMENTID"] == movement["ID"]:
```
bobintetley: Since this seems to be something you do a few times (getting a movement from your list of movement rows by its ID), I would make this an inner function that you can call to save a few loops, eg:

```
    def find_movement(id):
        for movement in movementrows:
            if id == movement.ID:
                return movement
        return None
```

You can then do ```if find_movement(row.MOVEMENTID) is None: obsoltemovements.append(row)``` here instead and below with the outboundmovement.

src/asm3/animal.py
```diff
@@ -3847,8 +3851,169 @@ def update_diary_linkinfo(dbo: Database, animalid: int, a: ResultRow = None, dia
...
+        dbo.delete("animallocation", row["ID"], username)
+        animallocations.remove(row)
+        rowid = row["ID"]
+        rowprevlocationid = row["PREVANIMALLOCATIONID"]
+        dbo.execute(
+            "UPDATE animallocation SET PrevAnimalLocationID = ? WHERE AnimalID = ? AND PrevAnimalLocationID = ?",
+            (rowprevlocationid, animalid, rowid)
+        )
+
+    ## Sync external movements
```
bobintetley, line 3927: More info on what this is doing and how it's doing it? Looks like the first part is checking and updating the date from movement to exit animal location, and adding a new one if it can't find one

src/asm3/animal.py
```diff
@@ -3847,8 +3851,169 @@ def update_diary_linkinfo(dbo: Database, animalid: int, a: ResultRow = None, dia
...
+                if asm3.i18n.remove_time(row["DATE"]) <= asm3.i18n.remove_time(movementrow["MOVEMENTDATE"]):
+                    mostrecentmovement = row
+                else:
+                    break
+            if mostrecentmovement:
+                fromid = mostrecentmovement["TOLOCATIONID"]
+                fromunit = mostrecentmovement["TOUNIT"]
+            insert_animallocation(dbo, username, animalid, animalname, sheltercode, fromid, fromunit, 0, '*', 0, movementrow["ID"])
+        
+        if movementrow["RETURNDATE"]:
```
bobintetley, line 3953: Ok, I'm trying to figure out what idmatch is about! It looks like the purpose of this section is to make sure there's an animallocation entry record if this movement is returned, creating one if it doesn't exist.

src/asm3/animal.py
```diff
@@ -3847,8 +3851,169 @@ def update_diary_linkinfo(dbo: Database, animalid: int, a: ResultRow = None, dia
...
+            idmatchfound = 0
+            for locationrow in animallocations:
+                if locationrow["MOVEMENTID"] == movementrow["ID"]:
+                    if idmatchfound == 0:
+                        idmatchfound = 1
```
bobintetley, line 3961: This idmatch thing is really confusing! I think it would be more straightforward if we look for the matching record, update it if we find and then create it if we didn't? Unless I missed something else this is doing

```
updated = False
locationrow = find_location(movementid=movementrow.ID)
if locationrow:
        ... update existing record
else:
    insert_animallocation ...
```    

see comment below for the find_location bit

src/asm3/animal.py
```diff
@@ -3847,8 +3851,169 @@ def update_diary_linkinfo(dbo: Database, animalid: int, a: ResultRow = None, dia
...
+            if idmatchfound < 2:
+                insert_animallocation(dbo, username, animalid, animalname, sheltercode, 0, '*', fromid, fromunit, movementid=movementrow["ID"], date=asm3.i18n.remove_time(movementrow["RETURNDATE"]))
+
+    ## Check that deceased date synced
+    if deceaseddate:
+        deathfound = False
+        fromid = 0
+        fromunit = ""
+        for locationrow in animallocations.copy():
+            if locationrow["ISDEATH"]:
```
bobintetley, line 3982: we need to find the death row and location rows quite a lot, these could be inner functions for readability and to save on some code:

```
def find_location(death=False, movementid=0):
    for l in animallocations:
        if movementid > 0 and l.MOVEMENTID = movementid: return l
        if death and l.ISDEATH == 1: return l
    return None
```

Also, there's no point iterating a copy above either because you don't seem to be modifying the list in this loop?

Makes this it just 
```
locationrow = find_location(death=True) 
if locationrow:
    update it
else:
    create it
```

src/asm3/animal.py
```diff
@@ -3847,8 +3851,169 @@ def update_diary_linkinfo(dbo: Database, animalid: int, a: ResultRow = None, dia
...
+            fromid = 0
+            insert_animallocation(dbo, username, animalid, animalname, sheltercode, fromid, fromunit, 0, '*', isdeath=1, date=deceaseddate)
+    else:
+        for locationrow in animallocations.copy():
+            if locationrow["ISDEATH"] == 1:
+                dbo.execute(
+                    "DELETE FROM animallocation WHERE ID = ?",
+                    (locationrow["ID"],)
+                )
+                animallocations.remove(locationrow)
```
bobintetley: same pattern as above with find_location - also, because you aren't iterating the list any more because the inner function did it, it's safe to delete from it without needing to make a copy!

src/asm3/animal.py
adamdspencer77: Removed.

src/asm3/animal.py
adamdspencer77: Removed.

src/asm3/animal.py
adamdspencer77: Done and done.

src/asm3/animal.py
adamdspencer77: Removal removed :)

src/asm3/animal.py
adamdspencer77: Changed.

src/asm3/animal.py

src/asm3/animal.py
adamdspencer77: Done.

src/asm3/animal.py

src/asm3/animal.py
adamdspencer77: Added more comments.

src/asm3/animal.py
adamdspencer77: It's counting how many rows in animallocation refer a particular movementid. An animal that has left the shelter should only have one but an animal that has been adopted and returned should have two. Hopefully new comments make this clearer.

src/asm3/animal.py
adamdspencer77: Done.

adamdspencer77: Didn't replace current loop following discussion. Needed to keep it here because it tracked previous location data needed to create missing rows.

src/asm3/animal.py
adamdspencer77: As the inner function didn't need to be created for conversation about I see no benefit from moving the loop into a function. I'm sure why I was removing locationrow from animallocations anyway. This is the end of the function and animallocations isn't referred to again.

PR #1832, opened 2025-07-31
Allow configuration of image scaling

adamdspencer77:


src/static/js/options.js
```diff
@@ -633,7 +633,18 @@ $(function() {
                             '<option value="{ownersurname}, {ownerforenames1} & {ownerforenames2}">' + _("Last, First & First") + '</option>' + 
                             '</select> '
                         }, 
-                        { id: "ownernameformat", post_field: "FirstDayOfWeek", label: _("When displaying calendars, the first day of the week is"), type: "select", options: html.list_to_options(['0|' + _("Sunday"), '1|' + _("Monday")]) }
+                        { id: "ownernameformat", post_field: "FirstDayOfWeek", label: _("When displaying calendars, the first day of the week is"), type: "select", options: html.list_to_options(['0|' + _("Sunday"), '1|' + _("Monday")]) },
+                        { id: "resizeimagesspec", post_field: "ResizeImagesSpec", label: _("Scale images to"), type: "select", options: html.list_to_options(
+                            [
+                                "(system)",
```
bobintetley: This should be a translatable string - the same one is used in change user settings

src/asm3/configuration.py
```diff
@@ -346,6 +346,7 @@
     "RecordSearchLimit": "1000",
     "ReloadMedical": "Yes",
     "ReportToolbar": "Yes",
+    "ResizeImagesSpec": "",
```
bobintetley: This doesn't need to be here - values that don't exist are empty string by default

src/asm3/media.py
```diff
@@ -73,6 +73,12 @@ def mime_type(filename: str) -> str:
         return types[ext]
     return "application/octet-stream"
 
+def get_resize_images_spec(dbo: Database):
+    scaleto = asm3.configuration.cstring(dbo, "ResizeImagesSpec")
```
bobintetley: Normal convention is to add a method to configuration.py called resize_images_spec() rather than calling cstring directly - that way defaults can be set/merged

src/static/js/options.js
adamdspencer77: Changed.

src/asm3/configuration.py

src/asm3/media.py
adamdspencer77: Changed.

src/static/js/options.js
```diff
@@ -633,7 +633,18 @@ $(function() {
...
+                        { id: "resizeimagesspec", post_field: "ResizeImagesSpec", label: _("Scale images to"), type: "select", options: html.list_to_options(
+                            [
+                                _("(use system)"),
+                                "1024x1024|1MP (1024px)",
+                                "1414x1414|2MP (1414px)",
+                                "1732x1732|3MP (1732px)",
+                                "2048x2048|4MP (2048px)",
+                                "2236x2236|5MP (2236px)",
+                                "2828x2828|8MP (2828px)",
+                                "4096z4096|16MP (4096px)"
```
bobintetley, line 646: typo, 4096z4096 instead of 4096x4096

PR #1833, opened 2025-08-03
bugfix(onlineform): Added canafforddonation to the list of values that can translate to DB fields for waiting list import

jacob-sabella:
It looks like most things were setup correctly for this happen correctly but the mapping to the "d" variable that is used to post to the database does not include the "canafforddonation" field. This prevented the value from being properly imported when creating a waiting list from an incoming form.


jacob-sabella:
@bobintetley please let me know if I need to add anything additional to this. Thank you.

PR #1836, opened 2025-08-04
Store comments in the log when they change

adamdspencer77:


src/static/js/options.js
```diff
@@ -726,6 +726,7 @@ $(function() {
...
                         { id: "animalnamechangelog", post_field: "AnimalNameChangeLog", label: _("When I change the name of an animal, make a note of it in the log with this type"), type: "check", xmarkup: ' <select data="AnimalNameChangeLogType" id="animalnamechangelogtype" class="asm-selectbox">' + html.list_to_options(controller.logtypes, "ID", "LOGTYPENAME") + '</select>' }, 
                         { id: "weightchangelog", post_field: "WeightChangeLog", label: _("When I change the weight of an animal, make a note of it in the log with this type"), type: "check", xmarkup: ' <select data="WeightChangeLogType" id="weightchangelogtype" class="asm-selectbox">' + html.list_to_options(controller.logtypes, "ID", "LOGTYPENAME") + '</select>' }, 
+                        { id: "animalcommentschangelog", post_field: "AnimalCommentsChangeLog", label: _("When I change the comments on an animal, make a note of it in the log with this type"), type: "check", xmarkup: ' <select data="AnimalCommentsChangeLogType" id="animalcommentschangelogtype" class="asm-selectbox">' + html.list_to_options(controller.logtypes, "ID", "LOGTYPENAME") + '</select>' },
```
bobintetley: The field is labeled "Description", so I'd make the translatable text "When I change the description or hidden comments on an animal, make a note of it in the log with this type"

src/asm3/animal.py
```diff
@@ -3892,6 +3895,18 @@ def insert_namechange_log(dbo: Database, username: str, animalid: int, newname:
             asm3.log.add_log(dbo, username, asm3.log.ANIMAL, animalid, asm3.configuration.animalname_change_log_type(dbo),
                 _("Name changed from '%s' to '%s'", l) % (oldname, newname))
 
+def insert_commentschange_log(dbo: Database, username: str, animalid: int, newcomments: str, oldcomments: str, newhiddencomments: str, oldhiddencomments: str) -> None:
+    """ Writes an entry to the log when an animal's comments/hidden comments change."""
+    # If the option is on and the name has changed, log it
```
bobintetley: redundant comment that can be removed

src/static/js/options.js
bobintetley: also, needs an entry in the manual in doc/manual/options.rst

adamdspencer77: Done.

src/asm3/animal.py
adamdspencer77: Removed.

PR #1837, opened 2025-08-06
Rota mobile interface

adamdspencer77:


src/main.py
```diff
@@ -1034,6 +1034,12 @@ def content(self, o):
         dbo = o.dbo
         animals = asm3.animal.get_shelterview_animals(dbo, o.lf)
         asm3.al.debug("mobile for '%s' (%s animals)" % (o.user, len(animals)), "main.mobile", dbo)
+        
+        userdata = asm3.users.get_user(dbo, o.user)
+        if userdata["OWNERID"]:
+            rotadata = asm3.person.get_rota(dbo, today(), add_days(today(), 7), userdata["OWNERID"])
+        else:
+            rotadata = False
```
bobintetley: Not keen on the mixing of return types? rotadata should be an empty list [] if there's no content. I tend to prefer to initialise vars first, so something like:

```
rotadata = []
if userdata["OWNERID"]:
        rotadata = asm3.person.get_rota(dbo, today(), add_days(today(), 7), userdata["OWNERID"])
```

PR #1838, opened 2025-08-06
Ability to disable home page alerts

adamdspencer77:


bobintetley:
Unless I'm missing something, I think maybe the issue wasn't clear on what was required? The job was to add a new config switch for every kind of alert that can be shown in main.js - so that they can be turned off individually if required. It looks like what's been done here is removing the configuration of some animal emblems hiding some alerts - which they should still do!

adamdspencer77:
Hopefully sorted now.

src/asm3/animal.py
```diff
@@ -1546,49 +1546,84 @@ def get_alerts(dbo: Database, lf: LocationFilter = None, age: int = 120) -> Resu
...
+    alertacuncomp = asm3.configuration.alert_ac_uncomp(dbo)
+    alertacfoll = asm3.configuration.alert_ac_foll(dbo)
+    alerttlover = asm3.configuration.alert_tl_over(dbo)
+    alertstexpsoon = asm3.configuration.alert_st_exp_soon(dbo)
+    alertstexp = asm3.configuration.alert_st_expired(dbo)
+    alertstlowbal = asm3.configuration.alert_st_low_bal(dbo)
+    alertgloballows = asm3.configuration.alert_global_lows(dbo)
+    alerttrnodrv = asm3.configuration.alert_tr_nodrv(dbo)
+    alertlngterm = asm3.configuration.alert_species_lng_term(dbo)
+    alertpublish = asm3.configuration.alert_publish(dbo)
```
bobintetley: Oh no! None of these functions are necessary and you don't have to check any of the yes/no values in  this query!

You can just use the boolean config options in the front end to hide the alert from the user when it's a yes/no rather than omitting it from the query output. Doesn't require a backend function at all then!

src/static/js/common_html.js
```diff
@@ -321,19 +321,16 @@ const html = {
             if (a.FLVRESULT == 2) { p.push(_("FLV+")); }
             s.push(html.icon("positivetest", p.join(" ")));
         }
-        if (config.bool("EmblemRabies") && !a.RABIESTAG && 
-            config.str("AlertSpeciesRabies").split(",").indexOf(String(a.SPECIESID)) != -1) {
+        if (config.bool("EmblemRabies") && !a.RABIESTAG) {
```
bobintetley: Isn't this change going to make it so the rabies emblem appears on all species of animals now where it didn't before?

src/static/js/common_html.js
```diff
@@ -321,19 +321,16 @@ const html = {
...
             s.push(html.icon("rabies", _("Rabies not given")));
         }
         if (config.bool("EmblemSpecialNeeds") && a.HASSPECIALNEEDS == 1) {
             s.push(html.icon("health", _("Special Needs")));
         }
-        if (config.bool("EmblemUnneutered") && a.NEUTERED == 0 && 
-            config.str("AlertSpeciesNeuter").split(",").indexOf(String(a.SPECIESID)) != -1) {
+        if (config.bool("EmblemUnneutered") && a.NEUTERED == 0) {
```
bobintetley: Same here, I might be misunderstanding, but this is going to show the not neutered emblem for all species of animals now?

src/static/js/common_html.js
```diff
@@ -321,19 +321,16 @@ const html = {
...
             s.push(html.icon("unneutered", _("Unaltered")));
         }
-        if (config.bool("EmblemNotMicrochipped") && a.IDENTICHIPPED == 0 && a.NONSHELTERANIMAL == 0 && 
-            config.str("AlertSpeciesMicrochip").split(",").indexOf(String(a.SPECIESID)) != -1) {
+        if (config.bool("EmblemNotMicrochipped") && a.IDENTICHIPPED == 0 && a.NONSHELTERANIMAL == 0) {
```
bobintetley: Yep, and same for microchips too!

src/static/js/common_html.js
adamdspencer77: Changed back.

src/static/js/common_html.js
adamdspencer77: Sorted.

src/static/js/common_html.js
adamdspencer77: Sorted.

PR #1843, opened 2025-08-11
Custom auto increment scheme

adamdspencer77:
Lots of people want to keep their own custom auto number schemes for things. The most recent one I've seen is for giving authorisation to vets (sort of vouchers, but not quite), but it's something a lot of people ask for.

Could we implement this as part of the additional field functionality? Add a new "Incrementing Number" type.

Like receipt numbers, we'd store the next value in the config table as AdditionalNextID (where ID is the additionalfieldid). This would allow the start position to be overridden where necessary. The field would probably include a button to retrieve the next number and increment the count.

### DIFF
--- a/src/static/js/header_additional.js
+++ b/src/static/js/header_additional.js
@@ -548,7 +548,7 @@ additional = {
         else if (f.FIELDTYPE == additional.TIME) { return tableform.render_time(v); }
         else if (f.FIELDTYPE == additional.NOTES) { v.classes += " asm-textareafixed"; return tableform.render_textarea(v); }
         else if (f.FIELDTYPE == additional.NUMBER) { return tableform.render_number(v); }
-        else if (f.FIELDTYPE == additional.NUMBER_INCREMENTED) { return tableform.render_number_incremented(v); }
+        else if (f.FIELDTYPE == additional.NUMBER_INCREMENTED) { v.xbutton = "<span class=\"ui-button-icon ui-icon ui-icon-refresh\" onclick='console.log($(this).parent().prev().val(\"Number\"))'></span>"; return tableform.render_number(v); }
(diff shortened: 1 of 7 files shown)


src/static/js/header_additional.js
```diff
@@ -547,6 +555,10 @@ additional = {
         else if (f.FIELDTYPE == additional.TIME) { return tableform.render_time(v); }
         else if (f.FIELDTYPE == additional.NOTES) { v.classes += " asm-textareafixed"; return tableform.render_textarea(v); }
         else if (f.FIELDTYPE == additional.NUMBER) { return tableform.render_number(v); }
+        else if (f.FIELDTYPE == additional.NUMBER_INCREMENTED) {
+            v.readonly = true;
+            return tableform.render_number(v);
+        }
```
bobintetley: Very very minor thing, I'd put these on one line to match the others

src/static/js/header_additional.js
```diff
@@ -183,6 +184,13 @@ additional = {
         $.each(fields, function(i, f) {
             if (f.NEWRECORD == 1) {
                 add.push(additional.render_field(f, includeids, classes, true));
+                if (f.FIELDTYPE == additional.NUMBER_INCREMENTED) {
+                    let formdata = { "mode": "nextid", "afid": f.ID };
+                    common.ajax_post("additional", formdata, function(response) {
+                        $("#add_" + f.ID).val(response);
```
bobintetley, line 191: Might be helpful to comment what we're doing here and why, that the ajax response will be coming back after render has completed and the content inserted into the DOM

src/static/js/header_additional.js
adamdspencer77: Done.

src/static/js/header_additional.js
adamdspencer77: Done.

PR #1844, opened 2025-08-12
Mobile UI not extracting the error message correctly

adamdspencer77:
If manual codes is turned on, and you add an animal through the mobile UI without supplying a code, the validation error is thrown and passed to the client, but the message is not parsed correctly and all that is shown to the user is a dialog with the word "error".

### DIFF
--- a/src/static/js/mobile.js
+++ b/src/static/js/mobile.js
@@ -36,7 +36,7 @@ const mobile = {
                 if (errorfunc) {
                     errorfunc(textstatus, response);
                 }
-                mobile.show_error(textstatus, response);
+                mobile.show_error(textstatus, jqxhr.responseText);
             }
         });
     },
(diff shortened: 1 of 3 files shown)


src/static/js/mobile.js
```diff
@@ -36,7 +36,7 @@ const mobile = {
                 if (errorfunc) {
                     errorfunc(textstatus, response);
                 }
-                mobile.show_error(textstatus, response);
+                mobile.show_error(textstatus, jqxhr.responseText);
```
bobintetley: jqxhr.responseText isn't always the message! Use common.get_error_response() to parse the message - see common.ajax_post for an example of how it works

adamdspencer77: Hopefully sorted.

PR #1846, opened 2025-08-19
Costs book/unpaid costs

adamdspencer77:
Some people use the cost paid date to track unpaid invoices (or invoices they are expecting that have not yet been received).

It's cumbersome using a report. Add a cost book screen with filters for Costs paid in the last 7 days, costs paid in the last month and unpaid costs under the Financial menu.

The menu option to open the screen should only appear if the option to track cost paid date is on.

### DIFF
--- a/src/static/js/animal_costs.js
+++ b/src/static/js/animal_costs.js
@@ -108,10 +108,11 @@ $(function() {
                                 }
                             },
                             onadd: function() {
+                                let response;
                                 if (controller.animal) {
-                                    var response = tableform.fields_post(dialog.fields, "mode=create&animal="  + controller.animal.ID, "animal_costs");
+                                    response = tableform.fields_post(dialog.fields, "mode=create&animal="  + controller.animal.ID, "animal_costs");
                                 } else {
-                                    var response = tableform.fields_post(dialog.fields, "mode=create", "animal_costs");
+                                    response = tableform.fields_post(dialog.fields, "mode=create", "animal_costs");
                                 }
                                 let row = {};
                                 row.ID = response;
@@ -123,12 +124,12 @@ $(function() {
                                     row.SHELTERCODE = animal_costs.lastanimal.SHELTERCODE;
                                     row.SHORTCODE = animal_costs.lastanimal.SHORTCODE;
                                 }
-                                let offset = parseInt($("#offset").val());
+                                let offset = $("#offset").val();
                                 let paiddate = row.COSTPAIDDATE;
                                 let datefloor = common.subtract_days(common.today_no_time(), offset);
                                 if (controller.animal) {
                                     controller.rows.push(row);
-                                } else if (!paiddate && !offset) {
+                                } else if (!paiddate && offset != '0') {
                                     controller.rows.push(row);
                                 } else if (format.date_js(paiddate, true) >= datefloor ) {
                                     controller.rows.push(row);


src/asm3/animal.py
```diff
@@ -4723,7 +4723,7 @@ def insert_cost_from_form(dbo: Database, username: str, post: PostedData) -> int
     if post.date("costdate") is None:
         raise asm3.utils.ASMValidationError(_("Cost date must be a valid date", l))
     ncostid = dbo.insert("animalcost", {
-        "AnimalID":         post.integer("animalid"),
+        "AnimalID":         post.integer("costanimalid"),
```
bobintetley: why the rename from animal to costanimalid? doesn't look like it conflicts with anything else and for consistency it's animal on every other screen?

src/asm3/animal.py
```diff
@@ -4741,6 +4741,7 @@ def update_cost_from_form(dbo: Database, username: str, post: PostedData) -> Non
     """
     costid = post.integer("costid")
     dbo.update("animalcost", costid, {
+        "AnimalID":         post.integer("costanimalid"),
```
bobintetley: ... animal

src/static/js/animal_costs.js
```diff
@@ -17,6 +18,7 @@ $(function() {
                 columns: 1,
                 width: 550,
                 fields: [
+                    { json_field: "ANIMALID", post_field: "costanimalid", label: _("Animal"), type: "animal", validation: "notzero" },
```
bobintetley: should be just animal rather than costanimalid

src/static/js/animal_costs.js
```diff
@@ -31,11 +33,25 @@ $(function() {
...
                 edit: async function(row) {
-                    await tableform.dialog_show_edit(dialog, row);
+                    await tableform.dialog_show_edit(dialog, row, {
+                        onload: function() {
+                            if (controller.animal) {
+                                $("#costanimalid").animalchooser("loadbyid", controller.animal.ID);
+                                $("#costanimalidrow").hide();
+                                animal_costs.costtype_change();
+                            } else {
+                                $("#costanimalidrow").show();
```
bobintetley: costanimalid -> animal

src/static/js/animal_costs.js
```diff
@@ -48,6 +64,16 @@ $(function() {
                         hideif: function() { return !config.bool("ShowCostPaid"); }
                     },
                     { field: "INVOICENUMBER", display: _("Invoice Number") },
+                    { field: "ANIMAL", display: _("Animal"),
```
bobintetley: doesn't really matter since you have a formatter, but I always liked the field to be a valid one from the controller, should be ANIMALID

src/static/js/animal_costs.js
```diff
@@ -63,14 +89,42 @@ $(function() {
...
+                                    $("#costanimalid").animalchooser("loadbyid", controller.animal.ID);
+                                    $("#costanimalidrow").hide();
+                                    animal_costs.costtype_change();
+                                } else {
+                                    $("#costanimalidrow").show();
+                                }
+                            }
+                        });
+                        if (controller.animal) {
+                            var response = await tableform.fields_post(dialog.fields, "mode=create&costanimalid="  + controller.animal.ID, "animal_costs");
```
bobintetley: always prefer let over var

src/static/js/animal_costs.js
```diff
@@ -63,14 +89,42 @@ $(function() {
...
+                            onload: function() {
+                                if (controller.animal) {
+                                    $("#costanimalid").animalchooser("loadbyid", controller.animal.ID);
+                                    $("#costanimalidrow").hide();
+                                    animal_costs.costtype_change();
+                                } else {
+                                    $("#costanimalidrow").show();
+                                }
+                            }
+                        });
```
bobintetley, line 141: think it's more consistent if we're using the on-x functions in a dict to not use await here with the dialog, and put your handler inside an onadd: function to make ajax the requests

src/static/js/animal_costs.js
```diff
@@ -31,11 +33,25 @@ $(function() {
...
+                        onload: function() {
+                            if (controller.animal) {
+                                $("#costanimalid").animalchooser("loadbyid", controller.animal.ID);
+                                $("#costanimalidrow").hide();
+                                animal_costs.costtype_change();
+                            } else {
+                                $("#costanimalidrow").show();
+                            }
+                        }
+                    });
```
bobintetley, line 64: for consistency, use onedit: function and you don't need to make the edit method async then, make your ajax calls from inside it

src/static/js/animal_costs.js
```diff
@@ -63,14 +89,42 @@ $(function() {
...
                         tableform.fields_update_row(dialog.fields, row);
                         row.COSTTYPENAME = common.get_field(controller.costtypes, row.COSTTYPEID, "COSTTYPENAME");
                         if (animal_costs.lastperson) { row.OWNERNAME = animal_costs.lastperson.OWNERNAME; }
-                        controller.rows.push(row);
+                        if (animal_costs.lastanimal) {
+                            row.ANIMALNAME = animal_costs.lastanimal.ANIMALNAME;
+                            row.SHELTERCODE = animal_costs.lastanimal.SHELTERCODE;
+                            row.SHORTCODE = animal_costs.lastanimal.SHORTCODE;
+                        }
+                        let offset = parseInt($("#offset").val());
```
bobintetley: does offset need to be parsed as an int? If the comparison is being done with == it will be coerced anyway and the value attribute of an option tag is a string. There's a well known gotcha with parseInt and assuming base 10 as well, use format.to_int instead if you need to parse integers from strings, but I suspect this could just be a ``` let offset = $("#offset").val(); ```

src/asm3/financial.py
```diff
@@ -305,6 +305,30 @@ def get_balance_fromto_date(dbo: Database, accountid: int, fromdate: datetime, t
     #if r.accounttype == INCOME or r.accounttype == EXPENSE: balance = abs(balance)
     return balance
 
+def get_costs(dbo: Database, offset: int = 0, sort: int = ASCENDING) -> Results:
+    """
+    Returns all animalcost records:
+    COSTTYPEID, COSTTYPENAME, COSTDATE, DESCRIPTION, OWNERID, INVOICENUMBER, ANIMALID, ANIMALNAME, SHORTCODE, SHELTERCODE
+    """
```
bobintetley, line 314: needs a comment explaining what offset does and how the behaviour changes when offset == 0

src/static/js/animal_costs.js
```diff
@@ -234,15 +324,20 @@ $(function() {
         },
```
bobintetley, line 340: there are some widgets that need to be destroyed here too:

```
common.widget_destroy("#animal");
common.widget_destroy("#person");
animal_costs.lastanimal = null;
animal_costs.lastperson = null;
```

src/asm3/animal.py
adamdspencer77: Yeah, I'd originally gone for animalid because that's what it returned but it was already taken on animal records. Changed to animal.

src/asm3/animal.py

src/static/js/animal_costs.js
adamdspencer77: Done.

src/static/js/animal_costs.js
adamdspencer77: Done.

src/static/js/animal_costs.js
adamdspencer77: Done.

src/static/js/animal_costs.js
adamdspencer77: If I switch the vars on lets in this if block then I can't refer to outside the if block when row.ID is declared just after the if. I guess I could move the row.ID declaration inside the if but this would involve duplicating some code if this is preferable.

src/static/js/animal_costs.js
adamdspencer77: This was the variable that tripped me up with it's truthiness. Before I added the parseInt it tripped up when it got to `} else if (!paiddate && !offset) {` 

I could change this to `} else if (!paiddate && !parseInt(offset)) {` if you think that's neater.

src/static/js/animal_costs.js
adamdspencer77: Don't really understand the mechanism but copied the structure from citations.js and seems to work :)

src/static/js/animal_costs.js
adamdspencer77: Hopefully sorted. When I used onedit I got a hang with no error messages. I couldn't find onedit: in the code base but found onchange, used that and all seems well. Want to check that this is what you had in mind before I resolve.

src/asm3/financial.py
adamdspencer77: Done.

src/static/js/animal_costs.js
adamdspencer77: Hopefully sorted. I popped them to the exiting destroy function above tableform.dialog_destroy(); Everything still seems to work.

PR #1852, opened 2025-08-24
Kombu Integration for Event Messaging

jacob-sabella:
This is basically just me taking the [kombu](https://github.com/celery/kombu/) library and passing the inserts of the audit log records to the service before the existing logic inserts into the tables. I tried to keep it off the primary thread to prevent it from being a requirement to respond to the UI. 

This works for me connecting to an Azure service bus Queue, but let me know if you need something different for your hosting situation, especially if the message volume or reliability requirements are different or if the current scope is too broad for the volume of messages sent. Not really expecting this to be merged in as is but wanted to start a conversation (unless you are cool with it this way).


bobintetley:
Yep, looks great, thanks for doing this!

PR #1853, opened 2025-08-26
Mail merge: add custom log

adamdspencer77:
Under the send email section of mail merges, if the field ID or OWNERID is present, allow the option to add a custom log message and type to everyone who was sent the email.

Many people have asked for this as they'd like to know when people have received their newsletters, etc. It also solves a problem for SAFE in Australia as they automatically send out unsuccessful applicant messages and would like to exclude people who have already been sent it in response to their last application.

On reflection, not sure about this, while it's popular with many, it could create a LOT of log records very fast (ironically, it would be very slow to add a thousand log records in one go).

A better solution is a way to link the same log to multiple records. Eg: with a new column containing a list of IDs or a separate table.

A separate logmulti table would be the answer for these. As well as the log message itself, it would need columns for LogTypeID, LinkTypeID and then a comma or pipe separated list of LinkIDs.

Since these messages are in a different table, it's easy enough to UNION them for the screens, but by default reports that output logs would not include them. I'm not sure if that's an issue since it's really only people that we would normally be considering for this at the moment. Print Person Record probably should include them.

### DIFF
--- a/src/asm3/log.py
+++ b/src/asm3/log.py
@@ -37,6 +37,20 @@ def add_log_email(dbo: Database, username: str, linktype: int, linkid: int, logt
     return add_log(dbo, username, linktype, linkid, logtypeid,
         "[%s] %s ::\n%s" % ( to, subject, body ))
 
+def add_logmulti(dbo: Database, username: str, linktype: int, linkids: list, logtypeid: int, logtext: str, logdatetime: datetime = None) -> int:
+    """
+    Adds a log entry. If logdatetime is blank, the date/time now is used.
+    """
+    linkidsstring = ",".join(linkids)
+    if logdatetime is None: logdatetime = dbo.now()
+    return dbo.insert("logmulti", {
+        "LogTypeID":        logtypeid,
+        "LinkIDs":          linkidsstring,
+        "LinkType":         linktype,
+        "Date":             logdatetime,
+        "Comments":         logtext
+    }, username)
+
 def get_log_find_simple(dbo: Database, q: str, limit: int = 0) -> Results:
     """
     Searches log notes for the term q
--- a/src/asm3/log.py
+++ b/src/asm3/log.py
@@ -39,9 +39,9 @@ def add_log_email(dbo: Database, username: str, linktype: int, linkid: int, logt
 
 def add_logmulti(dbo: Database, username: str, linktype: int, linkids: list, logtypeid: int, logtext: str, logdatetime: datetime = None) -> int:
     """
-    Adds a log entry. If logdatetime is blank, the date/time now is used.
+    Adds a logmulti entry. If logdatetime is blank, the date/time now is used.
     """
-    linkidsstring = ",".join(linkids)
+    linkidsstring = ",".join([str(linkid) for linkid in linkids] )
     if logdatetime is None: logdatetime = dbo.now()
     return dbo.insert("logmulti", {
         "LogTypeID":        logtypeid,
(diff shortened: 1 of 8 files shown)


src/asm3/utils.py
```diff
@@ -2175,11 +2177,18 @@ def do_send():
...
+                    linkids.append(r.OWNERID)
+                elif "ID" in r:
+                    linkids.append(r.ID)
             if "EMAILADDRESS2" in r: 
                 toadd = r.EMAILADDRESS2
                 if toadd is None or toadd.strip() == "": continue
                 asm3.al.debug("sending bulk email: to=%s, subject=%s" % (toadd, ssubject), "utils.send_bulk_email", dbo)
                 send_email(dbo, replyadd, toadd, "", "", ssubject, sbody, contenttype, exceptions=False, bulk=True)
+        if createlog:
+                asm3.log.add_logmulti(dbo, username, asm3.log.PERSON, linkids, logtypeid, logmessage)
```
bobintetley: I think doing this here is adding a circular dependency - by doing this, asm3.log imports asm3.utils, and asm3.utils imports asm3.log

Instead, I think creation of the logmulti record should be done in the mailmerge endpoint in main.py after calling send_bulk_email

adamdspencer77: Hopefully sorted.

src/asm3/utils.py
```diff
@@ -2151,7 +2151,7 @@ def _send_email(msg: MIMEMultipart, fromadd: str, tolist: List[str], dbo: Databa
             time.sleep(RETRY_SECS)
             _send_email(msg, fromadd, tolist, dbo=dbo, exceptions=exceptions, retries=retries-1)
 
-def send_bulk_email(dbo: Database, replyadd: str, subject: str, body: str, rows: Results, contenttype: str, unsubscribe: bool) -> None:
+def send_bulk_email(dbo: Database, replyadd: str, subject: str, body: str, rows: Results, contenttype: str, unsubscribe: bool, createlog: bool = False, logtypeid: int = 0, logmessage: str = '', username: str = '') -> None:
```
bobintetley: can't tell from the diff, looks like these extra, but no longer used parameters are still here and need removing?

src/static/js/mailmerge.js
```diff
@@ -114,39 +114,50 @@ $(function() {
...
+                                    return true;
+                                }
+                            }
+                        },
+                        { id: 'em-logmessage', post_field: 'logmessage', label: _("Log message"), type: 'textarea', validation: 'notblank',
+                            hideif: function() {
+                                if ( controller.fields.includes("ID") || controller.fields.includes("OWNERID") ) {
+                                    return false;
+                                } else {
+                                    return true;
```
bobintetley: For readability, I'd put this code inside an extra function - mailmerge.log_unavailable() just in case it changes in future or if other areas need to test it.

You can then just have ```hideif: mailmerge.log_unavailable``` for these 3 UI elements.

src/asm3/utils.py
adamdspencer77: Removed.

src/static/js/mailmerge.js
adamdspencer77: Done.

src/asm3/log.py
```diff
@@ -37,6 +37,20 @@ def add_log_email(dbo: Database, username: str, linktype: int, linkid: int, logt
     return add_log(dbo, username, linktype, linkid, logtypeid,
         "[%s] %s ::\n%s" % ( to, subject, body ))
 
+def add_logmulti(dbo: Database, username: str, linktype: int, linkids: list, logtypeid: int, logtext: str, logdatetime: datetime = None) -> int:
```
bobintetley, line 40: Not really an issue, but rather than using "list" as the type hint, you can use the capitalised version of list and dict that specifies what you expect the list or dict to contain. Eg: ``` linkids: List[int] ```

PR #1859, opened 2025-09-01
Tokens/links in comments

adamdspencer77:
The original ticket was this:

    It'd be cool if we could allow tokens in any of ASM's text fields. The same code that does things like updating href targets could find instances of #animalname, #sheltercode or @PersonCode @person_name and turn them into clickable links to the animal or owner.

It could actually be simpler than that. Use the hash character followed by a type indicator and colon then a term to indicate a link token. We don't need separate ones for animals and people, by having an "s" token that sends the term to the global search, it covers all bases, eg:

#s:a:animalname
#s:sheltercode
#s:personcode
#s:p:personname
#s:li:12345

The other thing we get asked for a LOT is to be able to link to a piece of media from text. We could do this with a #m: token.

The value for the media would need to be its ID, so we need to show a column containing the ID in the table view of media for the user to be able to copy it. When resolved to a link, it will just go to media?id=X - nothing else is needed, that endpoint can handle the mimetype and will use a 24 hour cache.

#m:1234

Since these are simple rewrites to wrap an a tag around them, this could be done by a tableform column formatter.

These links will always open in a new tab.

### DIFF
--- a/src/static/js/common_tableform.js
+++ b/src/static/js/common_tableform.js
@@ -160,13 +160,13 @@ const tableform = {
         if (v.includes('#s:')) {
             tokenfound = true;
             v = v.replace(/#s:\w{1,}:?\w{1,}/g, function(t) {
-                return '<a href="/search?q=' + t.replace(/#s:/, '') + '" target="_blank">' + t + '</a>';
+                return '<a href="/search?q=' + t.replace('#s:', '') + '" target="_blank">' + t + '</a>';
             });
         }
         if (v.includes('#m:')) {
             tokenfound = true;
             v = v.replace(/#m:\w{1,}/g, function(t) {
-                return '<a href="/media?id=' + t.replace(/#m:/, '') + '" target="_blank">' + t + '</a>';
+                return '<a href="/media?id=' + t.replace('#m:', '') + '" target="_blank">' + t + '</a>';
             });
         }
         if (config.bool("ShowFullCommentsInTables") || tokenfound) { return common.nulltostr(v).replace(/\n/g, "<br />"); }


src/static/js/common_widgets.js
```diff
@@ -62,7 +62,9 @@ $.fn.fromJSON = function(row) {
             n.richtextarea("value", row[f]);
         }
         else if (n.is("textarea")) {
-            n.html(row[f]);
+            n.textarea("value", row[f]);
+            n.change();
```
bobintetley: is n.change() necessary here since you're assigning the value?

src/static/js/common_widgets.js
```diff
@@ -1830,14 +1831,69 @@ $.widget("asm.textarea", {
...
+                    } else if (v.includes('co:')) {
+                        linkiconclass = 'asm-icon-cost';
+                    } else if (v.includes('lo:')) {
+                        linkiconclass = 'asm-icon-log';
+                    } else if (v.includes('vo:')) {
+                        linkiconclass = 'asm-icon-transactions';
+                    } else if (v.includes('ci:')) {
+                        linkiconclass = 'asm-icon-citation';
+                    }
+                    $('#' + tdid).append('<div class="asm-token-link"><span class="asm-icon ' + linkiconclass + '"></span><a href=/search?q=' + v + ' target="_blank">' + v + '</a></div> ');
```
bobintetley: href attribute isn't quoted and should be

src/static/js/common_tableform.js
```diff
@@ -2002,8 +2002,9 @@ const tableform = {
                 // Unescaped tags in textareas behave unpredictably
                 var s = row[v.json_field];
                 if (!s) { s = ""; }
-                s = s.replace(/</g, "&lt;").replace(/>/g, "&gt;");
-                n.val(html.decode(s));
+                n.textarea("value", s);
+                n.textarea('process_tokens');
+                n.change();
```
bobintetley: can't process_tokens call change if necessary? Also, for consistency I always use double quotes when quoting strings, the only time I use single quotes is if the string contains HTML, so I can use double quotes for HTML attributes

src/static/js/common_widgets.js
```diff
@@ -1830,14 +1831,69 @@ $.widget("asm.textarea", {
...
         });
 
+        t.on('paste keyup', function() {
+            self.process_tokens();
+        });
+        
+
+    },
+
+    process_tokens: function() {
```
bobintetley: think process_links might be a more descriptive name than process_tokens?

src/static/js/common_widgets.js
adamdspencer77: It was already in the existing code here. Moved it into the new textarea.value function.

src/static/js/common_widgets.js

src/static/js/common_widgets.js
adamdspencer77: Changed.

src/static/js/common_tableform.js
adamdspencer77: change() now called by textarea.value so removed from here. Single quotes swapped for double.

bobintetley:
I've realised now that this isn't the full job (and why I mentioned table formatters). 

The second part of this job is updating tableform.format_comments - this function outputs comment fields in tables and it should also be substituting these tokens for the links to the correct place too. This time, it should be a simple inline text replace with an <a> tag with target so that the clickable links appear in the table output.

src/static/js/common_tableform.js
```diff
@@ -156,7 +156,20 @@ const tableform = {
 
     /** Formats a value as comments (truncates to one line or shows full with \n -> <br/> based on config) */
     format_comments: function(row, v) {
-        if (config.bool("ShowFullCommentsInTables")) { return common.nulltostr(v).replace(/\n/g, "<br />"); }
+        let tokenfound = false;
+        if (v.includes('#s:')) {
+            tokenfound = true;
+            v = v.replace(/#s:\w{1,}:?\w{1,}/g, function(t) {
+                return '<a href="/search?q=' + t.replace(/#s:/, '') + '" target="_blank">' + t + '</a>';
```
bobintetley: Think this would be better as ``` t.replace("#s:", "") ```

Using /#s:/ will use regex when it's not needed here.

src/static/js/common_tableform.js
```diff
@@ -156,7 +156,20 @@ const tableform = {
...
+            });
+        }
+        if (v.includes('#m:')) {
+            tokenfound = true;
+            v = v.replace(/#m:\w{1,}/g, function(t) {
+                return '<a href="/media?id=' + t.replace(/#m:/, '') + '" target="_blank">' + t + '</a>';
```
bobintetley: same as above

src/static/js/common_tableform.js
adamdspencer77: Done.

src/static/js/common_tableform.js
adamdspencer77: Done.

PR #1864, opened 2025-09-03
service: online_form_js to embed online forms in the current page

adamdspencer77:
As an alternative to online_form_html, have an online_form_js method. This returns javascript code that injects the form (as generated by onlineform.py/get_onlineform_html) into the current page at a provided div, a little like adoptable_js.

The mechanism by which this should work is basically by writing an iframe tag into the supplied div. It outputs the iframe tag and then does the same clever thing that something like pym.js does to size that iframe dynamically to take up the same amount of space in the page as the content so there's no scrolling and it's less obvious an iframe is involved. This neatly avoids all of the potential problems with conflicting js libraries, reused global names, etc.


src/static/js/onlineform_embed.js
```diff
@@ -0,0 +1,30 @@
...
+//       with as many browsers as possible. It also does not use jQuery.
+//       Avoid use of let/const, async/await, destructuring, etc.
+
+(function() {
+    var div_id = "asm3-onlineform";
+
+    var render_iframe = function() {
+        var hostdiv = document.getElementById(div_id);
+        hostdiv.innerHTML = '<iframe id="asm3-form-iframe" scrolling="no" style="width: 100%;border: 0;height: auto;"></iframe>';
+        document.getElementById("asm3-form-iframe").src = 'http://localhost:5000/service?method=online_form_html&formid={TOKEN_FORMID}';
```
bobintetley: Accidental hard-coding here - this should contain a {SERVICE_URL} token to be substituted by the backend that serves this file.

```.src = '{SERVICE_URL}?method=online_form_html&formid={TOKEN_FORMID}'```

File: scripts/rollup/rollup.py [hunk 2]
```diff
@@ -11,7 +11,7 @@
     "common_animalchoosermulti.js", "common_personchooser.js", "common_tableform.js", "common_barcode.js", "common_microchip.js", "header.js",
     "header_additional.js", "header_edit_header.js" ]
 
-exclude = [ "animal_view_adoptable.js", "document_edit.js", "onlineform_extra.js", "report_toolbar.js" ]
+exclude = [ "animal_view_adoptable.js", "document_edit.js", "onlineform_extra.js", "report_toolbar.js", "onlineform_embed.js" ]
```
bobintetley: be nicer in alphabetical order!

src/asm3/html.py
```diff
@@ -46,7 +46,7 @@ def is_standalone_js(filename: str) -> bool:
     """
     Returns true if this is a standalone js file
     """
-    standalone = [ "animal_view_adoptable.js", "document_edit.js", "onlineform_extra.js", "report_toolbar.js" ]
+    standalone = [ "animal_view_adoptable.js", "document_edit.js", "onlineform_extra.js", "report_toolbar.js", "onlineform_embed.js" ]
```
bobintetley: should be alphabetical :)

src/asm3/onlineform.py
```diff
@@ -374,9 +374,22 @@ def get_onlineform_html(dbo: Database, formid: int, completedocument: bool = Tru
...
+            'ro.observe(document.querySelector("html"));',
+            '</script>'
+        ])
+        h.append(footer.replace("$$TITLE$$", form.NAME).replace("</body>", extrajs + "</body>"))
     return "\n".join(h)
 
+def get_onlineform_js(dbo: Database, formid: int) -> str:
+    """ Returns js that outputs online form into a host div """
+    js = asm3.utils.read_text_file("%s/static/js/onlineform_embed.js" % dbo.installpath)
+    js = js.replace("{TOKEN_FORMID}", str(formid))
```
bobintetley, line 395: need to substitute SERVICE_URL here as well

src/static/js/onlineform_embed.js
adamdspencer77: Hopefully sorted.

src/asm3/html.py
adamdspencer77: Done.

scripts/rollup/rollup.py
adamdspencer77: Done.

src/asm3/onlineform.py
adamdspencer77: Done.

src/static/js/onlineform_embed.js
```diff
@@ -0,0 +1,30 @@
+/*global alert */
+/*global asm3_adoptable_filters, asm3_adoptable_iframe, asm3_adoptable_iframe_height, asm3_adoptable_iframe_bgcolor, asm3_adoptable_iframe_fixed */
+/*global asm3_adoptable_translations, asm3_adoptable_extra, asm3_adoptable_filter, asm3_adoptable_limit, asm3_adoptable_sort, asm3_adoptable_style */
```
bobintetley: I think these globals got copied from animal_view_adoptable.js - they aren't valid here as this file doesn't seem to use any globals

src/asm3/onlineform.py
```diff
@@ -374,9 +374,22 @@ def get_onlineform_html(dbo: Database, formid: int, completedocument: bool = Tru
...
+            '   window.parent.postMessage(document.querySelector("html").offsetHeight, "*");',
+            '});',
+            'ro.observe(document.querySelector("html"));',
+            '</script>'
+        ])
+        h.append(footer.replace("$$TITLE$$", form.NAME).replace("</body>", extrajs + "</body>"))
     return "\n".join(h)
 
+def get_onlineform_js(dbo: Database, formid: int) -> str:
+    """ Returns js that outputs online form into a host div """
```
bobintetley: Returns js that outputs a responsive online form iframe into a host div!

src/asm3/onlineform.py
```diff
@@ -374,9 +374,22 @@ def get_onlineform_html(dbo: Database, formid: int, completedocument: bool = Tru
...
+    js = asm3.utils.read_text_file("%s/static/js/onlineform_embed.js" % dbo.installpath)
+    js = js.replace('{SERVICE_URL}', SERVICE_URL).replace('{TOKEN_FORMID}', str(formid))
```
bobintetley: tiny readability thing but I'd do the replaces on separate lines - and use double quotes

src/asm3/onlineform.py
```diff
@@ -374,9 +374,22 @@ def get_onlineform_html(dbo: Database, formid: int, completedocument: bool = Tru
     if completedocument:
         h.append(asm3.utils.nulltostr(form.FOOTER))
         footer = get_onlineform_footer(dbo)
-        h.append(footer.replace("$$TITLE$$", form.NAME))
+        extrajs = '\n'.join(['<script>',
```
bobintetley: Think I prefer just using string literal joiner since there's no extra benefit to doing it this way as there are no variables to substitute or include, eg:

```
extrajs = '<script>\n' \
    'let ro = new ResizeObserver(function(e) {' \
    '    window.parent.postMessage(document.querySelector("htmlk").offsetHeight, "*");' \
    '});' \
    ...
```

src/static/js/onlineform_embed.js
adamdspencer77: Removed.

src/asm3/onlineform.py

src/asm3/onlineform.py

src/asm3/onlineform.py

PR #1866, opened 2025-09-08
Contextual reports

adamdspencer77:
User suggestion, on the animal and person details screens have a "Reports" dropdown next to Documents.

The new dropdown should filter/show all reports with a single A S K A N I M A L or A S K P E R S O N tag in their source (new get_reports query).

The target link for each one is the normal link for the report with the extra parameters &mode=exec&ASK1=[ID]

This allows reports for the currently viewed record to be run without being asked for one.

### DIFF
--- a/src/asm3/reports.py
+++ b/src/asm3/reports.py
@@ -58,8 +58,19 @@ def get_ask_animal_reports(dbo: Database, superuser: int, roleids: str):
             filteredreports.append(report)
     return filteredreports
 
-def get_ask_person_reports(dbo: Database):
-    return dbo.query(f"SELECT ID, Title, Category, Revision FROM customreport WHERE {dbo.sql_ilike("SQLCommand", "?")} ORDER BY Title", ['%$ask person$%'])
+def get_ask_person_reports(dbo: Database, superuser: int, roleids: str):
+    filteredreports = []
+    allreports = dbo.query(f"SELECT ID, Title, Category, Revision FROM customreport WHERE {dbo.sql_ilike("SQLCommand", "?")} ORDER BY Title", ['%$ask person$%'])
+    roles = dbo.query("SELECT * FROM customreportrole")
+    for report in allreports:
+        viewroleids = []
+        for role in roles:
+            if role.REPORTID == report.ID and role.CANVIEW == 1:
+                viewroleids.append(str(role.ROLEID))
+        report.VIEWROLEIDS = "|".join(viewroleids)
+        if superuser or report.VIEWROLEIDS == "" or asm3.utils.list_overlap(report.VIEWROLEIDS.split("|"), roleids.split("|")):
+            filteredreports.append(report)
+    return filteredreports
 
 def get_available_reports(dbo: Database, include_with_criteria: bool = True) -> Results:
     """
--- a/src/asm3/reports.py
+++ b/src/asm3/reports.py
@@ -44,11 +44,22 @@ def get_all_report_titles(dbo: Database):
     """
     return dbo.query("SELECT ID, Title, Category, Revision FROM customreport WHERE SQLCommand NOT LIKE '0%' AND SQLCommand NOT LIKE '%$PARENT%' ORDER BY Title")
 
-def get_ask_animal_reports(dbo: Database):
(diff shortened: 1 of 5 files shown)


src/asm3/reports.py
```diff
@@ -44,6 +44,12 @@ def get_all_report_titles(dbo: Database):
     """
     return dbo.query("SELECT ID, Title, Category, Revision FROM customreport WHERE SQLCommand NOT LIKE '0%' AND SQLCommand NOT LIKE '%$PARENT%' ORDER BY Title")
 
+def get_ask_animal_reports(dbo: Database):
+    return dbo.query(f"SELECT ID, Title, Category, Revision FROM customreport WHERE {dbo.sql_ilike("SQLCommand", "?")} ORDER BY Title", ['%$ask animal%'])
```
bobintetley: wants the $ on the end of the pattern, otherwise this will pick up ```$ASK ANIMALS$``` reports, which we don't want

src/static/js/header_edit_header.js
```diff
@@ -822,6 +822,20 @@ edit_header = {
...
     },
 
+    // List of reports that contain '$ask animal'
+    report_list: function(reports, animalid) {
+        var s = [];
+        var lastcategory = "";
+        $.each(reports, function(i, r) {
+            if (r.CATEGORY != lastcategory) {
+                s.push('<li class="asm-menu-category">' + r.CATEGORY + '</li>');
+                lastcategory = r.CATEGORY;
```
bobintetley, line 832: we also need to check if the current user has permission to run this report - reports can have an optional list of roles (if there are none, then all users can run it)

src/static/js/person.js
```diff
@@ -235,6 +236,10 @@ $(function() {
                 edit_header.template_list(controller.templates, "PERSON", controller.person.ID),
                 '</ul>',
                 '</div>',
+                '<div id="button-report-body" class="asm-menu-body">',
+                edit_header.report_list(controller.reports, controller.person.ID),
```
bobintetley: Unless I'm missing something, your report_list function only accepts an animal ID?

src/asm3/reports.py

src/static/js/header_edit_header.js
adamdspencer77: Hopefully sorted.

src/static/js/person.js
adamdspencer77: Hopefully sorted.

src/main.py
```diff
@@ -1876,6 +1876,7 @@ def controller(self, o):
             "pickuplocations": asm3.lookups.get_pickup_locations(dbo),
             "publishhistory": asm3.animal.get_publish_history(dbo, a.ID),
             "posneg": asm3.lookups.get_posneg(dbo),
+            "reports": asm3.reports.get_ask_animal_reports(dbo, o['session'].superuser, o['session'].roleids),
```
bobintetley: little teeny thing, I really prefer double quotes

src/main.py
```diff
@@ -2859,7 +2860,7 @@ def post_all(self, o):
         quicklinks = post["quicklinksid"]
         quickreports = post["quickreportsid"]
         quickreportscfg = []
-        reports = asm3.reports.get_reports(o.dbo)
+        reports = asm3.reports.Mail(o.dbo)
```
bobintetley: What is Mail? It's not a member of reports.py and it doesn't seem to be in your changes? I'm not sure this change is relevant to this ticket either since this bit of code is updating quick reports

src/main.py
```diff
@@ -6317,6 +6318,7 @@ def controller(self, o):
             "homecheckhistory": asm3.person.get_homechecked(dbo, p.id),
             "jurisdictions": asm3.lookups.get_jurisdictions(dbo),
             "logtypes": asm3.lookups.get_log_types(dbo),
+            "reports": asm3.reports.get_ask_person_reports(dbo, o['session'].superuser, o['session'].roleids),
```
bobintetley: same again with the single quotes

src/main.py
adamdspencer77: Changed.

src/main.py
adamdspencer77: No idea why I would have changed anything here. I think I've put it back as it was before.

src/main.py
adamdspencer77: Changed.

PR #1871, opened 2025-09-16
Allow diary/calendar colour coding

adamdspencer77:
The calendarview.js code now supports bgcol and fgcol in the events sent to it.

Add a new integer field to the diary table for colour. This value is an index to an internal table of background/foreground colour values that the user can choose from when creating their diary note. The two values will be sent to calendarview.js for that event so that users can effectively colour code their diary notes.

It is also possible that we could have some fixed colours for the different types of calendar elements, such as medical items, clinic, etc.

### DIFF
--- a/src/asm3/diary.py
+++ b/src/asm3/diary.py
@@ -383,7 +383,7 @@ def insert_diary_from_form(dbo: Database, username: str, linktypeid: int, linkid
         email_note_on_change(dbo, get_diary(dbo, diaryid), username)
     return diaryid
 
-def insert_diary(dbo: Database, username: str, linktypeid: int, linkid: int, diarydate: datetime, diaryfor: str, subject: str, note: str, emailnow: bool = False) -> int:
+def insert_diary(dbo: Database, username: str, linktypeid: int, linkid: int, diarydate: datetime, diaryfor: str, subject: str, note: str, emailnow: bool = False, colourschemeid: int = 0) -> int:
     """
     Creates a diary note from the form data
     username: User creating the diary
@@ -402,6 +402,7 @@ def insert_diary(dbo: Database, username: str, linktypeid: int, linkid: int, dia
         "LinkInfo":         linkinfo,
         "DiaryDateTime":    diarydate,
         "DiaryForName":     diaryfor,
+        "ColourSchemeID":   colourschemeid,
         "Subject":          subject,
         "Note":             note
     }, username)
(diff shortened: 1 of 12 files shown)


src/asm3/diary.py
```diff
@@ -372,6 +372,7 @@ def insert_diary_from_form(dbo: Database, username: str, linktypeid: int, linkid
         "LinkInfo":         linkinfo,
         "DiaryDateTime":    post.datetime("diarydate", "diarytime"),
         "DiaryForName":     post["diaryfor"],
+        "ColourSchemeID":   post["diarycolourscheme"],
```
bobintetley: SQLite lets you get away with using the wrong types for parameters to insert/update queries. This will fail on postgresql since colourschemeid is an integer - should be post.integer("diarycolourscheme")

src/asm3/diary.py
```diff
@@ -431,6 +433,7 @@ def update_diary_from_form(dbo: Database, username: str, post: PostedData) -> No
     dbo.update("diary", diaryid, {
         "DiaryDateTime":    post.datetime("diarydate", "diarytime"),
         "DiaryForName":     post["diaryfor"],
+        "ColourSchemeID":   post["diarycolourscheme"],
```
bobintetley: same again, should be an integer

src/asm3/diary.py
```diff
@@ -510,6 +515,7 @@ def insert_diarytaskdetail_from_form(dbo: Database, username: str, post: PostedD
         "OrderIndex":           post.integer("orderindex"),
         "DayPivot":             post.integer("pivot"),
         "WhoFor":               post["for"],
+        "ColourSchemeID":       post["diarycolourscheme"],
```
bobintetley: post.integer

src/asm3/diary.py
```diff
@@ -523,6 +529,7 @@ def update_diarytaskdetail_from_form(dbo: Database, username: str, post: PostedD
...
```
bobintetley: post.integer!

src/asm3/lookups.py
```diff
@@ -90,6 +90,42 @@
...
+    "test": ["#48B3AF", "#476EAE"],
+    "boarding": ["#0BA6DF", "#EF7722"],
+    "clinic": ["#84994F", "#FFE797"],
+    "payment": ["#FCB53B", "#B45253"],
+    "incident": ["#476EAE", "#48B3AF"],
+    "transport": ["#EF7722", "#0BA6DF" ],
+    "equipment": ["#FFFFFF",  "#000000"],
+    "event": ["#331B3F", "#ACC7B4" ]
+}
+
```
bobintetley, line 128: Do these match the existing colours as they were before?

bobintetley (commented):
I think changes to dbupdate.py and a dbupdates script are missing from the pull request? ColourSchemeID needs to be added to the diary and diarytaskdetail tables right?

src/asm3/diary.py
adamdspencer77: Done.

src/asm3/diary.py
adamdspencer77: Done.

src/asm3/diary.py
adamdspencer77: Done.

src/asm3/diary.py
adamdspencer77: Done.

src/asm3/lookups.py
adamdspencer77: I've set them to blank which falls back the original existing colours.

PR #1872, opened 2025-09-16
Microchip Registrations with AVID USA

adamdspencer77:


src/asm3/configuration.py
```diff
@@ -112,6 +112,7 @@
     "AvidReRegistration": "No", 
     "AvidRegisterOverseas": "No",
     "AvidOverseasOriginCountry": "",
+    "AvidUSRegisterOverseas": "No",
```
bobintetley: Was this copied from the UK one? I'm presuming that Avid US does not have the same requirement to be able to register any type of chip selectively do they? This doesn't seem to be used in the publisher, so I think it can be removed.

src/asm3/configuration.py
```diff
@@ -823,6 +824,18 @@ def avid_overseas_origin_country(dbo: Database) -> str:
...
+    return cstring(dbo, "AVIDUSUsername")
+
+def avidus_password(dbo: Database) -> str:
+    return cstring(dbo, "AVIDUSPassword")
+
+def avidus_apikey(dbo: Database) -> str:
+    return cstring(dbo, "AVIDUSKey")
+
+def avidus_register_overseas(dbo: Database) -> bool:
+    return cboolean(dbo, "AvidUSRegisterOverseas", DEFAULTS["AvidUSRegisterOverseas"] == "Yes")
```
bobintetley: Redundant and doesn't need to be here

src/asm3/publishers/avid.py
```diff
@@ -0,0 +1,217 @@
...
+
+        avidusername = asm3.configuration.avidus_username(self.dbo)
+        avidpassword = asm3.configuration.avidus_password(self.dbo)
+        avidkey = asm3.configuration.avidus_apikey(self.dbo)
+
+        #
+        # Authentication Header
+        #
+        basic_auth_str = f'{avidusername}:{avidpassword}'
+        basic_auth_header = base64.b64encode(basic_auth_str.encode()).decode()
```
bobintetley: I try to avoid using Python libs directly because the interface on them has changed in the past. If you use our functions in the utils.py module instead, it will handle converting between strings for you.

```
basic_auth_header = asm3.utils.base64encode(basic_auth_str)
```

src/asm3/sitedefs.py
```diff
@@ -322,6 +323,9 @@ def get_dict(k: str, dv: Dict = {}) -> Dict:
 VETENVOY_US_HOMEAGAIN_RECIPIENTID = get_string("vetenvoy_us_homeagain_recipientid", "")
 VETENVOY_US_AKC_REUNITE_RECIPIENTID = get_string("vetenvoy_us_akc_reunite_recipientid", "")
 
+# AVID US Microchip resitration details
+AVID_US_API_KEY = get_string("avid_us_api_key", "iE#LD3Y(BVU&eW#xDqkzxLzPswALboU)")
```
bobintetley: This key shouldn't be here - it should be blank with you setting it in your asm3.conf

I hope this isn't a sensitive key since our repo is public!

src/asm3/publishers/avid.py
```diff
@@ -0,0 +1,217 @@
...
+                self.log("Processing: %s: %s (%d of %d)" % ( an["SHELTERCODE"], an["ANIMALNAME"], anCount, len(animals)))
+                self.updatePublisherProgress(self.getProgress(anCount, len(animals)))
+                # If the user cancelled, stop now
+                if self.shouldStopPublishing(): 
+                    self.stopPublishing()
+                    return
+                if not self.validate(an): continue
+                fields = self.processAnimal(an)
+                r = asm3.utils.post_json(AVID_US_POST_URL, asm3.utils.json(fields), config['headers'])
+                if r['response'] and asm3.utils.json_parse(r["response"])["status"] == 'COMPLETE':
```
bobintetley: mixed quoting types, always use " if possible. ' should only be used where you have a literal that contains " and you don't want to faff around

src/asm3/configuration.py
adamdspencer77: Yeah, I left the overseas stuff in originally. Removed.

src/asm3/configuration.py

src/asm3/publishers/avid.py
adamdspencer77: Yeah my bad. I think I'd used this early on but then switched it to copy their example when I was trouble shooting. Changed.

src/asm3/publishers/avid.py
adamdspencer77: Sorted.

src/asm3/sitedefs.py
adamdspencer77: Sorted.

src/static/js/publish_options.js
```diff
@@ -179,6 +179,15 @@ $(function() {
                                 '<option value="1">Do not generate and upload (not recommended)</option>', 
                                 classes: 'pbool preset' }
                         ]}, 
+                    { id: "tab-avidus", title: "AVID US Microchips", classes: 'localeus',
```
bobintetley: there should be a tag class for hasavid or something here so that it can be hidden on non-sheltermanager installations

src/asm3/sitedefs.py
```diff
@@ -281,6 +281,7 @@ def get_dict(k: str, dv: Dict = {}) -> Dict:
 ANIBASE_BASE_URL = get_string("anibase_base_url", "")
 ANIBASE_API_USER = get_string("anibase_api_user", "")
 ANIBASE_API_KEY = get_string("anibase_api_key", "")
+AVID_US_POST_URL = get_string("avid_us_post_url", "https://api.avidid.com/v1/registrations/")
```
bobintetley: This should be blank, and this being blank sets hasavid to false in the config.js endpoint like other publishers. The URL should only be in the asm3.conf file rather than the source

src/static/js/publish_options.js
adamdspencer77: Hopefully sorted.

PR #1877, opened 2025-09-22
online forms: Person home check fields

adamdspencer77:
Performing home checks could be a useful function for online forms. Add the following fields

homecheckpass - any non-empty and non-zero value qualifies as a yes and sets the homechecked flag on the target person and sets the homechecked date to today

homecheckedby - a person name that is partially matched against owner.OwnerName for all people with the homechecker flag to find who did the homecheck. Should be straightforward for shelters to create a lookup of person names for them to pick from

Manual needs updating to reflect these new fields.


doc/manual/onlineformfields.rst
```diff
@@ -33,6 +33,10 @@ mobiletelephone / celltelephone / mobiletelephone2 / celltelephone2
     The person's mobile number
 emailaddress / emailaddress2
     The person's email address
+homecheckpass
+    If a person has passed a homecheckpass
```
bobintetley: a homecheckpass? :D 

I would mention what type of field should be used, and what values the backend is looking for in order to set the homechecked flag on the person

src/asm3/person.py
```diff
@@ -1048,6 +1048,13 @@ def insert_person_from_form(dbo: Database, post: PostedData, username: str, geoc
     pid = dbo.get_id("owner")
     ownercode = post["ownercode"]
     if post["ownercode"] == "": ownercode = calculate_owner_code(pid, post["surname"])
+    homecheckedby = 0
+    if post["homecheckedby"]:
+        ilikename = dbo.sql_ilike("OwnerName", "?")
+        homecheckedby = dbo.query_int(
+            f"SELECT ID FROM owner WHERE IsHomeChecker = 1 AND (UPPER(OwnerCode) = ? OR {ilikename})",
+            [post["homecheckedby"].upper(), f"%{post["homecheckedby"]}%"]
+        )
```
bobintetley: This code should be in onlineform. It's the responsibility of create_person in onlineform.py to resolve the homechecker and set the correct ID value for homecheckedby.

src/asm3/person.py
```diff
@@ -1145,7 +1152,10 @@ def insert_person_from_form(dbo: Database, post: PostedData, username: str, geoc
             post["flags"] += ",excludefrombulkemail"
 
     # Update the flags
-    update_flags(dbo, username, pid, post["flags"].split(","))
+    flags = post["flags"].split(",")    
+    if post["homecheckpass"]:
+        flags.append("homechecked")
```
bobintetley: this should also be done in onlineform.py / create_person - it should be passing a "flags" element in the dictionary that contains the string "homechecked "

src/asm3/person.py
```diff
@@ -1464,7 +1474,17 @@ def merge(dictfield, fieldname):
...
     uv["OwnerName"] = calculate_owner_name(dbo, p.OWNERTYPE, p.OWNERTITLE, p.OWNERINITIALS, p.OWNERFORENAMES, p.OWNERSURNAME, "", "", "", \
                         p.OWNERTITLE2, p.OWNERINITIALS2, p.OWNERFORENAMES2, p.OWNERSURNAME2)
+    if "homecheckpass" in d.keys() and d["homecheckpass"]:
+        update_pass_homecheck(dbo, username, personid, comments="")
+    if "homecheckedby" in d.keys():
+        ilikename = dbo.sql_ilike("OwnerName", "?")
+        uv["HomeCheckedBy"] = dbo.query_int(
+            f"SELECT ID FROM owner WHERE IsHomeChecker = 1 AND (UPPER(OwnerCode) = ? OR {ilikename})",
+            [d["homecheckedby"].upper(), f"%{d["homecheckedby"]}%"]
+        )
```
bobintetley: This code shouldn't be here either, and it's repeating the same stuff from insert_person_from_form that doesn't belong there too!

doc/manual/onlineformfields.rst
adamdspencer77: Done.

src/asm3/person.py
adamdspencer77: Hopefully sorted.

src/asm3/person.py
adamdspencer77: Hopefully sorted. Had to work around some existing flag handling code.

src/asm3/person.py
adamdspencer77: Sorted.

src/asm3/person.py
```diff
@@ -1090,6 +1097,7 @@ def insert_person_from_form(dbo: Database, post: PostedData, username: str, geoc
         "FosterCapacity":   post.integer("fostercapacity"),
         "HomeCheckAreas":   post["areas"],
         "DateLastHomeChecked": post.date("homechecked"),
+        # "HomeCheckedBy":    homecheckedby,
```
bobintetley, line 1101: not sure this comment should be here?

src/asm3/person.py
```diff
@@ -1145,6 +1153,10 @@ def insert_person_from_form(dbo: Database, post: PostedData, username: str, geoc
             post["flags"] += ",excludefrombulkemail"
 
     # Update the flags
+    # flags = post["flags"].split(",")    
+    # if post["homecheckpass"]:
+    #     flags.append("homechecked")
+    # update_flags(dbo, username, pid, flags)
```
bobintetley, line 1160: probably shouldn't be here either!

PR #1879, opened 2025-09-24
Show more linked cost information on accounting transactions

adamdspencer77:
The accounts/transactions screen shows info about related payments, but it doesn't show much about related costs. In particular, had one customer who wants to see the invoice number from the linked cost. We could include a payee link the way we include the person link for payments.


src/asm3/financial.py
```diff
@@ -377,7 +377,11 @@ def get_transactions(dbo: Database, accountid: int, datefrom: datetime, dateto:
     elif reconciled == NONRECONCILED:
         recfilter = " AND Reconciled = 0"
     rows = dbo.query("SELECT t.*, srcac.Code AS SrcCode, destac.Code AS DestCode, " \
-        "o.OwnerName AS PersonName, o.ID AS PersonID, a.ID AS DonationAnimalID, " \
+        "CASE WHEN oac.OwnerName IS NOT NULL THEN oac.OwnerName " \
+        "ELSE o.OwnerName END AS PersonName, " \
+        "CASE WHEN oac.ID IS NOT NULL THEN oac.ID " \
```
bobintetley: For clarity, the rest of this query has separate fields for cost and payment, would it not make sense to do the same thing here?

ie. Have separate columns for DonationOwnerID, DonationOwnerName, CostOwnerID, CostOwnerName ? The UI could then just choose the non-blank ones and no need for the case statement

src/static/js/accounts_trx.js
```diff
@@ -129,6 +135,12 @@ $(function() {
                     }
                     desc += '</span>';
                 }
+                if (t.ANIMALCOSTID) {
```
bobintetley: There's always going to be a costanimalid if there's an animalcostid, would it make sense to merge those two conditions into one?

src/asm3/financial.py
adamdspencer77: Hopefully sorted.

src/static/js/accounts_trx.js
adamdspencer77: Sorted.

src/asm3/financial.py
```diff
@@ -394,7 +396,9 @@ def get_transactions(dbo: Database, accountid: int, datefrom: datetime, dateto:
         "aca.AnimalName AS CostAnimalName, aca.ID AS CostAnimalID, " \
         "CASE " \
         "WHEN EXISTS(SELECT ItemValue FROM configuration WHERE ItemName Like 'UseShortShelterCodes' AND ItemValue = 'Yes') " \
-        "THEN aca.ShortCode ELSE aca.ShelterCode END AS CostAnimalCode " \
+        "THEN aca.ShortCode ELSE aca.ShelterCode END AS CostAnimalCode, " \
+        # "oac.OwnerName AS ACOwnerName " \
```
bobintetley, line 400: errant comment?

Issue #862, opened 2020-06-14
Costs book/unpaid costs


bobintetley:
Some people use the cost paid date to track unpaid invoices (or invoices they are expecting that have not yet been received). 

It's cumbersome using a report. Add a cost book screen with filters for Costs paid in the last 7 days, costs paid in the last month and unpaid costs under the Financial menu.

The menu option to open the screen should only appear if the option to track cost paid date is on.

PR #1881, opened 2025-09-25
Show costs tab on suppliers and vets

adamdspencer77:
Suggestion from a user, they use the costs tab quite a lot on their animals, they'd like to be able to see costs from the person side as well.

Add a cost tab to person records, but only display it if the person has the vet or supplier flag. It could allow adding as well, like the other screens do with the person hidden and animal displayed for editing. Doing this work will make it easier to do the ticket about adding a costs book in future https://github.com/sheltermanager/asm3/issues/862


src/static/js/person_costs.js
```diff
@@ -0,0 +1,190 @@
+/*global $, _, asm, common, config, controller, dlgfx, edit_header, format, header, html, tableform, validate */
+
+$(function() {
+
+    "use strict";
+    const person_costs = {
```
bobintetley: This whole file shouldn't need to exist? 

The way it's done with payments and movements and other screens it that there's a single javascript file that handles all flavours of the screen. 

You should be able to use animal_costs.js for the UI and modify it slightly based on what mode it is in by adding routes.

src/main.py
```diff
@@ -6490,6 +6490,39 @@ def controller(self, o):
             "rows": donations
         }
 
+class person_costs(JSONEndpoint):
+    url = "person_costs"
```
bobintetley, line 6494: ```js_module = "animal_costs"``` to use the existing UI.

weirdly, I've just looked and I must have created a cost_book target already many years ago as it's already there in the code. Just needs a menu option and verifying it works.

src/main.py
```diff
@@ -6490,6 +6490,39 @@ def controller(self, o):
...
+        asm3.al.debug("got %d costs for person %s" % (len(costs), p["OWNERNAME"]), "main.person_costs", dbo)
+        return {
+            "name": "person_costs",
+            "rows": costs,
+            "person": p,
+            "costtypes": asm3.lookups.get_costtypes(dbo),
+            "tabcounts": asm3.person.get_satellite_counts(dbo, personid)[0]
+        }
+
+    def post_create(self, o):
```
bobintetley: the animal_costs.js UI will continue to post to the animal_costs endpoint for create/update/delete, so no need for these.

src/static/js/header_edit_header.js
```diff
@@ -652,6 +653,9 @@ edit_header = {
             if (perms && !common.has_permission(perms)) { return; } // don't show if no permission
             if ((key == "boarding") && config.bool("DisableBoarding")) { return; }
             if ((key == "citation" || key == "licence" || key == "investigation") && config.bool("DisableAnimalControl")) { return; }
+
+            if (key == "costs" && (!controller.person.ISVET && !controller.person.ISSUPPLIER)) { return; }
+
```
bobintetley: fine, but why the whitespace around this one line?

src/asm3/person.py
```diff
@@ -152,6 +152,23 @@ def embellish_latest_movement(dbo: Database, p: ResultRow) -> ResultRow:
         p.LATESTMOVETYPENAME = lm.LATESTMOVETYPENAME
     return p
 
+def get_costs(dbo: Database, personid: int, sort: int = ASCENDING) -> Results:
```
bobintetley: This should probably be in animal.py since that's where the other costs methods live (I could see an argument for moving them to financial.py). I'd call it get_costs_for_payee or something as well.

src/static/js/person_costs.js
adamdspencer77: Hopefully sorted.

src/main.py
adamdspencer77: Done.

src/main.py
adamdspencer77: Removed.

src/static/js/header_edit_header.js
adamdspencer77: It was just to make it easy to find while I was working on it :) Tidied.

src/asm3/person.py

PR #1893, opened 2025-10-07
Autocomplete additional field type

adamdspencer77:
It'd be nice to offer autocomplete as an additional field type (a textbox with lookup values).

Even easier now we have the autotext widget.

### DIFF
--- a/src/static/js/additional.js
+++ b/src/static/js/additional.js
@@ -142,8 +142,8 @@ $(function() {
           */
         check_type: function() {
             let ft = $("#type").select("value");
-            // Show lookups if field type is yes/no, lookup or multi-lookup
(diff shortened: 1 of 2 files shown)


src/asm3/dbupdates/35103.py
```diff
@@ -0,0 +1,4 @@
+from asm3.dbupdate import execute
+
+# Add the new autocomplete additional field type
+execute(dbo, "INSERT INTO lksfieldtype (ID, FieldType) VALUES (?, ?)", [ 16, _("Autocomplete", dbo.locale) ])
```
bobintetley: this should be in the default data function in asm3/dbupdate.py as well :)

adamdspencer77: Done.

PR #1894, opened 2025-10-08
Option to send reports as PDF

adamdspencer77:
At least one customer wants to be able to send their scheduled reports as PDFs rather than HTML in the body of the message. This could be an option on the report itself.

### DIFF
--- a/src/asm3/dbupdates/35104.py
+++ b/src/asm3/dbupdates/35104.py
@@ -1,5 +1,4 @@
 from asm3.dbupdate import add_column, add_index, execute
 
 add_column(dbo, "customreport", "DailyEmailSendAsPDF", dbo.type_integer)
-add_index(dbo, "customreport_DailyEmailSendAsPDF", "customreport", "DailyEmailSendAsPDF")
 execute(dbo, "UPDATE customreport SET DailyEmailSendAsPDF = 0")
--- a/src/asm3/dbupdate.py
+++ b/src/asm3/dbupdate.py
@@ -959,13 +959,13 @@ def sql_structure(dbo: Database) -> str:
         flongstr("DailyEmail", True),
         fint("DailyEmailHour", True),
         fint("DailyEmailFrequency", True),
+        fint("DailyEmailSendAsPDF", True),
         flongstr("SQLCommand", False),
         flongstr("HTMLBody", False),
         flongstr("Description"),
         fint("OmitHeaderFooter"),
         fint("OmitCriteria"),
-        fint("Revision", True),
-        fint("SendAsPDF") ))
+        fint("Revision", True) ))
     sql += index("customreport_Title", "customreport", "Title")
 
     sql += table("customreportrole", (


src/asm3/dbupdates/35104.py
```diff
@@ -0,0 +1,5 @@
+from asm3.dbupdate import add_column, add_index, execute
+
+add_column(dbo, "customreport", "SendAsPDF", dbo.type_integer)
+add_index(dbo, "customreport_SendAsPDF", "customreport", "SendAsPDF")
```
bobintetley: I think the field would be better named as DailyEmailSendAsPDF since it's related to the other 3 daily email columns? Other than that this looks good

src/asm3/reports.py
```diff
@@ -297,7 +297,8 @@ def insert_report_from_form(dbo: Database, username: str, post: PostedData) -> i
         "Description":          post["description"],
         "OmitHeaderFooter":     post.boolean("omitheaderfooter"),
         "OmitCriteria":         post.boolean("omitcriteria"),
-        "Revision":             post.integer("revision")
+        "Revision":             post.integer("revision"),
+        "SendAsPDF":            post.boolean("sendaspdf")
```
bobintetley: Put this near the other DailyEmail fields in the list rather than on the end!

src/asm3/reports.py
```diff
@@ -321,7 +322,8 @@ def update_report_from_form(dbo: Database, username: str, post: PostedData) -> N
         "DailyEmailFrequency":  post.integer("dailyemailfrequency"),
         "Description":          post["description"],
         "OmitHeaderFooter":     post.boolean("omitheaderfooter"),
-        "OmitCriteria":         post.boolean("omitcriteria")
+        "OmitCriteria":         post.boolean("omitcriteria"),
+        "SendAsPDF":            post.boolean("sendaspdf")
```
bobintetley: same, near the other related fields

src/asm3/reports.py
```diff
@@ -771,7 +773,12 @@ def email_daily_reports(dbo: Database) -> None:
         if body.find("NODATA") != -1 and not asm3.configuration.email_empty_reports(dbo): 
             asm3.al.debug("report '%s' contained no data and option is on to skip sending empty reports" % (r.TITLE), "reports.email_daily_reports", dbo)
             continue
-        asm3.utils.send_email(dbo, "", emails, "", "", r.TITLE, body, "html", exceptions=False, retries=3)
+        if r.SENDASPDF:
+            pdfdata = asm3.utils.html_to_pdf(dbo, body)
+            body = asm3.i18n._('Please find your report attached.', dbo.locale)
+            asm3.utils.send_email(dbo, "", emails, "", "", r.TITLE, body, "html", exceptions=False, retries=3, attachments=[ f"{asm3.i18n._('Report', dbo.locale)}.pdf", "application/pdf", pdfdata])
```
bobintetley: I don't think using the translatable string for Report for the filename is a good idea. Filenames will end up in email headers, which I'm not sure fully support unicode. I'd just use the ascii "report.pdf" instead since it's guaranteed to work everywhere

src/static/js/reports.js
```diff
@@ -83,6 +83,7 @@ $(function() {
                     { json_field: "DAILYEMAILFREQUENCY", post_field: "dailyemailfrequency", label: _("When"), type: "select", options: html.list_to_options(emailfreq, "ID", "DISPLAY") },
                     { json_field: "DAILYEMAILHOUR", post_field: "dailyemailhour", label: _("at"), type: "select",
                         options: { displayfield: "display", valuefield: "value", rows: emailhours }},
+                        { json_field: "SENDASPDF", post_field: "sendaspdf", label: _("Send in PDF format"), type: "check" },
```
bobintetley: indent looks weird

src/asm3/dbupdates/35104.py
adamdspencer77: Done.

src/asm3/reports.py

src/asm3/reports.py

src/asm3/reports.py

src/static/js/reports.js
adamdspencer77: Sorted.

PR #1898, opened 2025-10-09
Assign role to multiple users

adamdspencer77:
Have a dropdown list of roles that allows the same role to be assigned to all selected users on the users screen.

### DIFF
--- a/src/asm3/users.py
+++ b/src/asm3/users.py
@@ -11,7 +11,7 @@ import asm3.smcom
 import asm3.utils
 
 from asm3.sitedefs import BASE_URL
-from asm3.typehints import Database, PostedData, ResultRow, Results, Session
+from asm3.typehints import Database, List, PostedData, ResultRow, Results, Session
 
 import os
 import sys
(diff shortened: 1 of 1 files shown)


src/main.py
```diff
@@ -7920,6 +7920,11 @@ def post_create(self, o):
     def post_update(self, o):
         self.check(asm3.users.EDIT_USER)
         asm3.users.update_user_from_form(o.dbo, o.user, o.post)
+    
+    def post_addrole(self, o):
+        self.check(asm3.users.EDIT_USER)
+        for uid in o.post.integer_list("ids"):
+            o.dbo.insert("userrole", { "UserID": uid, "RoleID": o.post.integer("roleid") }, generateID=False)
```
bobintetley: This will throw an exception if the user already has this role and stop processing. 

I don't like calling database functions directly in the endpoint, there should be a function in users.py to add a role to a user if one doesn't already exist that should take care of error handling etc.

src/static/js/users.js
```diff
@@ -255,13 +262,45 @@ $(function() {
...
             $("#siterow").toggle( config.bool("MultiSiteEnabled") );
             validate.indicator([ "newpassword", "confirmpassword" ]);
+            $.each($("#button-assignrole-body li"), function(i, r) {
+                $(r).on("click", async function() {
+                    let userids = tableform.table_ids(this.table);
+                    let roleid = $(this).attr("data-roleid");
+                    await common.ajax_post("systemusers", "mode=addrole&ids=" + userids + "&roleid=" + roleid);
+                    $.each(userids.split(","), function(i, uid) {
+                        if (uid) {
+                            $.each(controller.rows, function(i, row) {
```
bobintetley, line 287: I think this processing should all be done at the backend - just send the list of users and the role to the endpoint, let it call the function in users.py that adds the role to the user and skips ones the user already has.

src/main.py
adamdspencer77: Hopefully sorted.

src/static/js/users.js
adamdspencer77: Hopefully sorted.

PR #1900, opened 2025-10-14
adoptapet: Remove redundant images

adamdspencer77:
Given how infrequently this happens, and that the issue tends to be filling up of quota, I think we should do the following:

-     Make a new maint_adoptapet endpoint that logs into their adoptapet account and deletes all of the images.
-     See maint_petfinder for a similar calling mechanism, you can just add a new removeAllImages method to the adoptapet publisher and call it from the endpoint. It should output each image it has removed and a total as confirmation.

Created endpoint. Nothing tested and lots in here that I don't understand.

### DIFF
--- a/src/asm3/publishers/adoptapet.py
+++ b/src/asm3/publishers/adoptapet.py
@@ -489,13 +489,11 @@ class AdoptAPetPublisher(FTPPublisher):
     
     def removeAllImages(self) -> str:
         """
-        This process sends a blank file to Adoptapet
-        to remove all the existing listings (so any existing images are forgotten). It
-        then touches the access date on all the publishable images for adoptable animals
-        so that PetFinder will get a new URL for them and download them again.
-        The return value is the publishing log.
+        Log in to the FTP account with AdoptAPet and remove all images
+        Useful when an account has gone over their 800MB quota. This is called 
+        by the maint_adoptapet endpoint.
+        Return value is the publishing log for display.
         """
-                
         self.log("AdoptAPetPublisher clearing listings ...")
 
         shelterid = asm3.configuration.adoptapet_user(self.dbo)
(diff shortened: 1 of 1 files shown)


src/asm3/publishers/adoptapet.py
```diff
@@ -486,4 +486,37 @@ def processAnimal(self, an: ResultRow) -> str:
...
+            raise Exception("No adoptapet.com shelter id has been set.")
 
+        if not self.openFTPSocket(): 
+            raise Exception("Failed opening FTP socket.")
+
+        # Touch all the date fields of publishable media for adoptable animals 
+        # so that they get a different photo URL due to the timestamp so PetFinder will 
+        # retrieve them again
+        self.dbo.execute("UPDATE media SET Date=? WHERE LinkTypeID=0 AND LinkID IN (SELECT ID FROM animal WHERE Adoptable=1) AND ExcludeFromPublish=0", 
+            [ self.dbo.now() ])
```
bobintetley: It isn't necessary to do this in this function, all we need to do is delete the images from the FTP server.

src/asm3/publishers/adoptapet.py
```diff
@@ -486,4 +486,37 @@ def processAnimal(self, an: ResultRow) -> str:
...
+        """
+                
+        self.log("AdoptAPetPublisher clearing listings ...")
+
+        shelterid = asm3.configuration.adoptapet_user(self.dbo)
+        if shelterid == "":
+            raise Exception("No adoptapet.com shelter id has been set.")
 
+        if not self.openFTPSocket(): 
+            raise Exception("Failed opening FTP socket.")
```
bobintetley, line 504: need to call ``` self.chdir("photos", "photos") ``` after opening the socket or you'll be deleting from the wrong folder

src/asm3/publishers/adoptapet.py
```diff
@@ -486,4 +486,37 @@ def processAnimal(self, an: ResultRow) -> str:
...
+        self.dbo.execute("UPDATE media SET Date=? WHERE LinkTypeID=0 AND LinkID IN (SELECT ID FROM animal WHERE Adoptable=1) AND ExcludeFromPublish=0", 
+            [ self.dbo.now() ])
+        
+        try:
+            nlst = self.socket.nlst("*.jpg")
+            for f in nlst:
+                self.log("delete image: %s" % f)
+                self.socket.delete(f)
+        except Exception as err:
+            self.logError("warning: failed deleting from FTP server: %s" % err, sys.exc_info())
```
bobintetley, line 510: This block can be replaced by self.clearExistingImages()

src/asm3/publishers/adoptapet.py
adamdspencer77: Removed.

src/asm3/publishers/adoptapet.py
adamdspencer77: Done.

src/asm3/publishers/adoptapet.py
adamdspencer77: Done.

PR #1901, opened 2025-10-14
Create waiting list from animal

adamdspencer77:
Similar to how an animal record can be created from a waiting list entry with a toolbar button, there should be an option to do the reverse from an animal record.

This scenario comes up when an adopter wants to return their animal but the shelter does not have room. Have a button on the animal toolbar which is only active when an animal has an open adoption, transfer or reclaim movement.

It creates a waiting list entry, uses the person from the last exit movement as the contact and puts an animal summary in the description (including its code, name, species, gender, breed and age).

### DIFF
--- a/src/asm3/animal.py
+++ b/src/asm3/animal.py
@@ -7110,23 +7110,16 @@ def create_waitinglist(dbo: Database, username: str, aid: int) -> int:
     """
     l = dbo.locale
     a = get_animal(dbo, aid)
-    description = "#s:" + a.SHELTERCODE
-    microchip = a.IDENTICHIPNUMBER
-    if a.IDENTICHIP2NUMBER:
-        microchip += " " + a.IDENTICHIP2NUMBER
-    neutered = "off"
-    if a.NEUTERED == 1:
-        neutered = "on"
     data = {
         "dateputon":        python2display(l, dbo.today()),
         "animalname":       a.ANIMALNAME,
         "species":          a.SPECIESID,
         "owner":            a.CURRENTOWNERID,
-        "description":      description,
-        "microchip":        microchip,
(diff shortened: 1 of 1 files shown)


src/static/js/animal.js
```diff
@@ -1329,6 +1332,38 @@ $(function() {
...
+                        "animalname=" + controller.animal.ANIMALNAME,
+                        "species=" + controller.animal.SPECIESID,
+                        "owner=" + controller.animal.CURRENTOWNERID,
+                        "description=" + description,
+                        "microchip=" + microchip,
+                        "breed=" + controller.animal.BREEDID,
+                        "sex=" + controller.animal.SEX,
+                        "neutered=" + neutered,
+                        "size=" + controller.animal.SIZE,
+                        "dateofbirth=" + format.date(controller.animal.DATEOFBIRTH),
```
bobintetley: description and animalname here will break the moment they contain an &, = or other reserved bit of punctuation as they would need to be encoded as a URI component (or make formdata an object)

I think rather than calling the waitinglist endpoint here, it would be neater to:

1.  add a new "create_waitinglist" function to animal.py where you can pass in the animal ID and it does the work of copying bits to make the waiting list record
2.  here, you'd just post the animal ID to a new post_waitinglist endpoint in main.py/class animal that calls animal.create_waitinglist and returns the new waiting list ID for the redirect.

waitinglist.py has a similar function called create_animal so this would be consistent. Note that your new create_waitinglist function wouldn't need to do all the media copying and extra stuff that the create_animal one does. It means that we aren't limited to what waitinglist/post_create can do.

adamdspencer77: Hopefully sorted. Noticed that I'm getting some tinymce errors popping up in the console, not sure if I need to pay attention to them.

Issue #1614, opened 2025-01-14
Person view roles


bobintetley:
Similar to how we do for incidents, allow an option to turn on extra view permissions for people.

This would allow for a role dropdown on person records, where users can assign the only roles that are allowed to view that specific record.

A lot of people have asked for this over the years.

The option should go under Remove->System->Remove fine-grained person permissions (set by default)

Like animalcontrol.py, person.py will need its own reduce_find_results functions that are called by the find routines to throw away rows the user does not have permission to view.

PR #1912, opened 2025-10-29
Animal view roles

adamdspencer77:
Similar to how we do for incidents (and people once https://github.com/sheltermanager/asm3/issues/1614 is done), allow an option to turn on extra view permissions for animals.

This would allow for a role dropdown on animals records, where users can assign the only roles that are allowed to view that specific record.

The option should go under Remove->System->Remove fine-grained animal permissions (set by default)

Like animalcontrol.py, animal.py will need its own reduce_find_results function that is called by animal find routines to throw away rows the user does not have permission to view. shelterview will need to call the animal one as well.

animal.py has a LocationFilter object that would be the logical place to implement the reduce function so that anything locationfilter enabled gets the functionality.

### DIFF
--- a/src/asm3/configuration.py
+++ b/src/asm3/configuration.py
@@ -54,6 +54,7 @@ DEFAULTS = {
     "AnimalNameChangeLog": "No",
     "AnimalNameChangeLogType": "3",
     "AnimalFiguresSplitEntryReason": "No",
+    "AnimalPermissions": "No",
     "AnimalSearchResultsNewTab": "No",
     "PersonSearchResultsNewTab": "No",
     "AnnualFiguresShowBabies": "Yes",
@@ -360,6 +361,7 @@ DEFAULTS = {
         "MembershipNumber,AdditionalFlags,OwnerAddress," \
         "OwnerTown,OwnerCounty,OwnerPostcode,HomeTelephone,WorkTelephone," \
         "MobileTelephone,EmailAddress",
+    "PersonPermissions": "No",
     "PetFinderSendStrays": "No",
     "PetFinderSendHolds": "No",
     "PetFinderSendAdopted": "No",


src/main.py
```diff
@@ -311,6 +311,25 @@ def check_animal(self, a: ResultRow) -> None:
...
+        # Does the user have any of the view roles?
+        userroles = []
+        for ur in session.dbo.query("SELECT RoleID FROM userrole INNER JOIN users ON userrole.UserID = users.ID WHERE users.UserName LIKE ?", [session.user]):
+            userroles.append(ur.ROLEID)
+        hasperm = False
+        if session.superuser:
+            hasperm = True
+        for ur in userroles:
+            if ur in viewroles:
+                hasperm = True
```
bobintetley: I'm not keen on all this functionality being here in main.py. 

I think this should move to users.py and be put in a reusable function. Make a function called "check_role_bool", similar to the check_permission_bool function that already exists in users.py

It should accept the Session object and a list of role IDs (your viewroles list above) 

It returns True if the user is a superuser, or if they have one of the role IDs in the list given.  It then reduces this endpoint code to a simple readable method call then.

src/main.py
```diff
@@ -311,6 +311,25 @@ def check_animal(self, a: ResultRow) -> None:
         lf = asm3.animal.LocationFilter(session.locationfilter, session.siteid, session.visibleanimalids)
         if not lf.match(a):
             raise asm3.utils.ASMPermissionError("animal not in location filter/site")
+        
+        viewroles = []
+        for rr in session.dbo.query("SELECT RoleID FROM animalrole WHERE AnimalID = ? AND CanView = 1", [a.ID]):
+            viewroles.append(rr.ROLEID)
```
bobintetley: There's a helper method in dbo that does this. You can call ``` dbo.query_list(sql) ``` instead and you will get back a list that contains the first column of the results. Saves you having to iterate the results of query() and put it in a list yourself.

``` viewroles = session.dbo.query_list("SELECT RoleID FROM animalrole WHERE AnimalID = ? AND CanView = 1", [a.ID]) ```

File: src/main.py (hunk 1 above)

File: src/main.py (hunk 2 above)

File: src/asm3/users.py [hunk 3]
```diff
@@ -264,6 +264,24 @@ def check_permission_map_bool(superuser: int, securitymap: str, flag: str) -> bo
     if has_security_flag(securitymap, flag): return True
     return False
 
+def check_role_bool(session: Session, roles: list) -> bool:
```
bobintetley: Rather than list, use the List typehint, then you can state what you expect to be in the list. In this case ``` List[int] ```

src/asm3/users.py
```diff
@@ -264,6 +264,24 @@ def check_permission_map_bool(superuser: int, securitymap: str, flag: str) -> bo
...
+    """
+    Returns True if a user has a role assigned from the provided roles list
+    """
+    if "superuser" not in session or "securitymap" not in session: return False
+    if session.superuser == 1: return True
+
+    userroles = []
+    for ur in session.dbo.query("SELECT RoleID FROM userrole INNER JOIN users ON userrole.UserID = users.ID WHERE users.UserName LIKE ?", [session.user]):
+        userroles.append(ur.ROLEID)
```
bobintetley: use query_list here - then you don't need the loop.  ``` userroles = session.dbo.query_list("SELECT RoleID ... ```

src/asm3/users.py
```diff
@@ -264,6 +264,24 @@ def check_permission_map_bool(superuser: int, securitymap: str, flag: str) -> bo
...
+    hasperm = False
+    if session.superuser:
+        hasperm = True
```
bobintetley: This if block isn't necessary - we already returned True above if superuser was set.

src/asm3/users.py
adamdspencer77: Done.

src/asm3/users.py
adamdspencer77: Done.

src/asm3/users.py
adamdspencer77: Removed.

PR #1913, opened 2025-11-03
Option to convert document to PDF on Media tab

adamdspencer77:
It would be great to be able to convert document created via Document Templates to PDF inside the media tab. For context, we usually send adopters their paperwork after their trial and it can be a mix of images of Vaccination Certificates, PDF document from vets etc, so if we have to email a document as PDF, we need to convert to PDF and then upload it to media. We can't use the Email PDF function as the Congratulations email contains a mix of media file types.

Not sure I understand how being able to convert documents to PDF would help with the scenario explained. Wouldn't allowing all media types to be included in the "Email PDF" function (images, documents and existing PDFs) be a more practical solution?

Yes, you're right, it probably would work better because it means they wouldn't have to store the PDF versions of the documents. Email PDF already accepts multiple documents and converts them, so I don't think it would be difficult to include but not convert non-html attachments too.

### DIFF
--- a/src/static/js/media.js
+++ b/src/static/js/media.js
@@ -89,8 +89,6 @@ $(function() {
                     if (rows.length > 0 && no_links()) {
                         $("#button-email").button("option", "disabled", false);
                     }
-                    // Only allow the email pdf button to be pressed if the
-                    // selection only contains documents
                     if (rows.length > 0) {
                         $("#button-emailpdf").button("option", "disabled", false); 
                     }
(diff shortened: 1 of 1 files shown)


src/static/js/media.js
```diff
@@ -91,7 +91,7 @@ $(function() {
                     }
                     // Only allow the email pdf button to be pressed if the
                     // selection only contains documents
```
bobintetley: this comment wants removing since it's no longer true

src/static/js/media.js
```diff
@@ -961,8 +961,8 @@ $(function() {
                     name: defaultname,
                     email: defaultemail,
                     toaddresses: toaddresses,
-                    subject: common.replace_all(media.selected_filenames(), ".html", ".pdf"),
-                    attachments: common.replace_all(media.selected_filenames(), ".html", ".pdf"),
+                    subject: media.selected_filenames(),
+                    attachments: media.selected_filenames(),
```
bobintetley: I think we still need to do the replace here - this is for the attachment filenames shown in the send email dialog, and we need to indicate to the user that any html files will be pdfs by the time the recipient gets them

src/static/js/media.js
adamdspencer77: Done.

src/static/js/media.js
adamdspencer77: Done.

PR #1915, opened 2025-11-04
Diary note sending interval

adamdspencer77:
While you can choose to send diary notes immediately, the only other option to follow them up is a daily reminder. For some, this is too much, have an option to remind weekly instead of daily.

### DIFF
--- a/src/static/js/options.js
+++ b/src/static/js/options.js
@@ -601,10 +601,7 @@ $(function() {
                         { id: "alldiaryhomepage", post_field: "AllDiaryHomePage", label: _("Show the full diary (instead of just my notes) on the home page"), type: "check", fullrow: true }, 
                         { id: "diarycompleteondeath", post_field: "DiaryCompleteOnDeath", label: _("Auto complete diary notes linked to animals when they are marked deceased"), type: "check", fullrow: true }, 
                         { id: "emaildiarynotes", post_field: "EmailDiaryNotes", label: _("Email users their outstanding diary notes once per day"), type: "check", fullrow: true }, 
-                        { id: "emaildiaryonchange", post_field: "EmailDiaryOnChange", label: _("Email users immediately when a diary note assigned to them is created or updated"), type: "check", fullrow: true }, 
-                        { id: "emaildiaryoncomplete", post_field: "EmailDiaryOnComplete", label: _("Email diary note creators when a diary note is marked complete"), type: "check", fullrow: true }, 
-                        { id: "emailmessages", post_field: "EmailMessages", label: _("When a message is created, email it to each matching user"), type: "check", fullrow: true }, 
-                        { id: "emaildiarynotesweekly", post_field: "EmailDiaryNotesWeekly", label: _("Email users their outstandiong diary notes weekly on "), type: "check", fullrow: true,
+                        { id: "emaildiarynotesweekly", post_field: "EmailDiaryNotesWeekly", label: _("Email users their outstanding diary notes weekly on "), type: "check", fullrow: true,
                             xmarkup: '<select id="emaildiarynotesweeklyday" data-post="EmailDiaryNotesWeeklyDay" class="asm-field asm-selectbox asm-widget">' + 
                             html.list_to_options(
                             [
@@ -616,7 +613,10 @@ $(function() {
                                 "5|" + _("Saturday"),
                                 "6|" + _("Sunday")
                             ]) + '</select>'
-                        }
+                        },
+                        { id: "emaildiaryonchange", post_field: "EmailDiaryOnChange", label: _("Email users immediately when a diary note assigned to them is created or updated"), type: "check", fullrow: true }, 
+                        { id: "emaildiaryoncomplete", post_field: "EmailDiaryOnComplete", label: _("Email diary note creators when a diary note is marked complete"), type: "check", fullrow: true }, 
+                        { id: "emailmessages", post_field: "EmailMessages", label: _("When a message is created, email it to each matching user"), type: "check", fullrow: true }
                     ]}, 
                     { id: "tab-display", title: _("Display"), fields: [
                         { id: "disableeffects", post_field: "rc:DisableEffects", label: _("Enable visual effects"), type: "check", fullrow: true }, 
--- a/doc/manual/options.rst
+++ b/doc/manual/options.rst
@@ -473,6 +473,10 @@ Diary and Messages
   For this option to work, you must have configured the system email in
   the Email tab of this screen and your users must have email addresses set.
 
+* Email users their outstanding diary notes once per week on: This option will cause
+  the system to send users an email containing their outstanding diary notes
+  as part of the overnight batch on the day of the week you specify.
+
 * Email users immediately when a diary note assigned to them is created or
   updated: This option will cause an email be sent to any users a diary note is
   assigned to as soon as you create or make a change to it. 


src/asm3/diary.py
```diff
@@ -51,12 +52,18 @@ def email_uncompleted_upto_today(dbo: Database) -> None:
...
-                    if n.linkinfo is not None and n.linkinfo != "": s += " / %s" % n.linkinfo
-                    s += " (%s)" % n.createdby
-                    s += "\n%s\n\n%s" % (n.note, n.comments)
-                    totalforuser += 1
+                    weekcheck = True
+                    if reminderperiod == 7:
+                        datedelta = (asm3.i18n.now() - n.diarydatetime).days
+                        if not datedelta or datedelta % 7:
+                            weekcheck = False
+                    if reminderperiod == 1 or ( reminderperiod == 7 and weekcheck ):
```
bobintetley: I think this was originally envisioned as a digest type option - so it basically would send all the outstanding diary notes for the user once per week on the same day. This looks to be sending on the day that's one week from the day on the diary note itself (unless I'm misunderstanding?).

src/static/js/options.js
```diff
@@ -598,12 +598,18 @@ $(function() {
                         { id: "medianotesfile", post_field: "DefaultMediaNotesFromFile", label: _("Prefill new media notes with the filename if left blank"), type: "check", fullrow: true }
                     ]}, 
                     { id: "tab-diaryandmessages", title: _("Diary and Messages"), fields: [
-                        { id: "alldiaryhomepage", post_field: "AllDiaryHomePage", label: _("Show the full diary (instead of just my notes) on the home page"), type: "check" }, 
-                        { id: "diarycompleteondeath", post_field: "DiaryCompleteOnDeath", label: _("Auto complete diary notes linked to animals when they are marked deceased"), type: "check" }, 
-                        { id: "emaildiarynotes", post_field: "EmailDiaryNotes", label: _("Email users their outstanding diary notes once per day"), type: "check" },
```
bobintetley, line 603: I would keep this option as it was, since it was the default and lots of people are used to it as it is (although I see you added a database update to fix that).

src/static/js/options.js
```diff
@@ -598,12 +598,18 @@ $(function() {
...
+                        { id: "alldiaryhomepage", post_field: "AllDiaryHomePage", label: _("Show the full diary (instead of just my notes) on the home page"), type: "check", fullrow: true }, 
+                        { id: "diarycompleteondeath", post_field: "DiaryCompleteOnDeath", label: _("Auto complete diary notes linked to animals when they are marked deceased"), type: "check", fullrow: true }, 
+                        { id: "emaildiaryonchange", post_field: "EmailDiaryOnChange", label: _("Email users immediately when a diary note assigned to them is created or updated"), type: "check", fullrow: true }, 
+                        { id: "emaildiaryoncomplete", post_field: "EmailDiaryOnComplete", label: _("Email diary note creators when a diary note is marked complete"), type: "check", fullrow: true }, 
+                        { id: "emailmessages", post_field: "EmailMessages", label: _("When a message is created, email it to each matching user"), type: "check", fullrow: true }, 
+                        { id: "emaildiarynotes", post_field: "EmailDiaryNotes", label: _("Email users their outstanding diary notes"), type: "select",
```
bobintetley: I would make this two new options, EmailDiaryNotesWeekly and EmailDiaryNotesWeeklyDay. The first is a tick box like the daily one and the second is a day of the week from 0-6. Can show it then as "Email users their outstanding diary notes once per week on ...."

This way, they aren't mutually exclusive, the daily one continues to work as it always has, and people can configure the weekly one and turn off the daily one if they want. 

It also makes the comparison code in email_uncompleted simpler because you're just going to check whether the day today is the one they configured

src/asm3/dbupdates/35106.py
```diff
@@ -0,0 +1,14 @@
...
+
+if queryresult:
+    itemvalue = queryresult[0].ITEMVALUE
+
+    if itemvalue == "Yes":
+        execute(dbo, "UPDATE configuration SET ItemValue = ? WHERE ItemName = ?", ["EmailDiaryNotes", "1"])
+    else:
+        execute(dbo, "UPDATE configuration SET ItemValue = ? WHERE ItemName = ?", ["EmailDiaryNotes", "-1"])
+else:
+    execute(dbo, "INSERT INTO configuration (ItemName, ItemValue) VALUES (?, ?)", ["EmailDiaryNotes", "1"])
```
bobintetley: this whole database update is not necessary, see notes below! :)

src/asm3/configuration.py
```diff
@@ -1117,8 +1117,8 @@ def email_clinic_reminder_days(dbo: Database) -> int:
 def email_clinic_reminder_template(dbo: Database) -> int:
     return cint(dbo, "EmailClinicReminderTemplate", DEFAULTS["EmailClinicReminderTemplate"])
 
-def email_diary_notes(dbo: Database) -> bool:
-    return cboolean(dbo, "EmailDiaryNotes", DEFAULTS["EmailDiaryNotes"] == "Yes")
+def email_diary_notes(dbo: Database) -> int:
+    return cint(dbo, "EmailDiaryNotes", DEFAULTS["EmailDiaryNotes"] == "1")
```
bobintetley: keep the existing type, add new ones as below ...

src/static/js/options.js
adamdspencer77: Done.

src/asm3/dbupdates/35106.py
adamdspencer77: Removed.

src/static/js/options.js
adamdspencer77: Done.

src/asm3/diary.py
adamdspencer77: Changed.

src/asm3/configuration.py
adamdspencer77: Hopefully sorted.

src/asm3/diary.py
```diff
@@ -64,6 +65,13 @@ def email_uncompleted_upto_today(dbo: Database) -> None:
                 if asm3.configuration.audit_on_send_email(dbo): 
                     asm3.audit.email(dbo, "system", asm3.configuration.email(dbo), u.emailaddress, "", "", subject, s)
 
+# def email_uncompleted_weekly(dbo: Database) -> None:
+#     """
+#     Goes through all system users and emails them their non completed diary notes if the day of the week
+#     is the one specified in the options - unless the option is turned off.
+#     """
+
```
bobintetley: Probably meant to remove this?

src/asm3/diary.py
```diff
@@ -31,7 +31,8 @@ def email_uncompleted_upto_today(dbo: Database) -> None:
     Goes through all system users and emails them their diary for the
     day - unless the option is turned off.
     """
-    if not asm3.configuration.email_diary_notes(dbo): return
+    if not asm3.configuration.email_diary_notes(dbo) and asm3.configuration.email_diary_notes_weekly(dbo) != asm3.i18n.today().isoweekday() - 1:
+        return
```
bobintetley, line 36: This condition is quite hard to understand at first glance. So, we're bailing out of the function if daily is NOT set, the weekly day doesn't match. 

I think there should be a comment here at least to explain the logic.

src/asm3/diary.py
adamdspencer77: Yep. Removed.

src/asm3/diary.py
adamdspencer77: Comment added.

PR #1916, opened 2025-11-05
Accounting transactions: person link

adamdspencer77:
Add an editable OwnerID column to transactions to link a transaction to a person. It could filter for people with the supplier flag by default.

We could default OwnerID automatically when creating transactions with financial.update_matching_donation_transaction and financial.update_matching_cost_transaction - it should not be edited on transactions with these links


src/static/js/accounts_trx.js
```diff
@@ -10,6 +10,7 @@ $(function() {
             return [
                 '<div id="dialog-edit" style="display: none" title="' + html.title(_("Edit transaction")) + '">',
                 tableform.fields_render([
+                    { post_field: "supplier", label: _("Supplier"), type: "person", personfilter: "supplier", personmode: "brief" },
```
bobintetley, line 13: Make sure this new chooser is destroyed in the destructor for the screen (the accounts_trx.destroy method)

```common.widget_destroy("#supplier", "personchooser")```

File: src/static/js/accounts_trx.js [hunk 2]
```diff
@@ -279,6 +280,15 @@ $(function() {
...
                 $("#trxid").val(row.ID);
+                $("#supplier").personchooser("clear");
+                if (row.ANIMALCOSTID || row.OWNERDONATIONID) {
+                    $("#supplierrow").hide();
+                } else {
+                    $("#supplierrow").show();
+                }
+                if (row.OWNERID) {
+                    $("#supplier").personchooser("loadbyid", row.OWNERID);
+                }
```
bobintetley, line 294: The link to this person needs displaying in the render_tablebody function so that the user can see it in the list of transactions

src/static/js/accounts_trx.js
adamdspencer77: Done.

src/static/js/accounts_trx.js
adamdspencer77: Hopefully sorted.

PR #1919, opened 2025-11-06
Hide clinic and boarding calendar views when remove options are enabled

adamdspencer77:
As title suggests, the calendar options for boarding and clinic should be hidden when the remove options are enabled but currently remain visible.

### DIFF
--- a/src/static/js/calendarview.js
+++ b/src/static/js/calendarview.js
@@ -27,9 +27,7 @@ $(function() {
 
         render: function() {
             const chk = function(id, data, icon, label, tag="") {
-                if (tag) {
-                    tag = " " + tag;
-                }
+                if (tag) {tag = " " + tag;}
                 return '<span class="asm-calendar-legend' + tag + '">' + 
                     html.icon(icon) + 
                     '<input id="' + id + '" data="' + data + '" type="checkbox" class="asm-checkbox" /> ' +


src/static/js/calendarview.js
```diff
@@ -26,8 +26,11 @@ $(function() {
         },
 
         render: function() {
-            const chk = function(id, data, icon, label) {
-                return '<span class="asm-calendar-legend">' + 
+            const chk = function(id, data, icon, label, tag="") {
+                if (tag) {
```
bobintetley: little style thing, but if a block is only one line in height, I prefer to put the whole lot on one line instead. 

``` if (tag) { tag= " " + tag; } ```

File: src/static/js/calendarview.js [hunk 2]
```diff
@@ -40,13 +43,13 @@ $(function() {
...
+                chk("toggle-clinic", "c", "health", _("Clinic"), "tagclinic"),
                 chk("toggle-donation", "p", "donation", _("Payment")),
-                chk("toggle-incident", "o", "call", _("Incident followup")),
-                chk("toggle-transport", "r", "transport", _("Transport")),
-                chk("toggle-traploan", "l", "traploan", _("Equipment loan")),
-                chk("toggle-event", "e", "event", _("Event")),
+                chk("toggle-incident", "o", "call", _("Incident followup"), "taganimalcontrol"),
+                chk("toggle-transport", "r", "transport", _("Transport"), "tagtransport"),
+                chk("toggle-traploan", "l", "traploan", _("Equipment loan"), "tagtraploan"),
+                chk("toggle-event", "e", "event", _("Event"), "tagevent"),
```
bobintetley, line 50: I might be missing something in this patch, but the code that checks the config options and hides these elements based on the new tag classes seems to be missing?

src/static/js/calendarview.js
adamdspencer77: Done.

src/static/js/calendarview.js
adamdspencer77: image
This code was already there. I just added the tags so the existing code had something to grab.

bobintetley: ah, didn't realise that was already in there, cool :)

PR #1943, opened 2025-11-11
Disease tracking

adamdspencer77:
One user asked for disease tracking, this is something competitors have. They'd like to be able to track the duration of an animal having a particular disease. Probably warrants more investigating in terms of what data needs to be stored and reported on.

**How about a "conditions" table recording the startdatetime and enddatetime of a condition. This could accommodate diseases, individual symptoms such as coughing and diarrhoea and other conditions such as pregnancy, seasons, nursing or cage rest. If a conditionstype table was also added then specific conditions such as Parvo and Cat Flu could be grouped into categories such as disease, injury, medically induced, natural.**

### DIFF
--- a/src/asm3/animal.py
+++ b/src/asm3/animal.py
@@ -2622,13 +2622,6 @@ def get_animalconditions(dbo: Database, animalid: int, sort: int = ASCENDING) ->
         sql += " ORDER BY ac.StartDatetime DESC"
     return dbo.query(sql, [animalid] )
 
-def get_conditions(dbo: Database) -> Results:
-    """
-    Returns all active lkcondition records:
-    """
-    sql = "SELECT * FROM lkcondition ORDER BY ConditionName"
-    return dbo.query(sql)
-
 def get_diets(dbo: Database, animalid: int, sort: int = ASCENDING) -> Results:
     """
     Returns diet records for the given animal:
--- a/src/asm3/dbupdates/35108.py
+++ b/src/asm3/dbupdates/35108.py
@@ -32,11 +32,8 @@ fields = ",".join([
 ])
 execute(dbo, dbo.ddl_add_table("lksconditiontype", fields) )
 
-for conditiontype in (
-    _("GI"),
-    _("Respiratory"),
-    _("Miscellaneous"),
-    _("Reproductive"),
-    _("Symptom")
-):
-    dbo.execute("INSERT INTO lksconditiontype (ConditionTypeName, Description, IsRetired) VALUES (?, ?, ?)", [conditiontype, "", 0] )
+dbo.execute("INSERT INTO lksconditiontype (ID, ConditionTypeName, Description, IsRetired) VALUES (?, ?, ?, ?)", [1, _("GI"), "", 0] )
+dbo.execute("INSERT INTO lksconditiontype (ID, ConditionTypeName, Description, IsRetired) VALUES (?, ?, ?, ?)", [2, _("Respiratory"), "", 0] )
+dbo.execute("INSERT INTO lksconditiontype (ID, ConditionTypeName, Description, IsRetired) VALUES (?, ?, ?, ?)", [3, _("Miscellaneous"), "", 0] )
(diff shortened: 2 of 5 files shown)


src/asm3/animal.py
```diff
@@ -2607,6 +2607,28 @@ def get_costs_for_payee(dbo: Database, personid: int, sort: int = ASCENDING) ->
...
+        "INNER JOIN lksconditiontype ct ON c.ConditionTypeID = ct.ID " \
+        "WHERE ac.AnimalID = ?"
+    if sort == ASCENDING:
+        sql += " ORDER BY ac.StartDatetime"
+    else:
+        sql += " ORDER BY ac.StartDatetime DESC"
+    return dbo.query(sql, [animalid] )
+
+def get_conditions(dbo: Database) -> Results:
+    """
```
bobintetley: This should be in lookups.py

src/asm3/dbupdates/35108.py
```diff
@@ -0,0 +1,42 @@
...
+execute(dbo, dbo.ddl_add_table("lksconditiontype", fields) )
+
+for conditiontype in (
+    _("GI"),
+    _("Respiratory"),
+    _("Miscellaneous"),
+    _("Reproductive"),
+    _("Symptom")
+):
+    dbo.execute("INSERT INTO lksconditiontype (ConditionTypeName, Description, IsRetired) VALUES (?, ?, ?)", [conditiontype, "", 0] )
```
bobintetley: Are you relying on some kind of autoincrement to set the ID here? I'm not sure one exists that is consistent across all 3 databases we support. Probably simpler to have 5 insert lines and explicitly set the ID like other updates.

src/asm3/lookups.py
```diff
@@ -39,6 +39,7 @@
     "lkcoattype":           (_("Coat Types"), "CoatType", _("Coat Type"), "", "add del", ("animal.CoatType",)),
     "citationtype":         (_("Citation Types"), "CitationName", _("Citation Type"), "CitationDescription", "add del ret cost", ("ownercitation.CitationTypeID",)),
     "lksclinicstatus":      (_("Clinic Statuses"), "Status", _("Status"), "", "", ("clinicappointment.Status",)),
+    "lkcondition":          (_("Conditions"), "ConditionName", _("Name"), "Description", "add del ret conditiontype haszoonotic", ("animalcondition.ConditionID",)),
```
bobintetley, line 42: still need an entry for lksconditiontype.

lks tables basically aren't editable if your language is English, but people in non-English locales can still edit them in order to translate the values.

src/asm3/animal.py
```diff
@@ -2607,6 +2607,28 @@ def get_costs_for_payee(dbo: Database, personid: int, sort: int = ASCENDING) ->
...
+    Returns all active lkcondition records:
+    """
+    sql = "SELECT * FROM lkcondition ORDER BY ConditionName"
+    return dbo.query(sql)
```
bobintetley: Seems kind of unnecessary to create a variable when the string literal can be the parameter

src/static/js/lookups.js
```diff
@@ -209,6 +216,7 @@ $(function() {
                             "mode=create&lookup=" + controller.tablename + "&namefield=" + controller.namefield, "lookups");
                         let row = {};
                         row.ID = response;
+                        console.log(dialog.fields);
```
bobintetley: debug code :)

src/asm3/animal.py
adamdspencer77: Moved.

src/asm3/dbupdates/35108.py
adamdspencer77: Done.

src/asm3/animal.py

src/asm3/lookups.py
adamdspencer77: Hopefully sorted.

src/static/js/lookups.js
adamdspencer77: Removed.

PR #1946, opened 2025-11-13
Diary note end time

adamdspencer77:
Have a field to allow an end time to be set on diary notes. They can be shown correctly on the diary calendar then. Also needs to be shown on diary editing screens and the home page.

### DIFF
--- a/src/asm3/diary.py
+++ b/src/asm3/diary.py
@@ -352,9 +352,9 @@ def insert_diary_from_form(dbo: Database, username: str, linktypeid: int, linkid
     post: A PostedData object
     """
     l = dbo.locale
-    if post["diarydatetimedate"] == "":
+    if post["diarydate"] == "":
         raise asm3.utils.ASMValidationError(asm3.i18n._("Diary date cannot be blank", l))
-    if post.date("diarydatetimedate") is None:
+    if post.date("diarydate") is None:
         raise asm3.utils.ASMValidationError(asm3.i18n._("Diary date is not valid", l))
     if post["subject"] == "":
         raise asm3.utils.ASMValidationError(asm3.i18n._("Diary subject cannot be blank", l))
@@ -371,8 +371,8 @@ def insert_diary_from_form(dbo: Database, username: str, linktypeid: int, linkid
         "LinkID":           linkid,
         "LinkType":         linktypeid,
         "LinkInfo":         linkinfo,
-        "DiaryDateTime":    post.datetime("diarydatetimedate", "diarydatetimetime"),
-        "DiaryEndDateTime": post.datetime("diaryenddatetimedate", "diaryenddatetimetime"),
+        "DiaryDateTime":    post.datetime("diarydate", "diarytime"),
+        "DiaryEndDateTime": post.datetime("diaryenddate", "diaryendtime"),
         "DiaryForName":     post["diaryfor"],
         "ColourSchemeID":   post.integer("diarycolourscheme"),
         "Subject":          post["subject"],
@@ -385,7 +385,7 @@ def insert_diary_from_form(dbo: Database, username: str, linktypeid: int, linkid
         email_note_on_change(dbo, get_diary(dbo, diaryid), username)
     return diaryid
 
-def insert_diary(dbo: Database, username: str, linktypeid: int, linkid: int, diarydate: datetime, diaryfor: str, subject: str, note: str, emailnow: bool = False, colourschemeid: int = 0, diaryenddatetime: datetime = None) -> int:
+def insert_diary(dbo: Database, username: str, linktypeid: int, linkid: int, diarydate: datetime, diaryfor: str, subject: str, note: str, emailnow: bool = False, colourschemeid: int = 0, diaryenddate: datetime = None) -> int:
     """
     Creates a diary note from the form data
     username: User creating the diary
@@ -403,7 +403,7 @@ def insert_diary(dbo: Database, username: str, linktypeid: int, linkid: int, dia
         "LinkType":         linktypeid,
         "LinkInfo":         linkinfo,
         "DiaryDateTime":    diarydate,
-        "DiaryEndDateTime": diaryenddatetime,
+        "DiaryEndDateTime": diaryenddate,
         "DiaryForName":     diaryfor,
         "ColourSchemeID":   colourschemeid,
         "Subject":          subject,
@@ -419,9 +419,9 @@ def update_diary_from_form(dbo: Database, username: str, post: PostedData) -> No
     Updates a diary note from form data
     """
     l = dbo.locale
-    if post["diarydatetimedate"] == "":
+    if post["diarydate"] == "":
         raise asm3.utils.ASMValidationError(asm3.i18n._("Diary date cannot be blank", l))
-    if post.date("diarydatetimedate") is None:
+    if post.date("diarydate") is None:
         raise asm3.utils.ASMValidationError(asm3.i18n._("Diary date is not valid", l))
     if post["subject"] == "":
         raise asm3.utils.ASMValidationError(asm3.i18n._("Diary subject cannot be blank", l))
@@ -434,8 +434,8 @@ def update_diary_from_form(dbo: Database, username: str, post: PostedData) -> No
 
     diaryid = post.integer("diaryid")
     dbo.update("diary", diaryid, {
-        "DiaryDateTime":    post.datetime("diarydatetimedate", "diarydatetimetime"),
-        "DiaryEndDateTime": post.datetime("diaryenddatetimedate", "diaryenddatetimetime"),
+        "DiaryDateTime":    post.datetime("diarydate", "diarytime"),
+        "DiaryEndDateTime": post.datetime("diaryenddate", "diaryendtime"),
         "DiaryForName":     post["diaryfor"],
         "ColourSchemeID":   post.integer("diarycolourscheme"),
         "Subject":          post["subject"],
--- a/src/static/js/diary.js
+++ b/src/static/js/diary.js
@@ -21,8 +21,8 @@ $(function() {
                     { json_field: "COLOURSCHEMEID", post_field: "diarycolourscheme", label: _("Color Scheme"), 
                         type: "selectcolour", defaultval: 1, 
                         callout: _("The color scheme to be used when displaying this note on the calendar and home screen") },
-                    { json_field: "DIARYDATETIME", post_field: "diarydatetime", label: _("Date"), type: "datetime", validation: "notblank", defaultval: new Date() },
-                    { json_field: "DIARYENDDATETIME", post_field: "diaryenddatetime", label: _("Until"), type: "datetime", hideif: function() {
+                    { json_field: "DIARYDATETIME", post_field: "diary", label: _("Date"), type: "datetime", validation: "notblank", defaultval: new Date() },
+                    { json_field: "DIARYENDDATETIME", post_field: "diaryend", label: _("Until"), type: "datetime", hideif: function() {
                             return config.bool("DisableDiaryEndDatetime");
                         }
                     },
@@ -44,8 +44,8 @@ $(function() {
                             try {
                                 tableform.fields_update_row(dialog.fields, row);
                                 row.COLOURSCHEMEID = $("#diarycolourscheme").selectcolour("value");
-                                let enddate = $("#diaryenddatetimedate").val();
-                                let endtime = $("#diaryenddatetimetime").val();
+                                let enddate = $("#diaryenddate").val();
+                                let endtime = $("#diaryendtime").val();
                                 if (endtime) { endtime += ":00"; }
                                 if (enddate) { row.DIARYENDDATETIME = format.date_iso(enddate + " " + endtime); }
                                 await tableform.fields_post(dialog.fields, "mode=update&diaryid=" + row.ID, "diary");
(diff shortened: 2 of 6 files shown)


src/asm3/diary.py
```diff
@@ -433,6 +435,7 @@ def update_diary_from_form(dbo: Database, username: str, post: PostedData) -> No
     diaryid = post.integer("diaryid")
     dbo.update("diary", diaryid, {
         "DiaryDateTime":    post.datetime("diarydate", "diarytime"),
+        "DiaryEndDateTime": post.datetime("diarydate", "diaryendtime"),
```
bobintetley: I can't see it in this patch, but there should be a similar line in insert_diary_from_form as well i think?

src/static/js/diary.js
```diff
@@ -23,6 +23,10 @@ $(function() {
                         callout: _("The color scheme to be used when displaying this note on the calendar and home screen") },
                     { json_field: "DIARYDATETIME", post_field: "diarydate", label: _("Date"), type: "date", validation: "notblank", defaultval: new Date() },
                     { json_field: "DIARYDATETIME", post_field: "diarytime", label: _("Time"), type: "time" },
+                    { json_field: "DIARYENDDATETIME", post_field: "diaryendtime", label: _("Until"), type: "time", 
+                        hideif: function() {
+                            return config.bool("DisableDiaryEndDatetime");
+                        } },
```
bobintetley: Very minor thing, I prefer little one line functions like this on one line to save space

```   hideif: function() { return config.bool("DisableDiaryEndDatetime"); } }, ```

File: src/static/js/diary.js [hunk 3]
```diff
@@ -81,7 +85,18 @@ $(function() {
...
                     { field: "DIARYFORNAME", display: _("For") },
-                    { field: "DIARYDATETIME", display: _("Date"), formatter: tableform.format_datetime, initialsort: true, initialsortdirection: "desc" },
+                    { field: "DIARYDATETIME", display: _("Date"), initialsort: true, initialsortdirection: "desc",
+                        formatter: function(row) {
+                            let starttime = tableform.format_time_blank(row, row.DIARYDATETIME);
+                            let endtime = tableform.format_time_blank(row, row.DIARYENDDATETIME);
+                            let dateoutput = format.date(row.DIARYDATETIME);
+                            if (starttime) { dateoutput += " " + starttime; }
+                            if ( !config.bool("DisableDiaryEndDatetime") ) {
+                                if (endtime) { dateoutput += _(" to ") + endtime; }
```
bobintetley: not keen on using _(" to ") as a translation string because it assumes the word order of target languages.

Use _("{0} to {1}") and replace {0} and {1} with the dates so that translators are better able to control the translation. I wouldn't be surprised if we don't already have "{0} to {1}" as an existing string too.

src/static/js/main.js
```diff
@@ -475,6 +475,12 @@ $(function() {
                 if (d.LINKTYPE == 7) { link = "incident?id=" + d.LINKID; }
                 s.push('<tr title="' + html.title(common.substitute(_("Added by {0} on {1}"), { "0": d.CREATEDBY, "1": format.date(d.CREATEDDATE) })) + '">');
                 s.push('<td>' + format.date(d.DIARYDATETIME));
+                let starttime = tableform.format_time_blank(d, d.DIARYDATETIME);
+                if (starttime) { s.push(' ' + starttime ); }
+                if ( !config.bool("DisableDiaryEndDatetime") ) {
+                    let endtime = tableform.format_time_blank(d, d.DIARYENDDATETIME);
+                    if (endtime) { s.push(_(" to ") + endtime); }
```
bobintetley: same again, "{0} to {1}"

src/static/js/options.js
```diff
@@ -942,6 +942,7 @@ $(function() {
                         { id: "disablerota", post_field: "DisableRota", label: _("Remove the rota functionality from menus and screens"), type: "check" }, 
                         { id: "disablestockcontrol", post_field: "DisableStockControl", label: _("Remove the stock control functionality from menus and screens"), type: "check" }, 
                         { id: "disabletransport", post_field: "DisableTransport", label: _("Remove the transport functionality from menus and screens"), type: "check" }, 
+                        { id: "disablediaryenddatetime", post_field: "DisableDiaryEndDatetime", label: _("Remove the diary note end datetime functionality from menus and screens"), type: "check" },
```
bobintetley: datetime is a bit programmer-y. "Remove the end date and time fields when displaying and editing diary notes"

src/static/js/diary.js
bobintetley: In addition to displaying the end date time here, it needs to be shown on the home page in main.js as well.

src/asm3/diary.py
adamdspencer77: I think it's already in there.

src/static/js/diary.js
adamdspencer77: Done.

src/static/js/diary.js
adamdspencer77: Updated.

src/static/js/options.js
adamdspencer77: Changed.

src/static/js/main.js
adamdspencer77: Sorted.

src/static/js/diary.js
```diff
@@ -23,6 +23,10 @@ $(function() {
                         callout: _("The color scheme to be used when displaying this note on the calendar and home screen") },
                     { json_field: "DIARYDATETIME", post_field: "diarydate", label: _("Date"), type: "date", validation: "notblank", defaultval: new Date() },
                     { json_field: "DIARYDATETIME", post_field: "diarytime", label: _("Time"), type: "time" },
+                    { json_field: "DIARYENDDATETIME", post_field: "diaryendtime", label: _("Until"), type: "time", hideif: function() {
+                            return config.bool("DisableDiaryEndDatetime");
+                        }
+                    },
```
bobintetley, line 28: Use the datetime type with tableform instead for both DIARYDATETIME and DIARYENDDATETIME and get rid of the separate date and time field for DIARYDATETIME. 

```
{ json_field: "DIARYDATETIME", post_field: "diary", label: _("Date"), type: "datetime" }, 
{ json_field: "DIARYENDDATETIME", post_field: "diaryend", label: _("Until"), type: "datetime"},
```

This will give a separate date and time for each field and allow for multiple day diary notes.

src/asm3/diary.py
bobintetley: Not seeing it in the patch though? :)

src/asm3/diary.py
```diff
@@ -372,6 +372,7 @@ def insert_diary_from_form(dbo: Database, username: str, linktypeid: int, linkid
         "LinkType":         linktypeid,
         "LinkInfo":         linkinfo,
         "DiaryDateTime":    post.datetime("diarydate", "diarytime"),
+        "DiaryEndDateTime": post.datetime("diarydate", "diaryendtime"),
```
bobintetley: this prevents multi-day notes, since the date portion is always from diarydate. If you make the change below to use datetime fields, you can use post.datetime("diaryenddate", "diaryendtime")

src/static/js/diary.js
```diff
@@ -41,6 +45,9 @@ $(function() {
                             try {
                                 tableform.fields_update_row(dialog.fields, row);
                                 row.COLOURSCHEMEID = $("#diarycolourscheme").selectcolour("value");
+                                let endtime = $("#diaryendtime").val();
+                                if (endtime) { endtime += ":00"; }
+                                row.DIARYENDDATETIME = format.date_iso($("#diarydate").val() + " " + endtime);
```
bobintetley: can construct from diaryenddate with the above change

src/asm3/diary.py
adamdspencer77: It's hiding from the patch notes :)

src/static/js/diary.js
adamdspencer77: Done.

src/static/js/diary.js
adamdspencer77: Hopefully sorted.

src/asm3/diary.py
adamdspencer77: Hopefully sorted.

src/static/js/diary.js
```diff
@@ -21,8 +21,11 @@ $(function() {
                     { json_field: "COLOURSCHEMEID", post_field: "diarycolourscheme", label: _("Color Scheme"), 
                         type: "selectcolour", defaultval: 1, 
                         callout: _("The color scheme to be used when displaying this note on the calendar and home screen") },
-                    { json_field: "DIARYDATETIME", post_field: "diarydate", label: _("Date"), type: "date", validation: "notblank", defaultval: new Date() },
-                    { json_field: "DIARYDATETIME", post_field: "diarytime", label: _("Time"), type: "time" },
+                    { json_field: "DIARYDATETIME", post_field: "diarydatetime", label: _("Date"), type: "datetime", validation: "notblank", defaultval: new Date() },
+                    { json_field: "DIARYENDDATETIME", post_field: "diaryenddatetime", label: _("Until"), type: "datetime", hideif: function() {
```
bobintetley: the datetime tableform fields appends date and time onto the end of the name you give it - you'll get more readable names if you called these two fields "diary" and "diaryend" - which will create diarydate, diarytime, diaryenddate and diaryendtime

src/asm3/diary.py
```diff
@@ -417,9 +419,9 @@ def update_diary_from_form(dbo: Database, username: str, post: PostedData) -> No
     Updates a diary note from form data
     """
     l = dbo.locale
-    if post["diarydate"] == "":
+    if post["diarydatetimedate"] == "":
         raise asm3.utils.ASMValidationError(asm3.i18n._("Diary date cannot be blank", l))
-    if post.date("diarydate") is None:
+    if post.date("diarydatetimedate") is None:
         raise asm3.utils.ASMValidationError(asm3.i18n._("Diary date is not valid", l))
```
bobintetley, line 425: If you use the suggested name scheme below (datetime field named "diary") then these don't have to change they will continue to be diarydate and diarytime.

src/asm3/diary.py
```diff
@@ -432,7 +434,8 @@ def update_diary_from_form(dbo: Database, username: str, post: PostedData) -> No
 
     diaryid = post.integer("diaryid")
     dbo.update("diary", diaryid, {
-        "DiaryDateTime":    post.datetime("diarydate", "diarytime"),
+        "DiaryDateTime":    post.datetime("diarydatetimedate", "diarydatetimetime"),
+        "DiaryEndDateTime": post.datetime("diaryenddatetimedate", "diaryenddatetimetime"),
```
bobintetley: again, these wouldn't have to change, and you can use diaryenddate and diaryendtime for the end one with the name suggested below

src/static/js/main.js
```diff
@@ -474,7 +474,21 @@ $(function() {
...
+                let dateoutput = startdate;
+                if ( !config.bool("DisableDiaryEndDatetime") ) {
+                    if ( starttime || endtime ) { dateoutput += " " + tableform.format_time(d, d.DIARYDATETIME); }
+                    if ( enddate && enddate != startdate ) {
+                        dateoutput = _("{0} to {1}").replace(/\{0\}/g, dateoutput).replace(/\{1\}/g, enddate);
+                        if (endtime) { dateoutput += " " + endtime; }
+                    } else if (enddate) {
+                        dateoutput = _("{0} to {1}").replace(/\{0\}/g, dateoutput).replace(/\{1\}/g, endtime);
+                    }
+                }
```
bobintetley: This code to format and display the start and end date/time for diaries is now copy and pasted in two places (here and diary.js). 

I would make this a generic function in the format object (which lives in common.js) - something like datetime_from_to = function(isostart, isoend) {}

We can then use it in both these places and anywhere else in future that we've got 2 iso dates and want to output 2 dates/times as a range. In this case here, the code above would be replaced with ``` dateoutput = format.datetime_from_to(d.DIARYDATETIME, d.DIARYENDDATETIME); ```

src/static/js/main.js
```diff
@@ -474,7 +474,21 @@ $(function() {
...
-                s.push('<td>' + format.date(d.DIARYDATETIME));
+                let startdate = format.date(d.DIARYDATETIME);
+                let starttime = tableform.format_time_blank(d, d.DIARYDATETIME);
+                let enddate = format.date(d.DIARYENDDATETIME);
+                let endtime = tableform.format_time_blank(d, d.DIARYENDDATETIME);
+                let dateoutput = startdate;
+                if ( !config.bool("DisableDiaryEndDatetime") ) {
+                    if ( starttime || endtime ) { dateoutput += " " + tableform.format_time(d, d.DIARYDATETIME); }
+                    if ( enddate && enddate != startdate ) {
+                        dateoutput = _("{0} to {1}").replace(/\{0\}/g, dateoutput).replace(/\{1\}/g, enddate);
```
bobintetley: You don't need to use regex here. You know there's only one {0} token in the string, so just use strings for readability and efficiency. 

``` dateoutput = _("{0} to {1}").replace("{0}", dateoutput).replace("{1}", enddate); ```

src/static/js/diary.js
adamdspencer77: Done.

src/asm3/diary.py
adamdspencer77: Done.

src/asm3/diary.py
adamdspencer77: Done.

src/static/js/main.js
adamdspencer77: Done.

src/static/js/main.js
adamdspencer77: Done.

adamdspencer77:
Plumbed in diary notes created by Bulk change animals while I was poking around.

PR #1949, opened 2025-11-17
Point of sale

adamdspencer77:
ASM should have an electronic point of sale interface. New tables

stockitem (name, stockitemcategoryid, description, price, barcode (string))
stockitemcategory (name, description)
salesreceipt (salesshopid, date, total)
salesreceiptdetail (salesreceiptid, date, stockitem, price)
salesshop (name, description)

Have a till type interface (needs investigation) to select/scan items and produce a total

android barcode scanning? can call a custom url so we could have a service call submit barcode data with the current receipt

the stock tables have existed for years at this point, it's the sales recording that we're missing. A sales stockusage record should be created for each item sold and that stockusage should have a link to a salesreceiptdetail ID. I have no idea what I was thinking of with the salesshop table - this issue is about 5 years old.


src/asm3/html.py
```diff
@@ -631,6 +631,7 @@ def menu_structure(l: str, publisherlist: Dict, reports: MenuItems, mailmerges:
             ( asm3.users.VIEW_DONATION, "", "tagstock", "--cat", "", _("Stock control", l) ),
             ( asm3.users.VIEW_STOCKLEVEL, "", "tagstock", "stock_level", "asm-icon-stock", _("Stock levels", l) ),
             ( asm3.users.VIEW_STOCKLEVEL, "", "tagstock", "stock_usage", "asm-icon-stock-usage", _("Stock usage", l) ),
+            ( asm3.users.VIEW_STOCKLEVEL, "", "tagstock", "pos", "asm-icon-transactions", _("Point of Sale", l) ),
```
bobintetley: Add another class to this menu item, and hide it based on a new config value for DisablePOS 

This config value should be on by default, and while it should have an option in the remove tab of options, I would leave it commented out for now so that users can't enable it until we're ready.

src/static/css/asm.css
```diff
@@ -1131,3 +1131,361 @@ select:disabled, input:disabled, textarea:disabled {
 {
     .no-print, .no-print * { display: none !important; }
 }
+
```
bobintetley: There's so much CSS for the POS that I think it might be better moved into its own file at static/css/asm-pos.css

We can add another directive to load the new css to the html.py/bare_header function

src/asm3/html.py
adamdspencer77: Hopefully sorted. I'd made a tab within the options to define locations related to stock movements via the POS. Rather than commenting it out completely I've linked it to the same config value.

src/static/css/asm.css
adamdspencer77: Hopefully sorted.

PR #1955, opened 2025-12-02
Video uploads

adamdspencer77:
With the amount of work that has been done to push file storage off to a CDN asynchronously, it would now be possible for us to support direct video uploads where we could not previously.

Places affected by this that I can think of off the top of my head:

media.js (support file type and size limit, 10MB? Default video indicator)
media.py (handle storage, may not need anything doing)
petrescue.py (include preferred video in photo_urls)
service.py (animal_video call needs implementing to get the preferred)
dbfs.py (local cache time should be shorter for videos, eg: 2 days)

SM service handler wants to make sure the CDN is caching video downloads served via the animal_video service call with a decent ttl.

Unrelated directly to the task, but we could also look at reviewing our storage costs/rates/tiers and make it a bit cheaper.

For simplicity, maybe only accept mp4 container videos with a file size limit of 10MB per video. Probably make accepting video an option that is off by default since it will make it easy for people to run up storage bills quickly.


src/asm3/dbfs.py
```diff
@@ -167,7 +167,8 @@ def _cache_ttl(self, name: str) -> int:
         #name = name.lower()
         #if name.endswith(".jpg") or name.endswith(".jpeg"): return (86400 * 7) # Cache images for a week
         #return (86400 * 2) # Cache everything else for two days
-        return (86400 * 7) # Cache everything for 1 week
+        if name.endswith(".mp4"): return (86400 * 2) # Cache images for 2 days
```
bobintetley: surely the comment should be videos? :D

src/asm3/media.py
```diff
@@ -292,6 +298,100 @@ def thumb_mrec(mm):
...
         return nopic()
 
+def get_video_file_data(dbo: Database, mode: str, iid: str = "", seq: int = 0, justdate: bool = False) -> Tuple[datetime, bytes]:
+    """
+    Gets a video
+    mode: animal | media | animalthumb | person | personthumb | dbfs
+    iid: (str) The id of the animal for animal/thumb mode or the media record
+        or a template path for dbfs mode
+    seq: (int) If the mode is animal or person, returns image X for that person/animal
+         The first image is always the preferred photo and seq is 1-based.
```
bobintetley: comments should be referring to video?

src/asm3/media.py
```diff
@@ -292,6 +298,100 @@ def thumb_mrec(mm):
...
+        if justdate: return NOPIC_DATE
+        return (NOPIC_DATE, b"NOPIC")
+    def mrec(mm):
+        if mm is None: return nopic()
+        if justdate: return mm.DATE
+        return (mm.DATE, asm3.dbfs.get_string_id(dbo, mm.DBFSID))
+    def thumb_mrec(mm):
+        if mm is None: return thumb_nopic()
+        if justdate: return mm.DATE
+        return (mm.DATE, scale_image(asm3.dbfs.get_string_id(dbo, mm.DBFSID), asm3.configuration.thumbnail_size(dbo)))
```
bobintetley: should this not be calling your new video_thumbnail function?

src/asm3/media.py
```diff
@@ -746,6 +847,19 @@ def get_signature_link(dbo: Database, mid: int, post: PostedData) -> str:
...
     return url
 
+def get_video_thumbnail(dbo: Database, dbfsid: int) -> bytes:
+    inputfile = tempfile.NamedTemporaryFile(suffix=".mp4", delete=False)
+    vdata = asm3.dbfs.get_string_id(dbo, dbfsid)
+    inputfile.write(vdata)
+    inputfile.close()
+    outputfile = tempfile.NamedTemporaryFile(suffix=".jpg", delete=False)
+
+    os.system(VIDEO_THUMBNAIL_CMD % { "output": outputfile.name, "input": inputfile.name})
```
bobintetley: Use asm3.utils.cmd instead of os.system - it gives you a few more options and allows you to capture the output to look for known errors

src/asm3/publishers/petrescue.py
```diff
@@ -298,6 +298,9 @@ def processAnimal(self, an: ResultRow, all_desexed=False, adoptable_in="",
             location_postcode.startswith("2") or location_postcode.startswith("3")):
             microchip_number = asm3.utils.iif(an.IDENTICHIPPED == 1, an.IDENTICHIPNUMBER, "")
 
+        photourls = self.getPhotoUrls(an.ID) # List of photo URL strings
+        if VIDEO_ENABLED:
+            photourls.append(self.getVideoUrls(an.ID))
```
bobintetley: I think this will add one element to photo urls that contains another list. To join the two lists together, just do ``` photourls = photourls + self.getVideoUrls(an.ID) ```

src/asm3/media.py
```diff
@@ -292,6 +298,100 @@ def thumb_mrec(mm):
...
+                return (dbo.now(), asm3.dbfs.get_string_filepath(dbo, sid))
+            else:
+                # Only name was given
+                return (dbo.now(), asm3.dbfs.get_string(dbo, sid))
+
+    elif mode == "nopic":
+        if asm3.dbfs.file_exists(dbo, "nopic.jpg"):
+            return (dbo.now(), asm3.dbfs.get_string_filepath(dbo, "/reports/nopic.jpg"))
+        else:
+            return (dbo.now(), asm3.utils.read_binary_file(dbo.installpath + "media/reports/nopic.jpg"))
```
bobintetley: might be worth adding a novid.jpg for clarity

src/asm3/dbfs.py
adamdspencer77: Sorted.

src/asm3/media.py
adamdspencer77: Sorted.

src/asm3/media.py
adamdspencer77: Hopefully sorted.

src/asm3/media.py
adamdspencer77: Sorted.

src/asm3/publishers/petrescue.py
adamdspencer77: Done.

src/asm3/media.py
adamdspencer77: Done.

src/asm3/media.py
```diff
@@ -292,6 +298,100 @@ def thumb_mrec(mm):
...
+        or a template path for dbfs mode
+    seq: (int) If the mode is animal or person, returns video X for that person/animal
+         The first video is always the preferred video and seq is 1-based.
+    if justdate is True, returns the last modified date
+    if justdate is False, returns a tuple containing the last modified date and video data
+    """
+    def nopic():
+        NOPIC_DATE = datetime(2011, 1, 1)
+        if justdate: return NOPIC_DATE
+        return (NOPIC_DATE, b"NOPIC")
```
bobintetley: I'd call these novid()/NOVID_DATE etc for clarity

src/asm3/media.py
```diff
@@ -292,6 +298,100 @@ def thumb_mrec(mm):
...
+                return (dbo.now(), asm3.dbfs.get_string(dbo, sid))
+
+    elif mode == "nopic":
+        if asm3.dbfs.file_exists(dbo, "novid.jpg"):
+            return (dbo.now(), asm3.dbfs.get_string_filepath(dbo, "/reports/novid.jpg"))
+        else:
+            return (dbo.now(), asm3.utils.read_binary_file(dbo.installpath + "media/reports/novid.jpg"))
+
+    else:
+        return nopic()
```
bobintetley: novid()

src/asm3/media.py
```diff
@@ -746,6 +847,17 @@ def get_signature_link(dbo: Database, mid: int, post: PostedData) -> str:
...
     return url
 
+def get_video_thumbnail(dbo: Database, dbfsid: int) -> bytes:
+    inputfile = tempfile.NamedTemporaryFile(suffix=".mp4", delete=False)
+    vdata = asm3.dbfs.get_string_id(dbo, dbfsid)
+    inputfile.write(vdata)
+    inputfile.close()
+    outputfile = tempfile.NamedTemporaryFile(suffix=".jpg", delete=False)
+    asm3.utils.cmd(VIDEO_THUMBNAIL_CMD % { "output": outputfile.name, "input": inputfile.name}, False)
+    idata = outputfile.read()
```
bobintetley, line 864: I think you still need to remove the temporary files after use since delete=False, you need to call os.unlink on inputfile.name and outputfile.name before leaving the function

src/asm3/sitedefs.py
```diff
@@ -410,3 +410,7 @@ def get_dict(k: str, dv: Dict = {}) -> Dict:
 # Toggle for enabling AMQP messages from audit log calls
 AMQP_ENABLED = get_boolean("amqp_enabled", False)
 
+# Max video upload size in bytes
+VIDEO_ENABLED = get_boolean("video_enabled", False)
+VIDEO_SIZE_LIMIT = get_integer("video_size_limit", 10485760)
+VIDEO_THUMBNAIL_CMD = get_string("video_thumbnail_cmd", "ffmpeg -v 0 -y -an -i %(input)s -ss 00:00:2.000 -vframes 1 %(output)s")
```
bobintetley, line 416: I'd also add this to the asm3.conf.example file in /scripts

src/asm3/media.py
```diff
@@ -746,6 +847,17 @@ def get_signature_link(dbo: Database, mid: int, post: PostedData) -> str:
     url = "%s?account=%s&method=sign_document&email=%s&formid=%d&token=%s" % (SERVICE_URL, dbo.name(), asm3.utils.strip_email_address(post["to"]).replace("@", "%40"), mid, token)
     return url
 
+def get_video_thumbnail(dbo: Database, dbfsid: int) -> bytes:
+    inputfile = tempfile.NamedTemporaryFile(suffix=".mp4", delete=False)
+    vdata = asm3.dbfs.get_string_id(dbo, dbfsid)
+    inputfile.write(vdata)
+    inputfile.close()
+    outputfile = tempfile.NamedTemporaryFile(suffix=".jpg", delete=False)
+    asm3.utils.cmd(VIDEO_THUMBNAIL_CMD % { "output": outputfile.name, "input": inputfile.name}, False)
```
bobintetley: The return code and/or output should be checked here, if the return code is non-zero then we know something went wrong and can raise one of our HTTP flavoured exceptions for web.py to catch, eg: ``` raise asm3.utils.ASMError("Failed creating video thumbnail") ```

src/asm3/media.py
```diff
@@ -746,6 +847,17 @@ def get_signature_link(dbo: Database, mid: int, post: PostedData) -> str:
     url = "%s?account=%s&method=sign_document&email=%s&formid=%d&token=%s" % (SERVICE_URL, dbo.name(), asm3.utils.strip_email_address(post["to"]).replace("@", "%40"), mid, token)
     return url
 
+def get_video_thumbnail(dbo: Database, dbfsid: int) -> bytes:
+    inputfile = tempfile.NamedTemporaryFile(suffix=".mp4", delete=False)
+    vdata = asm3.dbfs.get_string_id(dbo, dbfsid)
```
bobintetley, line 857: While we're currently capped at 10MB per video, there's nothing to stop it being larger in future. Every time a user visits the media tab of a record that contains videos, this is going to cause us to pull the whole video file out of object storage so that we can generate a thumbnail for it. This could cause performance issues as no doubt we will get at least one customer who stuffs animals with 50 videos and the cache time for the video file in the DBFS is only 2 days.

Since these thumbnails are small and video files are immutable, I would suggest that this function uses asm3.cachedisk to cache thumbnails we've generated for a dbfsid and keep them around for a week, but use the "touch" method of cachedisk when retrieving. This updates the ttl every time you read the thumbnail so it will stay around for another week after reading.

So, check the cache first to see if we already have a thumbnail for this dbfsid and if not, generate it and put it in the cache for next time.

```
cache_ttl = 86400 * 7
cache_key = f"vid_thumbnail_{dbfsid}"
idata = asm3.cachedisk.touch(cache_key, dbo.name(), cache_ttl)
if idata is not None: return idata

... get the thumbnail ...

asm3.cachedisk.put(cache_key, dbo.name(), idata, cache_ttl)
```

src/main.py
```diff
@@ -671,6 +671,12 @@ def content(self, o):
             if o.post["nopic"] == "404": self.notfound()
             self.redirect("image?db=%s&mode=nopic" % o.dbo.name())
 
+class video_thumbnail(ASMEndpoint):
+    url = "video_thumbnail"
+
+    def content(self, o):
+        return asm3.media.get_video_thumbnail(o.dbo, o.post.integer("dbfsid"))
```
bobintetley, line 689: Need to be setting a content-type header here and cache control hints so that the user's browser will not ask for thumbnails it has seen before. Given that videos are immutable, and that we're always going to pass the db=name parameter to this endpoint, the thumbnail can be very long lived in the browser. Eg:

```
self.content.type("video/mp4")
self.cache_control(CACHE_ONE_MONTH, CACHE_ONE_DAY)
```

Normally I'd say check the return value before setting the cache control header (like the image() endpoint does), but in this case, we're going to make get_video_thumbnail throw ASMError if it can't generate the thumbnail, so it's ok to rely on web.py to handle it if something goes wrong as it will catch the Exception and translate it into a 500 with our message and we will never cache broken thumbnails

src/main.py
```diff
@@ -671,6 +671,12 @@ def content(self, o):
             if o.post["nopic"] == "404": self.notfound()
             self.redirect("image?db=%s&mode=nopic" % o.dbo.name())
 
+class video_thumbnail(ASMEndpoint):
+    url = "video_thumbnail"
```
bobintetley, line 683: We want this endpoint to be cacheable by Cloudflare, but they won't cache anything that has a session cookie in the response. Set ```session_cookie = False``` here

src/asm3/media.py
```diff
@@ -292,6 +298,100 @@ def thumb_mrec(mm):
...
+            return dbo.now()
+        else:
+            if sid.startswith("/"):
+                # Complete path was given
+                return (dbo.now(), asm3.dbfs.get_string_filepath(dbo, sid))
+            else:
+                # Only name was given
+                return (dbo.now(), asm3.dbfs.get_string(dbo, sid))
+
+    elif mode == "nopic":
```
bobintetley: mode should be novid, although there's nothing really using any of this function right now apart from animal mode :)

src/asm3/media.py
adamdspencer77: Hopefully sorted.

src/asm3/media.py
adamdspencer77: Changed.

src/asm3/media.py
adamdspencer77: Hopefully sorted.

src/asm3/sitedefs.py
adamdspencer77: Hopefully sorted.

src/asm3/media.py
adamdspencer77: Done.

src/asm3/media.py
adamdspencer77: Hopefully sorted.

src/main.py
adamdspencer77: Done.

src/main.py
adamdspencer77: Done.

src/asm3/media.py
adamdspencer77: Done.

PR #1957, opened 2025-12-03
Address lookup in online forms

adamdspencer77:
Similar to how we have on person objects, a postcode/address lookup for UK online forms would be handy.

### DIFF
--- a/src/static/js/onlineform_extra.js
+++ b/src/static/js/onlineform_extra.js
@@ -622,7 +622,7 @@ $(document).ready(function() {
     } catch(err) {
         log.error(err, err);
     }
-    SMCOM = true;
+    
     if (SMCOM && ["en_CA", "fr_CA", "en_GB", "en_IE", "nl"].includes(LOCALE) && $(".asm-onlineform-postcode").length && $(".asm-onlineform-address").length) {
         $(".asm-onlineform-postcode").after('&nbsp;<span id="postcodelookup"><img src="/static/images/icons/find.png" style="height: 15px;cursor: pointer;"></span>');
         $("#postcodelookup").click(function() {


src/asm3/onlineform.py
```diff
@@ -233,9 +234,14 @@ def get_onlineform_html(dbo: Database, formid: int, completedocument: bool = Tru
                 '<label class="asm-onlineform-checkboxlabel" for="%s">%s</label>' % \
                 (fid, cname, required, fid, f.LABEL))
         elif f.FIELDTYPE == FIELDTYPE_TEXT:
-            h.append('<input class="asm-onlineform-text" type="text" id="%s" name="%s" %s %s />' % ( fid, cname, autocomplete, requiredtext))
+            if f.FIELDNAME == "postcode" or f.FIELDNAME == "zipcode": extraclass = " asm-onlineform-postcode"
+            elif f.FIELDNAME == "address": extraclass = " asm-onlineform-address"
+            elif f.FIELDNAME == "town": extraclass = " asm-onlineform-town"
+            elif f.FIELDNAME == "country": extraclass = " asm-onlineform-county"
+            h.append(f'<input class="asm-onlineform-text{extraclass}" type="text" id="%s" name="%s" %s %s />' % ( fid, cname, autocomplete, requiredtext))
```
bobintetley: minor thing, but I'd rather the space was outside extraclass for readability. 
```class="asm-onlineform-text {extraclass}"```

File: src/static/js/onlineform_extra.js [hunk 2]
```diff
@@ -614,5 +614,51 @@ $(document).ready(function() {
...
 
-});
+    let ro = new ResizeObserver(function(e) {
+            window.parent.postMessage(document.querySelector("html").offsetHeight, "*");
+    });
+    ro.observe(document.querySelector("html"));
+
+    if (["en_CA", "fr_CA", "en_GB", "en_IE", "nl"].includes(LOCALE) && $(".asm-onlineform-postcode").length && $(".asm-onlineform-address").length) {
+        $(".asm-onlineform-postcode").after('&nbsp;<span id="postcodelookup"><img src="/static/images/icons/find.png" style="height: 15px;"></span>');
+        $("#postcodelookup").click(async function() {
```
bobintetley: You aren't actually using await in here, so there's no need for this function to be declared async - and async still isn't supported by some older browsers (which we aim for compatibility with in this one particular file)

src/static/js/onlineform_extra.js
```diff
@@ -614,5 +614,51 @@ $(document).ready(function() {
...
+            }
+            let formdata = "mode=postcodelookup&country=" + country + "&postcode=" + postcode + "&locale=" + LOCALE + "&account=" + USERACCOUNT;
+            $.ajax({
+                type: "POST",
+                url:  "person_embed",
+                data: formdata,
+                dataType: "text",
+                success: function(response) {
+                    let rows = jQuery.parseJSON(response);
+                    console.log(rows);
```
bobintetley, line 654: debug code left in :)

src/static/js/onlineform_extra.js
```diff
@@ -614,5 +614,51 @@ $(document).ready(function() {
         }
     });
 
-});
+    let ro = new ResizeObserver(function(e) {
+            window.parent.postMessage(document.querySelector("html").offsetHeight, "*");
+    });
+    ro.observe(document.querySelector("html"));
```
bobintetley: I'd wrap this resizeobserver stuff in a try/catch block and log errors to the console or something - just in case someone is still using old Safari or IE or something where it isn't supported

src/asm3/onlineform.py

src/static/js/onlineform_extra.js
adamdspencer77: Removed.

src/static/js/onlineform_extra.js
adamdspencer77: Removed.

src/static/js/onlineform_extra.js
adamdspencer77: Hopefully sorted.

src/asm3/onlineform.py
```diff
@@ -233,9 +234,14 @@ def get_onlineform_html(dbo: Database, formid: int, completedocument: bool = Tru
                 '<label class="asm-onlineform-checkboxlabel" for="%s">%s</label>' % \
                 (fid, cname, required, fid, f.LABEL))
         elif f.FIELDTYPE == FIELDTYPE_TEXT:
-            h.append('<input class="asm-onlineform-text" type="text" id="%s" name="%s" %s %s />' % ( fid, cname, autocomplete, requiredtext))
+            if f.FIELDNAME == "postcode" or f.FIELDNAME == "zipcode": extraclass = "asm-onlineform-postcode"
+            elif f.FIELDNAME == "address": extraclass = "asm-onlineform-address"
+            elif f.FIELDNAME == "town": extraclass = "asm-onlineform-town"
+            elif f.FIELDNAME == "country": extraclass = "asm-onlineform-county"
```
bobintetley: class should asm-onlineform-country ?

src/static/js/onlineform_extra.js
```diff
@@ -614,5 +614,54 @@ $(document).ready(function() {
...
+        let ro = new ResizeObserver(function(e) {
+                window.parent.postMessage(document.querySelector("html").offsetHeight, "*");
+        });
+        ro.observe(document.querySelector("html"));
+    } catch(err) {
+        log.error(err, err);
+    }
+
+    if (["en_CA", "fr_CA", "en_GB", "en_IE", "nl"].includes(LOCALE) && $(".asm-onlineform-postcode").length && $(".asm-onlineform-address").length) {
+        $(".asm-onlineform-postcode").after('&nbsp;<span id="postcodelookup"><img src="/static/images/icons/find.png" style="height: 15px;"></span>');
```
bobintetley: postcode lookup can only work in SM databases since it relies on our smcom_geo service. We should only inject the button if this is an SM database (set a new global variable SMCOM like we are doing with LOCALE and use asm3.smcom.active() to set it to true or false)

src/static/js/onlineform_extra.js
```diff
@@ -614,5 +614,54 @@ $(document).ready(function() {
...
+            
+            let postcode = $(".asm-onlineform-postcode").val();
+            if (!postcode) {
+                $("#postcodelookup img").attr("src", "/static/images/icons/find.png");
+                return;
+            }
+            let formdata = "mode=postcodelookup&country=" + country + "&postcode=" + postcode + "&locale=" + LOCALE + "&account=" + USERACCOUNT;
+            $.ajax({
+                type: "POST",
+                url:  "person_embed",
```
bobintetley: I don't know why check_logged_in is set to False on the person_embed endpoint, but I don't think it should be. person_embed should definitely be only accessible by logged in users since it requires view person permissions for almost all of its features.

I think rather than using person_embed, we should have a postcode_lookup endpoint in main.py that does not require us to be logged in, and that this code should be calling. Making it separate from person_embed means that we can add rate limit controls or better cache later if we find it is getting attacked/flooded. It will still be doing the same thing though :)

src/asm3/onlineform.py
adamdspencer77: I think it was the country that needed to be county. Changed.

src/static/js/onlineform_extra.js
adamdspencer77: Done.

src/static/js/onlineform_extra.js
adamdspencer77: Done.

src/asm3/onlineform.py
adamdspencer77: County and Country now (hopefully) properly represented.

PR #1961, opened 2025-12-09
New Forms menu to show links to online forms flagged as internal

adamdspencer77:
Create a new "Forms" menu to the left of Mail on the desktop UI.

This menu should disappear if there are no online forms that are flagged as internal. Otherwise, it shows a dropdown of clickable links to internal online forms to complete them.

### DIFF
--- a/src/static/js/animal.js
+++ b/src/static/js/animal.js
@@ -352,20 +352,20 @@ $(function() {
             ];
             
             let typemap = {
-                "AIL": [ _("Littermate"), "animal?id=" ],
-                "AFA": [ "", "animal?id=" ],
-                "AFP": [ "", "person?id=" ],
-                "AFW": [ "", "waitinglist?id=" ]
+                "AIL": [ _("Littermate"), "animal?id={linkid}" ],
+                "AFA": [ "", "animal?id={linkid}" ],
+                "AFP": [ "", "person?id={linkid}" ],
+                "AFW": [ "", "waitinglist?id={linkid}" ],
+                "APL": [ "", "animal_find_results?mode=ADVANCED&filter=includedeceased&litterid={linkdisplay}" ]
             };
 
             $.each(controller.links, function(i, v) {
                 h.push('<tr>');
                 h.push('<td>' + v.TYPEDISPLAY + '</td>');
-                if (v.TYPE == "APL") {
-                    h.push('<td>' + v.LINKDISPLAY + '</td>');
-                } else {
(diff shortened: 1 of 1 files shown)


src/asm3/reports.py
```diff
@@ -732,6 +732,21 @@ def get_mailmerges_menu(dbo: Database, roleids: str = "", superuser: bool = Fals
...
+    our menu structure.
+    The return value is a list of online forms with a tuple containing URL and
+    name.
+    roleids: comma separated list of roleids for the current user
+    superuser: true if the user is a superuser
+    """
+    mv = []
+    forms = dbo.query("SELECT ID, Name FROM onlineform WHERE InternalUse = 1 ORDER BY Name")
+    for form in forms:
+        mv.append( ( asm3.users.VIEW_ONLINE_FORMS, "", "", "/onlineform_view?formid=%d" % form.ID, "", form.NAME ) )
```
bobintetley, line 747: leave off the leading forward slash - otherwise you'll break installations where the system isn't at the root of the domain (which is virtually all open source installs with an asm3 prefix!)

src/static/js/header.js
```diff
@@ -230,21 +230,24 @@ header = {
...
-                menus.push("<div id=\"asm-menu-" + name + "-body\" class=\"asm-menu-body\">");
-                // If the option is on or there are more than 120 items to show, 
-                // render report and mail merge menus in accordions by category instead
-                if ((config.bool("ReportMenuAccordion") || items.length > 120) && (name == "reports" || name == "mailmerge")) {
-                    menus.push('<input id="search-' + name + '" type="text" placeholder="' + _("Filter") + '" class="asm-menu-filter asm-textbox">');
-                    self.menu_html_accordion_renderer(menus, items, name);
-                } else if (name == "reports" || name == "mailmerge") {
-                    menus.push('<input id="search-' + name + '" type="text" placeholder="' + _("Filter") + '" class="asm-menu-filter asm-textbox">');
-                    self.menu_html_flat_renderer(menus, items, { breakafter: 25 });
+                if (name == "internalforms" && items.length == 0) {
```
bobintetley, line 233: Really don't like this empty if block with everything else being covered by the else. I think it would be clearer if there being no items short-circuited the function instead of putting everything in an else block.

```
$.each(asm.menustructure, function(im, vm) {
    let permission = vm[0], name = vm[1], display = vm[2], items = vm[3];
    if (name == "internalforms" && items.length == 0) { return; }
    if (asm.superuser || asm.securitymap.indexOf(permission + " ") != -1) {
        ...
```

PR #1962, opened 2025-12-09
Medical Type: fluid treatment

adamdspencer77:
Add a new medical type for "fluid treatment" (or possibly fluid therapy, or just fluids, see what looks right, something that indicates rehydration).

### DIFF
--- a/src/asm3/dbupdates/35115.py
+++ b/src/asm3/dbupdates/35115.py
@@ -1,0 +1,8 @@
+from asm3.dbupdate import execute
+
+dbo = dbo
+l = dbo.locale
+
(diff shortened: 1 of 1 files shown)


src/asm3/dbupdates/35115.py
```diff
@@ -0,0 +1,8 @@
+from asm3.dbupdate import execute
+
+dbo = dbo
+l = dbo.locale
+
+## Add fluids to lksmedicaltype
```
bobintetley, line 6: Future reference, the flake8 lint tool that we use really does not like comments starting with ## instead of a single #

PR #1968, opened 2025-12-15
Regular debits in accounting package

adamdspencer77:
It'd be great if we could add some info for regular debits:
day to create, day of month, description, amount, from account, to account
Then the daily batch job could check if the day to create is today and do it.

### DIFF
--- a/src/asm3/financial.py
+++ b/src/asm3/financial.py
@@ -1406,11 +1406,6 @@ def update_regulardebit_from_form(dbo: Database, username: str, post: PostedData
         "Comments":         post["comments"]
     }, username)
 
-    if post.integer("person"):
-        concatsql = dbo.sql_concat(("OwnerName", "' '", "OwnerCode"))
-        return dbo.query_string(f"SELECT {concatsql} AS OwnerName FROM owner WHERE ID = ?", [post.integer("person")])
-
-
 def delete_regulardebit(dbo: Database, username: str, rdid: int) -> None:
     """
     Deletes a regular debit
--- a/src/static/js/regular_debits.js
+++ b/src/static/js/regular_debits.js
@@ -79,8 +79,8 @@ $(function() {
                         },
                         onchange: async function() {
                             tableform.fields_post(dialog.fields, "mode=update&id=" + row.ID, "regular_debits")
-                                .then(function(response) {
-                                    row.OWNERNAME = response;
+                                .then(function() {
+                                    row.OWNERNAME = $("#personrow .asm-embed-name").text();
                                     tableform.fields_update_row(dialog.fields, row);
                                     tableform.table_update(table);
                                     tableform.dialog_close();
@@ -158,6 +158,7 @@ $(function() {
                         let response = await tableform.fields_post(dialog.fields, "mode=create", "regular_debits");
                         let row = {};
                         row.ID = response;
+                        row.OWNERNAME = $("#personrow .asm-embed-name").text();
                         tableform.fields_update_row(dialog.fields, row);
                         controller.rows.push(row);
                         tableform.table_update(table);
@@ -192,23 +193,10 @@ $(function() {
         },
 
         period_changed: function() {
-            if ($("#period").val() == 1) {
-                $("#weekdayrow").hide();
-                $("#dayofmonthrow").hide();
-                $("#monthrow").hide();
-            } else if ($("#period").val() == 7) {
-                $("#weekdayrow").show();
-                $("#dayofmonthrow").hide();
-                $("#monthrow").hide();
-            } else if ($("#period").val() == 30) {
-                $("#weekdayrow").hide();
-                $("#dayofmonthrow").show();
-                $("#monthrow").hide();
-            } else {
-                $("#weekdayrow").hide();
-                $("#dayofmonthrow").show();
-                $("#monthrow").show();
-            }
+            let period = $("#period").val();
+            $("#weekdayrow").toggle( period == 7 );
+            $("#dayofmonthrow").toggle( period == 30 || period == 365 );
+            $("#monthrow").toggle( period == 365 );
         },
 
         validation: function() {


src/asm3/financial.py
```diff
@@ -685,6 +715,17 @@ def get_recent_licences(dbo: Database) -> Results:
         "WHERE ol.IssueDate >= ? " \
         "ORDER BY ol.IssueDate DESC", [dbo.today(offset=-30)])
 
+def get_regulardebits(dbo: Database, filter="all") -> Results:
+    concatsql = dbo.sql_concat(("OwnerName", "' '", "OwnerCode"))
+    wheresql = ""
+    today = asm3.i18n.format_date(dbo.today())
+    if filter == "active":
+        wheresql = f" WHERE '{today}' BETWEEN rd.StartDate AND rd.EndDate OR ( rd.StartDate <= '{today}' AND rd.EndDate IS NULL )"
```
bobintetley: While i18n.format_date does default to ISO8601 YYYY-MM-DD, I don't think it includes the time info in the output and that will cause errors when this runs on PostgreSQL.

Use ```today = dbo.sql_value(dbo.today())``` instead to get a date parameter for use in queries. It adds the quoting too, so you don't need to quote {today} in your query if you're using dbo.sql_value(). Behind the scenes, it calls dbo.sql_date() which specifically handles dates.

dbo.sql_value() is a helper function that looks at the type of variable you give it and outputs something appropriate for use in a query. It's what you should use when you can't use a normal ? query parameter for whatever reason

src/asm3/financial.py
```diff
@@ -685,6 +715,17 @@ def get_recent_licences(dbo: Database) -> Results:
...
+    elif filter == "inactive":
+        wheresql = f" WHERE rd.StartDate > '{today}' OR rd.EndDate < '{today}'"
+    sql = "SELECT rd.*, " + concatsql + " AS OwnerName FROM regulardebit rd LEFT JOIN owner o ON rd.OwnerID = o.ID" + wheresql
```
bobintetley: Use the f string syntax like above :), sql = f"SELECT rd.*, {concatsql} AS OwnerName ...

src/asm3/financial.py
```diff
@@ -1321,6 +1362,58 @@ def delete_account(dbo: Database, username: str, aid: int) -> None:
     dbo.delete("accountsrole", "AccountID=%d" % aid)
     dbo.delete("accounts", aid, username)
 
+def insert_regulardebit_from_form(dbo: Database, username: str, post: PostedData) -> int:
+    """
+    Creates a regular debit from posted form data 
+    """
+
+    rdid = dbo.insert("regulardebit", {
```
bobintetley: teeny thing, but since you aren't using rdid for anything, why not just ```return dbo.insert``` ?

src/asm3/financial.py
```diff
@@ -1321,6 +1362,58 @@ def delete_account(dbo: Database, username: str, aid: int) -> None:
...
+        "Period":           post.integer("period"),
+        "Weekday":          post.integer("weekday"),
+        "DayOfMonth":       post.integer("dayofmonth"),
+        "Month":            post.integer("month"),
+        "Comments":         post["comments"]
+    }, username)
+
+    if post.integer("person"):
+        concatsql = dbo.sql_concat(("OwnerName", "' '", "OwnerCode"))
+        return dbo.query_string("SELECT " + concatsql + " AS OwnerName FROM owner WHERE ID = ?", [post.integer("person")])
```
bobintetley: f strings :D

src/asm3/dbupdate.py
```diff
@@ -1751,6 +1751,20 @@ def fclob(name: str, nullable: bool = True) -> str:
     sql += index("publishlog_PublishDateTime", "publishlog", "PublishDateTime")
     sql += index("publishlog_Name", "publishlog", "Name")
 
+    sql += table("regulardebit", (
+        fid(),
+        fint("OwnerID"),
+        fdate("StartDate"),
+        fdate("EndDate", True),
```
bobintetley, line 1758: given there are queries that filter this table by StartDate and EndDate, it would make sense to have indexes on these columns, both here and in the update script

src/asm3/financial.py
```diff
@@ -46,6 +46,36 @@
 ASCENDING = 0
 DESCENDING = 1
 
+def create_trx_from_regular_debits(dbo: Database, username: str) -> List[int]:
```
bobintetley, line 49: a docstring here would be helpful, just to indicate what the function does and that it's intended to be called once per day as part of the overnight batch

src/asm3/financial.py
```diff
@@ -46,6 +46,36 @@
...
+        elif rd.PERIOD == 30 and rd.DAYOFMONTH == today.day:
+            debitdue = True
+        elif rd.PERIOD == 365 and rd.MONTH == today.month and rd.DAYOFMONTH == today.day:
+            debitdue = True
+        
+        if debitdue:
+            trxlist.append(
+                dbo.insert("accountstrx", {
+                    "TrxDate":              today,
+                    "Description":          asm3.i18n._("Transaction created automatically from a regular debit."),
```
bobintetley: The description to put on the transaction should be part of the regulardebit table. I think I assumed this is what the comments column was for but I guess it makes sense for the user to be able to set both a comment about the regular debit and the description for the transaction. 

If the regular transaction is for say the gas bill going out every month, then the user should be able to say they want a description of gas bill or whatever on it!

src/asm3/financial.py
adamdspencer77: Cool. Done.

src/asm3/financial.py
adamdspencer77: Done.

src/asm3/financial.py
adamdspencer77: Done. I think a was monitoring rdid during testing :)

src/asm3/financial.py
adamdspencer77: Done.

src/asm3/dbupdate.py
adamdspencer77: Done.

src/asm3/financial.py
adamdspencer77: Done.

src/asm3/financial.py
adamdspencer77: Done.

src/static/js/regular_debits.js
```diff
@@ -0,0 +1,275 @@
...
+                edit: function(row) {
+                    tableform.fields_populate_from_json(dialog.fields, row);
+                    tableform.dialog_show_edit(dialog, row, {
+                        onvalidate: function() {
+                            return regular_debits.validation();
+                        },
+                        onchange: async function() {
+                            tableform.fields_post(dialog.fields, "mode=update&id=" + row.ID, "regular_debits")
+                                .then(function(response) {
+                                    row.OWNERNAME = response;
```
bobintetley: it shouldn't be necessary to return the ownername field from the update - you can just grab it from the person widget instead.

$("#person").personchooser("get_selected").OWNERNAME

src/asm3/financial.py
```diff
@@ -1321,6 +1365,58 @@ def delete_account(dbo: Database, username: str, aid: int) -> None:
...
+        "Weekday":          post.integer("weekday"),
+        "DayOfMonth":       post.integer("dayofmonth"),
+        "Month":            post.integer("month"),
+        "Description":      post["description"],
+        "Comments":         post["comments"]
+    }, username)
+
+    if post.integer("person"):
+        concatsql = dbo.sql_concat(("OwnerName", "' '", "OwnerCode"))
+        return dbo.query_string(f"SELECT {concatsql} AS OwnerName FROM owner WHERE ID = ?", [post.integer("person")])
```
bobintetley: This condition/code isn't necessary, see comment below on getting person from the chooser

src/static/js/regular_debits.js
```diff
@@ -0,0 +1,275 @@
...
+                $("#monthrow").hide();
+            } else if ($("#period").val() == 30) {
+                $("#weekdayrow").hide();
+                $("#dayofmonthrow").show();
+                $("#monthrow").hide();
+            } else {
+                $("#weekdayrow").hide();
+                $("#dayofmonthrow").show();
+                $("#monthrow").show();
+            }
```
bobintetley: while this is fine, you could probably use toggle to make this a bit easier to read:

```
let period = $("#period").val();
$("#weekdayrow").toggle( period == 7 );
$("#dayofmonthrow").toggle( period == 30 || period == 365 );
$("#monthrow").toggle( period == 365 );
```

src/static/js/regular_debits.js
adamdspencer77: $("#person").personchooser("get_selected").OWNERNAME didn't include the owner code. Went for $("#personrow .asm-embed-name").text() instead. Also noticed that the ownername column was left undefined when a new regular debit was added, fixed.

src/asm3/financial.py
adamdspencer77: Removed.

src/static/js/regular_debits.js
adamdspencer77: Done.

PR #1970, opened 2025-12-16
Medical treatment times: per‑dose scheduling and time‑aware Medical Book

binbashburns:
### Add Treatment times (per day) to the Add/Edit Medical Regimen dialog:
- Single Treatment → one time field.
- Multiple Treatments → one “Dose N” time per treatment per day.
- Defaults are evenly spaced between `DefaultShiftStart` and `DefaultShiftEnd`.

Screenshot 2025-12-16 at 08 29 31

### Store per‑dose times in `DateRequired` and update Medical Book:
- Date filters (“Due today”, “Due next week”, etc.) now include all times on the selected days.
- Rows become **overdue** once `DateRequired` (date + time) is in the past, not just when the date has passed.

Screenshot 2025-12-16 at 08 40 32

### Animals manual updated so Medical Book docs mention required/given times and configurable per‑day treatment times.
Screenshot 2025-12-16 at 08 58 01


src/static/js/medical.js
```diff
@@ -195,9 +253,11 @@ $(function() {
                     { field: "COSTPAIDDATE", display: _("Paid"), formatter: tableform.format_date,
                         hideif: function() { return !config.bool("ShowCostPaid"); }
                     },
-                    { field: "DATEREQUIRED", display: _("Required"), formatter: tableform.format_date, initialsort: true,
```
bobintetley, line 198: I would keep this as the single Required column, but make the formatter a custom function that outputs the date and time if it has one

src/static/js/medical.js
```diff
@@ -735,6 +806,40 @@ $(function() {
...
+        },
+
+        rebuild_treatment_times: function() {
+            const sm = $("#singlemulti").val();
+            const row = $("#treatmenttimesrow");
+            const table = row.closest("table");
+
+            table.find("tr.asm-tx-time-row").remove();
+
+            if (sm != medical.TREATMENT_MULTI && sm != medical.TREATMENT_SINGLE) {
```
bobintetley: I think the existing functionality should be retained as it is, just because we have many users who fear change and for many this will be a step too far. 

I would add a new constant for TREATMENT_MULTI_TIME and an entry in the frequency dropdown (Multiple Treatments with times) to trigger the time behaviour. This keeps it a bit neater and the time functionality and code for outputting the new dose fields etc can be kept behind a condition.

src/static/js/medical.js
```diff
@@ -735,6 +806,40 @@ $(function() {
...
+            row.show();
+
+            let html = "";
+            for (let i = 1; i <= perday; i++) {
+                const t = medical.treatment_time_for({ TREATMENTNUMBER: i, TOTALTREATMENTS: perday }) || "";
+                html += '<tr class="asm-formfield asm-tx-time-row">' +
+                        '<td><label>' + _("Dose {0}").replace("{0}", i) + '</label></td>' +
+                        '<td><input type="text" class="asm-textbox asm-timebox" ' +
+                        'name="treatmenttime' + i + '" value="' + t + '"/></td>' +
+                        '</tr>';
```
bobintetley: Use the tableform render functions instead of hand cranking the HTML, eg:

```
tableform.render_time({ name: "treatmenttime" + i, 
    value: t, 
    label: _("Dose {0}").replace("{0}", i), 
    rowclasses: "asm-tx-time-row });
```

src/static/js/medical.js
```diff
@@ -444,7 +504,16 @@ $(function() {
...
-                        await tableform.fields_post(dialog.fields, "mode=create", "medical");
+                        let perday = format.to_int($("#timingrule").val() || "1");
+                        let times = [];
+                        for (let i = 1; i <= perday; i++) {
+                            times.push($("input[name='treatmenttime" + i + "']").val());
+                        }
+                        let extra = "";
+                        if (times.length) {
+                            extra = "&treatmenttimes=" + encodeURIComponent(times.join(","));
+                        }
```
bobintetley, line 541: this code to read the time fields seems to be duplicated twice in this module and should be put in its own function for reuse

bobintetley (commented):
This is really impressive, thank you for your hard work!

src/static/js/medical.js
```diff
@@ -59,6 +60,33 @@ $(function() {
             return pad(rh) + ":" + pad(rm);
         },
 
+        format_required: function(row, v) {
+            if (!v) { return ""; }
+            const date = format.date(v);
+            const time = format.time(v, "%H:%M", true);
+            if (time) { return '<span style="white-space: nowrap;">' + date + " " + time + "</span>"; }
+            return date;
+        },
```
binbashburns, line 69: > I would keep this as the single Required column, but make the formatter a custom function that outputs the date and time if it has one

**Changes**: medical.js adds medical.format_required and uses it for the Required column.

src/static/js/medical.js
binbashburns: Thank you for the feedback! Addressed here: https://github.com/sheltermanager/asm3/pull/1970#discussion_r2637133089

src/static/js/medical.js
```diff
@@ -10,6 +10,7 @@ $(function() {
         TREATMENT_SINGLE: 0,
         TREATMENT_MULTI: 1,
         TREATMENT_CUSTOM: 2,
+        TREATMENT_MULTI_TIME: 3,
```
binbashburns, line 13: > I think the existing functionality should be retained as it is, just because we have many users who fear change and for many this will be a step too far.
> 
> I would add a new constant for TREATMENT_MULTI_TIME and an entry in the frequency dropdown (Multiple Treatments with times) to trigger the time behaviour. This keeps it a bit neater and the time functionality and code for outputting the new dose fields etc can be kept behind a condition.

**Changes**: New constant and dropdown option; time UI/behavior now only runs when this option is selected.

src/static/js/medical.js
binbashburns: Thank you for the feedback! Addressed here: https://github.com/sheltermanager/asm3/pull/1970#discussion_r2637134912

src/static/js/medical.js
```diff
@@ -832,11 +857,12 @@ $(function() {
...
-                        '<td><label>' + _("Dose {0}").replace("{0}", i) + '</label></td>' +
-                        '<td><input type="text" class="asm-textbox asm-timebox" ' +
-                        'name="treatmenttime' + i + '" value="' + t + '"/></td>' +
-                        '</tr>';
+                html += tableform.render_time({
+                    name: "treatmenttime" + i,
+                    value: t,
+                    label: _("Dose {0}").replace("{0}", i),
+                    rowclasses: "asm-tx-time-row"
+                });
```
binbashburns, line 865: > Use the tableform render functions instead of hand cranking the HTML, eg:
> 
> ```
> tableform.render_time({ name: "treatmenttime" + i, 
>     value: t, 
>     label: _("Dose {0}").replace("{0}", i), 
>     rowclasses: "asm-tx-time-row });
> ```

**Change**: `rebuild_treatment_times` now uses `tableform.render_time`.

src/static/js/medical.js
binbashburns: Addressed here: https://github.com/sheltermanager/asm3/pull/1970#discussion_r2637136397

src/static/js/medical.js
```diff
@@ -59,6 +60,33 @@ $(function() {
...
+        read_treatment_times: function() {
+            if (!medical.treatment_times_enabled()) { return []; }
+            const perday = medical.treatment_times_per_day();
+            if (!perday || perday <= 0) { return []; }
+            let times = [];
+            for (let i = 1; i <= perday; i++) {
+                times.push($("input[name='treatmenttime" + i + "']").val());
+            }
+            return times;
+        },
```
binbashburns, line 88: > this code to read the time fields seems to be duplicated twice in this module and should be put in its own function for reuse

**Change**: Centralized time field reading in `medical.read_treatment_times()` and reused it for create/edit.

src/static/js/medical.js
binbashburns: Thank you!
Addressed here: https://github.com/sheltermanager/asm3/pull/1970#discussion_r2637137923

src/static/js/medical.js
```diff
@@ -167,7 +192,14 @@ $(function() {
...
                 overdue: function(row) {
-                    return !row.DATEGIVEN && row.STATUS == 0 && format.date_js(row.DATEREQUIRED) < new Date();
+                    if (row.DATEGIVEN || row.STATUS != 0) { return false; }
+                    const req = row.DATEREQUIRED;
+                    if (!req) { return false; }
+                    const hasTime = !!format.time(req, "%H:%M", true);
+                    if (hasTime) {
+                        return format.date_js(req) < new Date();
+                    }
+                    return format.date_js(req, true) < common.today_no_time();
```
binbashburns, line 202: **Behavior fix (follow‑up from debugging after making the requested changes)**

**Issue:** Date‑only rows (no time) were being treated as overdue immediately.  

Image


**Fix:** `medical.js` now uses a date‑only comparison when no time is present.

```js
// src/static/js/medical.js
overdue: function(row) {
    if (row.DATEGIVEN || row.STATUS != 0) { return false; }
    const req = row.DATEREQUIRED;
    if (!req) { return false; }
    const hasTime = !!format.time(req, "%H:%M", true);
    if (hasTime) {
        return format.date_js(req) < new Date();
    }
    return format.date_js(req, true) < common.today_no_time();
},
```

**Screenshots after the fix (verification)**

Multiple Treatments (no times) — all rows are normal, Required shows date‑only.

Image

Frequency dropdown includes new “Multiple Treatments with times” option.

Image

Multiple Treatments with times — Required shows time and overdue highlights are time‑aware.

Image

PR #1990, opened 2026-01-16
fix: prevent deletion of default lookups

### LINKED ISSUE #1989: Prevent lookups designated as the default from being deleted
As the title, stop lookups that are set as the default under Settings->Options->Defaults from being deleted.

To implement, add a new element to the LOOKUPS table in lookups.py for the name of the config element that may refer to this as the default.

Then, the lookups.delete_lookup function can check to see if the lookup the user wants to delete is present as the value in the config element for the default for that lookup and raise an exception if it is.

jsssun:
Fixes #1989

- Added default config key metadata to LOOKUP_TABLES
- Prevent deletion of lookups that are currently configured as defaults


src/asm3/lookups.py
```diff
@@ -70,7 +70,7 @@
     "pickuplocation":       (_("Pickup Locations"), "LocationName", _("Location"), "LocationDescription", "add del ret", ("animal.PickupLocationID", "animalcontrol.PickupLocationID")),
     "reservationstatus":    (_("Reservation Statuses"), "StatusName", _("Status"), "StatusDescription", "add del ret", ("adoption.ReservationStatusID",)),
     "site":                 (_("Sites"), "SiteName", _("Site"), "", "add del", ("users.SiteID","internallocation.SiteID")),
-    "species":              (_("Species"), "SpeciesName", _("Species"), "SpeciesDescription", "add del ret pubspec", ("animal.SpeciesID", "onlineformfield.SpeciesID", "animallost.AnimalTypeID", "animalfound.AnimalTypeID")),
+    "species":              (_("Species"), "SpeciesName", _("Species"), "SpeciesDescription", "add del ret pubspec", ("animal.SpeciesID", "onlineformfield.SpeciesID", "animallost.AnimalTypeID", "animalfound.AnimalTypeID"), "AFDefaultSpecies"),
```
bobintetley, line 73: Thanks for submitting this patch :)

The default config key you've added to species also needs adding to all of the other lookups in this list where applicable. See configuration.py/DEFAULTS to locate all of the keys that start with the name "AFDefault" or "Default" to match the lookup

jsssun: Thanks for the guidance! I've added the appropriate default config keys

PR #1994, opened 2026-01-22
onlineform field: checkbox (align label)

adamdspencer77:
Many people ask about the checkbox layout on online forms. We follow normal UI guidelines by putting the label to the right of the checkbox, instead of the label in the left column and box on the right like the other fields. Allow this to be an option.

    Make a new form field type of CHECKBOX_AL that appears in the type dropdown as "Checkbox (align label)" and outputs the label and checkbox this way in get_onlineform_html

    Make sure the code in insert_onlineformincoming_from_form recognises this new type as well as regular CHECKBOX when adding flags

### DIFF
--- a/src/asm3/onlineform.py
+++ b/src/asm3/onlineform.py
@@ -220,6 +220,11 @@ def get_onlineform_html(dbo: Database, formid: int, completedocument: bool = Tru
             h.append('<td class="asm-onlineform-td asm-onlineform-raw" colspan="2">')
         elif f.FIELDTYPE == FIELDTYPE_CHECKBOX:
             h.append('<td class="asm-onlineform-td">%s</td><td class="asm-onlineform-td">' % requiredspan)
+        elif f.FIELDTYPE == FIELDTYPE_CHECKBOX_AL:
+            h.append('<td class="asm-onlineform-td">')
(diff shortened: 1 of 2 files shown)


src/asm3/onlineform.py
```diff
@@ -218,6 +220,11 @@ def get_onlineform_html(dbo: Database, formid: int, completedocument: bool = Tru
             h.append('<td class="asm-onlineform-td asm-onlineform-raw" colspan="2">')
         elif f.FIELDTYPE == FIELDTYPE_CHECKBOX:
             h.append('<td class="asm-onlineform-td">%s</td><td class="asm-onlineform-td">' % requiredspan)
+        elif f.FIELDTYPE == FIELDTYPE_CHECKBOX_AL:
+            h.append('<td class="asm-onlineform-td">')
+            h.append('<label for="%s">%s %s</label>' % ( fid, f.LABEL, requiredspan ))
+            h.append('</td>')
+            h.append('<td class="asm-onlineform-td">')
```
bobintetley, line 227: Unless I'm missing something, this code block doesn't seem to be necessary? CHECKBOX_AL has the same label behaviour as everything that isn't a checkbox or raw markup so this could be removed?

adamdspencer77: CHECKBOX_AL doesn't have a tooltip. That field is used for flags.

PR #2010, opened 2026-02-20
Move screens: multiple animals, workflow and document templates

adamdspencer77:
A user pointed out that the generate documentation screen isn't really that relevant to go to after processing an adoption with the checkout (since you already generated the doc and it got sent to the adopter for signing).

I think it's a bad workflow for these screens in general, because after hitting the button, you're given a full screen list of document templates, all of which open a new tab with the document in that you have to then save. This is clumsy.

This ticket is for 2 things:

    The move screens should have a multi-select dropdown where you choose one or more document templates you would like generating. The docs are immediately linked/saved to the media tabs of the animal/person.

The final screen in the workflow instead should show info about the records just created with links, and include links to all the documents that were just generated so that the user can edit and print them if they like.

Also, because these screens can update a fair few records, it would make sense for there to be some kind of confirmation dialog before proceeding after the button is clicked.

    Multiple animals support: With the change to documents, these screens should be able to support multiple animals fairly easily. This would solve a long-standing request from customers who want to be able to adopt multiple animals to the same person at the same time and generate all the paperwork.

It could also make adopting of bonded animals more straightforward since we could automatically resolve and pull in the bonded animals.

The final question is whether the multiple animal support is beneficial to all of the move workflows, or just adoption in particular.


src/asm3/movement.py
```diff
@@ -624,20 +624,6 @@ def insert_adoption_from_form(dbo: Database, username: str, post: PostedData, cr
...
-    if a.BONDEDANIMALID is not None and a.BONDEDANIMALID != 0 and a.BONDEDANIMAL1ARCHIVED == 0 and a.BONDEDANIMALID not in creating:
-        asm3.al.debug("Found bond to animal %d, creating adoption..." % a.BONDEDANIMALID, "movement.insert_adoption_from_form", dbo)
-        newdata = dict(post.data)
-        newdata["animal"] = str(a.BONDEDANIMALID)
-        insert_adoption_from_form(dbo, username, asm3.utils.PostedData(newdata, dbo.locale), creating, create_payments = False)
-    if a.BONDEDANIMAL2ID is not None and a.BONDEDANIMAL2ID != 0 and a.BONDEDANIMAL2ARCHIVED == 0 and a.BONDEDANIMAL2ID not in creating:
-        asm3.al.debug("Found bond to animal %d, creating adoption..." % a.BONDEDANIMAL2ID, "movement.insert_adoption_from_form", dbo)
-        newdata = dict(post.data)
-        newdata["animal"] = str(a.BONDEDANIMAL2ID)
-        insert_adoption_from_form(dbo, username, asm3.utils.PostedData(newdata, dbo.locale), creating, create_payments = False)
```
bobintetley, line 642: If we're keeping the old move_adopt, this functionality needs to stay here. Probably gated behind an optional boolean argument to this function adoptBonded: bool = True  that can be set to false by the caller from your move_adopt_multi endpoint.

src/asm3/movement.py
```diff
@@ -705,11 +691,11 @@ def insert_adoption_from_form(dbo: Database, username: str, post: PostedData, cr
     cost_create = post.boolean("costcreate")
     if cost_amount > 0 and cost_type != "" and cost_create:
         boc_dict = {
-            "animal"            : post["animal"],
+            "animalid"          : post["animal"],
             "type"              : cost_type,
             "costdate"          : post["movementdate"],
             "costpaid"          : post["movementdate"],
-            "cost"              : post["costamount"]
+            "cost"              : post["cost"]
```
bobintetley: is the old move_adopt still sending this as costamount and your new code as just "cost"? Seems like your new code should be sending it as costamount too

src/static/js/common_animalchoosermulti.js
```diff
@@ -350,5 +365,43 @@ $.fn.animalchoosermulti = asm_widget({
...
+                    });
+                }
+            });
+        } else {
+            $.each(newval.split(","), function(i, v) {
+                let rec = self.get_row(t, parseInt(v));
+                let disp = "<a class=\"asm-embed-name\" href=\"animal?id=" + rec.ID + "\">" + rec.CODE + " - " + rec.ANIMALNAME + "</a>";
+                o.display.append(disp);
+                o.display.append("<br />");
+            });
```
bobintetley: This feels like a lot of duplicate code to me? If there was a function in the widget responsible for fetching the rows then you could call it and this function only needs to worry about getting and setting the value then.

I would make that row getting function return a promise so that the value setting code can do .then(function()) or just await it. Search for $.Deferred in common_tableform.js for examples of how to write a function that returns a promise.

src/static/js/options.js
```diff
@@ -806,7 +803,8 @@ $(function() {
                         { id: "autoinsurance", post_field: "UseAutoInsurance", label: _("Use Automatic Insurance Numbers"), type: "check" }, 
                         { id: "insurancestart", post_field: "UseAutoAutoInsuranceStartInsurance", label: _("Start at"), type: "number" }, 
                         { id: "insuranceend", post_field: "AutoInsuranceEnd", label: _("End at"), type: "number" }, 
-                        { id: "insurancenext", post_field: "AutoInsuranceNext", label: _("Next"), type: "number" } 
+                        { id: "insurancenext", post_field: "AutoInsuranceNext", label: _("Next"), type: "number" },
+                        // { id: "insurablespecies", post_field: "InsurableSpecies", label: _("Insurable Species"), type: "selectmulti", options: html.list_to_options(controller.species, "ID", "SPECIESNAME") }
```
bobintetley: Probably meant to remove this?

src/main.py
```diff
@@ -5917,12 +6005,190 @@ class move_transfer(JSONEndpoint):
...
+        return {
+            "additional": asm3.additional.set_next_id(dbo, asm3.additional.get_additional_fields(dbo, 0, "movement", asm3.additional.MOVEMENT_ADOPTION)),
+            # "additional": asm3.additional.get_additional_fields(dbo, 0, "movement", asm3.additional.MOVEMENT_ADOPTION),
+            "donationtypes": asm3.lookups.get_donation_types(dbo),
+            "accounts": asm3.financial.get_accounts(dbo, onlybank=True),
+            "paymentmethods": asm3.lookups.get_payment_methods(dbo),
+            "templates": asm3.template.get_document_templates(dbo, "movement"),
+            "templatesemail": asm3.template.get_document_templates(dbo, "email"),
+            "taxrates": asm3.lookups.get_tax_rates(dbo)
+        }
```
bobintetley: Might be missing something, but I can't see that move_workflow will ever be hit by a GET request, so it shouldn't need this controller method and can be ASMEndpoint instead of JSONEndpoint

src/main.py
```diff
@@ -5917,12 +6005,190 @@ class move_transfer(JSONEndpoint):
...
+            raise asm3.utils.ASMValidationError("No template given for paperwork")
+        if paperwork and post.integer("sigemailtemplateid") == 0:
+            raise asm3.utils.ASMValidationError("No email template given for request signature email")
+        createdocuments = []
+        incnumberwidgets = []
+        movementdata = {}
+
+        ## Create post data for virtual incrementing number fields
+        for key in post.data.keys():
+            if key[:13] == "asm-incnumber":
```
bobintetley: key.startswith("asm-incnumber") would make more sense to me

src/main.py
```diff
@@ -5917,12 +6005,190 @@ class move_transfer(JSONEndpoint):
...
+                incnumberwidgets.append(key[13:])
+        for key in incnumberwidgets:
+            post.data[key] = post.data["asm-incnumber" + key]
+            del post.data["asm-incnumber" + key]
+
+        ## Loop through animals
+        animalcount = 0
+        for animalid in post["animals"].split(","):
```
bobintetley: use enumerate instead - ``` for animalcount, animalid in enumerate(post["animals"].split(",")) ```

src/main.py
```diff
@@ -5917,12 +6005,190 @@ class move_transfer(JSONEndpoint):
...
+            post["insurance"] = post["insurance" + animalid]
+            post["costamount"] = post["animalcost" + animalid]
+            post["cost"] = post["animalcost" + animalid]
+            movementid = 0
+
+            ## Reconstruct payment post data for this animal only
+            virtualpostkeys = []
+            postkeys = []
+            for key in post.data.keys():
+                postkeys.append(key)
```
bobintetley: why copy keys into a new list? you could just iterate post.data.keys() in the line below as far as I can tell? Or iterate a copy()

src/main.py
```diff
@@ -5917,12 +6005,190 @@ class move_transfer(JSONEndpoint):
...
+            for key in postkeys:
+                if key[:12] == "paymentinput":
```
bobintetley: key.startswith is a bit more readable

src/static/js/common_widgets.js
```diff
@@ -198,6 +198,9 @@ $.fn.toPOST = function(includeblanks = false) {
         else if (t.hasClass("asm-selectcolour")) {
             post.push(pname + "=" + encodeURIComponent(t.selectcolour("value")));
         }
+        else if (t.hasClass("asm-incnumber") && markincnumbers) {
+            post.push("asm-incnumber" + pname + "=" + encodeURIComponent(parseInt(t.attr("value"))));
+        }
```
bobintetley: Not keen on this hack being in here. If you want to manipulate the name of the post variable, it can be done in your move_workflow.js code rather than here!

src/static/js/common_tableform.js
```diff
@@ -1713,7 +1704,8 @@ const tableform = {
         return tableform._render_formfield(v, v.markup);
     },
 
-    render_number: function(v) {
+    render_number: function(v, inc=false) {
+        if (inc) { v.classes += " asm-incnumber"; }
```
bobintetley: this inc parameter and line isn't necessary - just set v.classes in your caller in header_additional.js

src/static/js/header_additional.js
```diff
@@ -415,7 +415,7 @@ additional = {
         else if (f.FIELDTYPE == additional.TIME) { return tableform.render_time(v); }
         else if (f.FIELDTYPE == additional.NOTES) { v.classes += " asm-textareafixed"; return tableform.render_textarea(v); }
         else if (f.FIELDTYPE == additional.NUMBER) { return tableform.render_number(v); }
-        else if (f.FIELDTYPE == additional.NUMBER_INCREMENTED) { v.readonly = true; return tableform.render_number(v); }
+        else if (f.FIELDTYPE == additional.NUMBER_INCREMENTED) { v.readonly = true; return tableform.render_number(v, true); }
```
bobintetley: set v.classes to asm-incnumber here just like textareafixed does above and no need to pass the true argument to render_number (no need for the arg in the function)

src/main.py
```diff
@@ -5917,12 +6005,190 @@ class move_transfer(JSONEndpoint):
...
+        incnumberwidgets = []
+        movementdata = {}
+
+        ## Create post data for virtual incrementing number fields
+        for key in post.data.keys():
+            if key[:13] == "asm-incnumber":
+                incnumberwidgets.append(key[13:])
+        for key in incnumberwidgets:
+            post.data[key] = post.data["asm-incnumber" + key]
+            del post.data["asm-incnumber" + key]
```
bobintetley: I think both these loops could be rewritten as:

```
for key in post.data.keys().copy():
    if key.startswith("asm-incnumber"):
        post.data[key.replace("asm-incnumber", "")] = post.data[key]
        del post.data[key]
```

src/main.py
```diff
@@ -5917,12 +6005,190 @@ class move_transfer(JSONEndpoint):
...
+            postkeys = []
+            for key in post.data.keys():
+                postkeys.append(key)
+            for key in postkeys:
+                if key[:12] == "paymentinput":
+                    aid = key[12:].split("||")[0]
+                    if aid == animalid or (aid == "0" and animalcount == 1):
+                        virtualfield = key.split("||")[1]
+                        virtualpostkeys.append(virtualfield)
+                        post[virtualfield] = post[key]
```
bobintetley, line 6111: I'm not sure what this virtualfield bit is doing, if you needed them to be named a certain way, could it not have been done in the payment widget in the client?

src/main.py
```diff
@@ -5917,12 +6005,190 @@ class move_transfer(JSONEndpoint):
...
+            elif post["movementtype"] == "reclaim":
+                movementid = asm3.movement.insert_reclaim_from_form(dbo, o.user, post)
+            movementdata[animalid] = movementid
+
+            ## Remove virtual payment posts
+            for vf in virtualpostkeys.copy():
+                try:
+                    post.data.pop(vf)
+                except:
+                    pass
```
bobintetley: except pass is a big no-no! I'll be honest, I'm not sure what this virtualpostkeys thing is about or why they would need to be removed? Most places will just ignore post values that they aren't interested in

src/asm3/movement.py
adamdspencer77: Done.

src/asm3/movement.py
adamdspencer77: Sorted.

src/static/js/common_animalchoosermulti.js
adamdspencer77: Removed code for retrieving rows after discussion. Left comment pointing out potential race condition for future us :)

src/static/js/common_tableform.js
adamdspencer77: Done.

src/static/js/common_widgets.js

src/static/js/header_additional.js
adamdspencer77: Done.

src/static/js/options.js
adamdspencer77: Yep. Done :)

src/main.py
adamdspencer77: Yeah, originally this started out as a replacement for move_adopt, changed.

src/main.py

src/main.py
adamdspencer77: Ended up after some faffing, going for:

```
for key in list(post.data.keys()).copy():
            if key.startswith("asm-incnumber"):
                incnumberwidgets.append(key[13:])
                post.data[key[13:]] = post.data[key]
                del post.data[key]
```

incnumberwidgets is used later on.

src/main.py

src/main.py
adamdspencer77: I'd hit an issue with not being able to use .copy() on post.data.keys(). Went for list(post.data.keys()) instead, seems to work fine.

```
virtualpostkeys = []
            for key in list(post.data.keys()):
                if key[:12] == "paymentinput":
                    aid = key[12:].split("||")[0]
                    if aid == animalid or (aid == "0" and animalcount == 1):
                        virtualfield = key.split("||")[1]
                        virtualpostkeys.append(virtualfield)
                        post[virtualfield] = post[key]
```

src/main.py

src/main.py
adamdspencer77: Discussed and left be.

src/main.py
adamdspencer77: Sorted.

PR #2013, opened 2026-02-24
maint_find_replace: Screen to fix typos in manufacturer, city and state

adamdspencer77:
There are a few places where we use autocomplete fields and the source of their data is previous values entered into that field.

What invariably happens is that users enter garbage full of typos instead of picking the correct value from the autocomplete when they see it (or typing it correctly in the first place).

The worst offending fields for this are vaccination manufacturer, person city and person state.

Create a new hidden screen (ie. no menu entry) called maint_find_replace. This screen should have a row for each of these fields.:

                     FIND          -> REPLACE WITH
vacc manufacturer [find]           ->[          ]  [go]
person city [find]                ->[          ]  [go]
person state [find]                ->[          ]  [go]

The find and replace columns show a select/dropdown that contains the current options for that field. Replace has the same list of options, allowing you to swap one typo value for another.

The go button does a post to the backend to do the actual replace. A bit of js after the post should remove the selected option in the find dropdown from the list in both find and replace so that a page transition isn't needed to refresh the dropdowns.

### DIFF
--- a/src/asm3/medical.py
+++ b/src/asm3/medical.py
@@ -634,12 +634,9 @@ def get_vaccinations_expiring_two_dates(dbo: Database, start: Database, end: Dat
         "AND a.DeceasedDate Is Null %s %s " \
         "ORDER BY av.DateExpires, a.AnimalName" % (shelterfilter, locationfilter), (start, end))
 
-def get_vacc_manufacturers(dbo: Database, includeblank=False) -> List[str]:
+def get_vacc_manufacturers(dbo: Database) -> List[str]:
     
-    if includeblank:
-        rows = dbo.query("SELECT DISTINCT Manufacturer FROM animalvaccination WHERE Manufacturer Is Not Null ORDER BY Manufacturer")
-    else:
-        rows = dbo.query("SELECT DISTINCT Manufacturer FROM animalvaccination WHERE Manufacturer Is Not Null AND Manufacturer <> '' ORDER BY Manufacturer")
+    rows = dbo.query("SELECT DISTINCT Manufacturer FROM animalvaccination WHERE Manufacturer Is Not Null AND Manufacturer <> '' ORDER BY Manufacturer")
     mf = []
     for r in rows:
         mf.append(r.MANUFACTURER)
@@ -860,13 +857,13 @@ def complete_test(dbo: Database, username: str, testid: int, newdate: datetime, 
     dbo.update("animaltest", testid, v, username)
     update_animal_tests(dbo, username, testid)
 
-def replace_manufacturers(dbo: Database, username: str, post: PostedData) -> None:
+def replace_manufacturers(dbo: Database, username: str, find: str, replace: str) -> None:
     """
-    Replaces the manufacturer in all vaccination records from posted form data
+    Replaces the manufacturer in all vaccination records
     """
     
-    return dbo.update("animalvaccination", "Manufacturer = '%s'" % post["manufacturerfind"], {
-        "Manufacturer": post["manufacturerreplace"]
+    return dbo.update("animalvaccination", "Manufacturer = %s" % dbo.sql_value(find), {
+        "Manufacturer": replace
     }, username)
 
 def reschedule_test(dbo: Database, username: str, testid: int, newdate: datetime, comments: str) -> None:
--- a/src/asm3/person.py
+++ b/src/asm3/person.py
@@ -254,11 +254,14 @@ def get_staff_volunteers(dbo: Database, siteid: int = 0) -> Results:
     if siteid is not None and siteid != 0: sitefilter = "AND o.SiteID = %s" % siteid
     return dbo.query(get_person_query(dbo) + " WHERE o.IsStaff = 1 OR o.IsVolunteer = 1 %s ORDER BY o.IsStaff DESC, o.OwnerSurname, o.OwnerForeNames" % sitefilter)
 
-def get_towns(dbo: Database) -> List[str]:
+def get_towns(dbo: Database, excludeblanks: bool = False) -> List[str]:
     """
     Returns a list of all towns
     """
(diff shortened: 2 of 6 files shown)


src/asm3/medical.py
```diff
@@ -634,8 +634,12 @@ def get_vaccinations_expiring_two_dates(dbo: Database, start: Database, end: Dat
         "AND a.DeceasedDate Is Null %s %s " \
         "ORDER BY av.DateExpires, a.AnimalName" % (shelterfilter, locationfilter), (start, end))
 
-def get_vacc_manufacturers(dbo: Database) -> List[str]:
-    rows = dbo.query("SELECT DISTINCT Manufacturer FROM animalvaccination WHERE Manufacturer Is Not Null AND Manufacturer <> '' ORDER BY Manufacturer")
+def get_vacc_manufacturers(dbo: Database, includeblank=False) -> List[str]:
```
bobintetley: Why would we ever want to include blanks in the list?

src/asm3/medical.py
```diff
@@ -856,6 +860,15 @@ def complete_test(dbo: Database, username: str, testid: int, newdate: datetime,
     dbo.update("animaltest", testid, v, username)
     update_animal_tests(dbo, username, testid)
 
+def replace_manufacturers(dbo: Database, username: str, post: PostedData) -> None:
+    """
+    Replaces the manufacturer in all vaccination records from posted form data
+    """
+    
+    return dbo.update("animalvaccination", "Manufacturer = '%s'" % post["manufacturerfind"], {
```
bobintetley: This is vulnerable to a SQL injection attack as soon as a manufacturer has a ' in it. 

Make your clause ``` "Manufacturer = %s" % dbo.sql_value(find) ``` to have it escaped correctly.

src/asm3/medical.py
```diff
@@ -856,6 +860,15 @@ def complete_test(dbo: Database, username: str, testid: int, newdate: datetime,
     dbo.update("animaltest", testid, v, username)
     update_animal_tests(dbo, username, testid)
 
+def replace_manufacturers(dbo: Database, username: str, post: PostedData) -> None:
```
bobintetley: Only pass posted data around if it's really necessary - I don't think it is for a function that only needs 2 parameters! have find and replace arguments instead.

Doing this makes it easier to write clean unit tests as well.

src/asm3/person.py
```diff
@@ -2193,6 +2193,24 @@ def remove_people_only_cancelled_reserve(dbo: Database, years: int = None, usern
...
+        "OwnerTown": post["cityreplace"]
+    }, username)
+
+def replace_states(dbo: Database, username: str, post: PostedData) -> None:
+    """
+    Replaces the county/state in all person records from posted form data
+    """
+    
+    return dbo.update("owner", "OwnerCounty = '%s'" % post["statefind"], {
+        "OwnerCounty": post["statereplace"]
```
bobintetley: Same deal as above, don't use post and escape your SQL values.

src/static/js/maint_find_replace.js
```diff
@@ -0,0 +1,140 @@
...
+    const maint_find_replace = {
+
+        render: function() {
+            let s = [
+                html.content_header(_("Find/replace")),
+                '<table id="findreplacetable">',
+                '<tbody>',
+                '<tr><td>' + _("Vaccine manufacturer") + '</td><td>',
+                tableform.fields_render([
+                    { post_field: "manufacturerfind", type: "select", justwidget: true,
```
bobintetley: This could have been a single fields_render call where each element had a label property instead of multiple fields_render calls stuck together with bits of HTML. Is there a reason you did it this way?

src/static/js/maint_find_replace.js
```diff
@@ -0,0 +1,140 @@
...
+                let formdata = "mode=replacemanufacturers&" + $("#manufacturerfind, #manufacturerreplace").toPOST();
+                let result = await common.ajax_post("maint_find_replace", formdata);
+                $("#manufacturerfind option:selected").remove();
+                let termpresent = false;
+                $.each($("#manufacturerfind option"), function(i, v) {
+                    if ( $(v).val() == $("#manufacturerreplace").val() ) {
+                        termpresent = true;
+                        return;
+                    }
+                });
```
bobintetley: Using a loop like this seems pretty clumsy when it's behaviour we've got to do 3 times.

I would modify the fn.select widget in common_widgets.js, add a new method to it for hasOption that returns true if the value you give it is present in the options list.

This loop and the ones below can then be replaced with something like ``` let termpresent = $("#manufacturerfind").select("hasOption", $("#manufacturerreplace").val()); ```

src/asm3/medical.py
adamdspencer77: I can think of some unlikely edge cases, but mostly I added this for consistency, the town and county queries included blanks and it seemed odd that manufacturer didn't match :)

src/asm3/medical.py
adamdspencer77: Changed.

src/asm3/medical.py
adamdspencer77: Changed.

src/asm3/person.py
adamdspencer77: Sorted.

src/static/js/maint_find_replace.js
adamdspencer77: Done.

src/static/js/maint_find_replace.js
adamdspencer77: Restructured.

src/asm3/medical.py
adamdspencer77: Changed so that manufacturer, towns and counties exclude blanks.

PR #2020, opened 2026-03-03
media: Download zip

adamdspencer77:
Have a new toolbar button on the media tab for "Download Zip". It should be greyed out until there is a selection.

It should then post the list of IDs to the backend and have the response be a zip file containing all the files. See wordprocessor/openoffice integration for an example of how to use the zip library.

### DIFF
--- a/src/asm3/media.py
+++ b/src/asm3/media.py
@@ -1,6 +1,7 @@
 
 import asm3.al
 import asm3.animal
+import asm3.asynctask
 import asm3.audit
 import asm3.cachedisk
 import asm3.configuration
@@ -9,7 +10,7 @@ import asm3.log
 import asm3.utils
 from asm3.i18n import _
 from asm3.sitedefs import RESIZE_IMAGES_DURING_ATTACH, RESIZE_IMAGES_SPEC, SCALE_PDF_DURING_ATTACH, SCALE_PDF_CMD, SERVICE_URL, VIDEO_THUMBNAIL_CMD, WATERMARK_FONT_BASEDIRECTORY
-from asm3.typehints import Database, Dict, PostedData, ResultRow, Results, Tuple, List
+from asm3.typehints import Database, Dict, List, PostedData, ResultRow, Results, Tuple
 
 from datetime import datetime
 import os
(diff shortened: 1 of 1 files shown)


src/asm3/media.py
```diff
@@ -1593,4 +1594,22 @@ def watermark_with_transparency(dbo: Database, imagedata: bytes, animalname: str
         asm3.al.error("failed watermarking image: %s" % str(err), "media.watermark_with_transparency")
         return imagedata
 
+def zip_media_files_by_id(dbo: Database, mediaids: List[int]):
+    ## Create the zip file
+    zo = asm3.utils.bytesio()
+    zfo = zipfile.ZipFile(zo, 'w')
+    for mid in mediaids:
```
bobintetley, line 1603: Should be calling asynctask.set_progress_max with len(mediaids) before starting this loop and asynctask.increment_progress_value during this loop so the user sees a progress meter of files being zipped up.

src/asm3/media.py
```diff
@@ -165,7 +166,7 @@ def get_media_export(dbo: Database) -> Results:
             m.DBFSNAME = "%s%s" % (m.DBFSID, ext)
     return rows
 
-def get_media_file_data(dbo: Database, mid: int) -> Tuple[datetime, str, bytes]:
+def get_media_file_data(dbo: Database, mid: int) -> Tuple[datetime, str, str, bytes]:
```
bobintetley, line 170: well spotted :)

src/asm3/media.py
```diff
@@ -2,13 +2,14 @@
...
 import asm3.audit
+import asm3.cachedisk
 import asm3.configuration
 import asm3.dbfs
 import asm3.log
 import asm3.utils
 from asm3.i18n import _
 from asm3.sitedefs import RESIZE_IMAGES_DURING_ATTACH, RESIZE_IMAGES_SPEC, SCALE_PDF_DURING_ATTACH, SCALE_PDF_CMD, SERVICE_URL, VIDEO_THUMBNAIL_CMD, WATERMARK_FONT_BASEDIRECTORY
-from asm3.typehints import Database, Dict, PostedData, ResultRow, Results, Tuple
+from asm3.typehints import Database, Dict, PostedData, ResultRow, Results, Tuple, List
```
bobintetley: always in alphabetical order if possible

src/asm3/media.py
adamdspencer77: :)

src/asm3/media.py
adamdspencer77: Done.

src/asm3/media.py
adamdspencer77: Done but runs too fast for me to confirm that the bar is moving when tested :)

PR #2022, opened 2026-03-10
Petco Love Lost integration

adamdspencer77:


src/asm3/publishers/petcolovelost.py
```diff
@@ -0,0 +1,286 @@
...
+        self.setLastError("")
+        self.setStartPublishing()
+        
+        auth = getAuthDetails(self.dbo)
+
+        cheaders = {
+            'x-api-key': auth["apikey"],
+            'Authorization': 'Bearer ' + auth["accesstoken"]
+        }
+        c = asm3.utils.get_url(f'{auth["url"]}/v2/animals/?limit=25&offset=0', cheaders)
```
bobintetley: limit=25 seems sus? This doesn't seem to be a loop so we're not chunking this data, are we just stopping at 25 animals?

src/asm3/publishers/petcolovelost.py
```diff
@@ -0,0 +1,286 @@
...
+        cjson = asm3.utils.json_parse(c["response"])
+        publishedanimalids = []
+        for pa in cjson["pets"]:
+            publishedanimalids.append((pa["metadata"]["system_animal_id"], pa["id"]))
```
bobintetley: Is the purpose of this mechanism to figure out which animals we've sent previously?

If so, it might be simpler to just use our animalpublished table, which we update when you call markAnimalsPublished().

the getAnimalDataQuery can then just use a subquery in its where clause to check the animalpublished table for this animal and petco and ignore the ones that have an entry, or an entry with a date older than x, etc.

src/asm3/publishers/petcolovelost.py
```diff
@@ -0,0 +1,286 @@
...
+        removalcount = 0
+        for p in publishedanimalids:
+            pvalid = False
+            for an in animals:
+                if p[0] == an["ID"]:
+                    pvalid = True
+            if not pvalid:
+                removeAnimal(auth, p[1])
+                removalcount = removalcount + 1
+        self.log("Removed %s obsolete animals" % ( str(removalcount)))
```
bobintetley, line 261: If we use our animalpublished table above, we can use a simple list comprehension to identify animals that we sent previously who weren't in the list this time and remove them.

petrescue and savourlife both do this.

src/main.py
```diff
@@ -7531,8 +7086,19 @@ def controller(self, o):
...
+    def post_pcllpublished(self, o):
+        auth = asm3.publishers.petcolovelost.getAuthDetails(o.dbo)
+        return asm3.publishers.petcolovelost.getPublishedAnimals(auth)
+
     def post_poll(self, o):
         return "%s|%d|%s" % (asm3.asynctask.get_task_name(o.dbo), asm3.asynctask.get_progress_percent(o.dbo), asm3.asynctask.get_last_error(o.dbo))
+    
+    def post_pcllpurge(self, o):
+        auth = asm3.publishers.petcolovelost.getAuthDetails(o.dbo)
+        return asm3.publishers.petcolovelost.purge(auth)
```
bobintetley: these pcl endpoints do not belong under publish. Should be at least publish_options since that's where they're called from. purge and published should be removed once it's working

src/static/js/publish_options.js
```diff
@@ -450,6 +461,69 @@ $(function() {
         },
 
         bind: function() {
+            $("#button-pcllgetshelterid")
+                .button({ icons: { primary: "ui-icon-arrowrefresh-1-s" }, text: false })
+                .click(async function() {
+                    validate.dirty(false);
+                    $(".pcllbutton").hide();
+                    $("#pclldata").html('<p><img src="/static/images/wait/rolling_3a87cd.svg" style="height: 15px;vertical-align: middle;"/> ' + _("Waiting for Petco Love Lost..") + '</p>');
```
bobintetley: these debug strings don't need to be translatable. No need to wrap them with _("")

src/static/js/publish_options.js
```diff
@@ -450,6 +461,69 @@ $(function() {
...
+                    $("#button-pcllpublished").show();
+                });
+            
+            $("#button-pcllpurge")
+                .button()
+                .click(async function() {
+                    $(".pcllbutton").hide();
+                    $("#pclldata").html('<p><img src="/static/images/wait/rolling_3a87cd.svg" style="height: 15px;vertical-align: middle;"/> ' + _("Waiting for Petco Love Lost..") + '</p>');
+                    let formdata = "mode=pcllpurge";
+                    await common.ajax_post("publish", formdata);
```
bobintetley: should be publish_options rather than publish

src/asm3/publishers/petcolovelost.py
adamdspencer77: Reworked to use animalpublished table to establish which animals are already published rather than using petco data.

src/asm3/publishers/petcolovelost.py
adamdspencer77: Switched to using animalpublished table. Went with updating rather than ignoring animals that are already published using asm3.animal.get_extra_id.

src/asm3/publishers/petcolovelost.py
adamdspencer77: Done.

src/main.py
adamdspencer77: Moved.

src/static/js/publish_options.js
adamdspencer77: Tidied.

src/static/js/publish_options.js
adamdspencer77: Changed.

src/asm3/publishers/petcolovelost.py
```diff
@@ -0,0 +1,309 @@
...
+        "username": username,
+        "password": password,
+        "apikey": apikey,
+        "url": url,
+        "shelterid": shelterid,
+        "accesstoken": responsejson["accessToken"]
+    }
+
+def getActualPublishedAnimals(auth: Dict) -> Dict:
+    ## This is hidden from public, just useful for testing/debugging
```
bobintetley: Use pydoc rather than comments:

```
""" This is not used in normal publishing, but useful for testing/debugging to see what has been received by Petco """
```

src/asm3/publishers/petcolovelost.py
```diff
@@ -0,0 +1,309 @@
...
+    headers = {
+        'x-api-key': auth["apikey"],
+        'Authorization': 'Bearer ' + auth["accesstoken"]
+    }
+    response = asm3.utils.get_url(f'{auth["url"]}/v2/animals/?limit=100&offset=0', headers)
+    return response["response"]
+
+def getRecordedPublishedAnimals(dbo: Database) -> Results:
+    return dbo.query(
```
bobintetley, line 97: add pydoc comment, what is this function for? I presume retrieving the list of animals that we we've sent according to animalpublished

src/asm3/publishers/petcolovelost.py
```diff
@@ -0,0 +1,309 @@
...
+        getAnimalDataQuery() +
+        f"INNER JOIN animalpublished ap ON ap.AnimalID = a.ID WHERE ap.PublishedTo = '{IDTYPE_PETCOLOVELOST}'"
+    )
+
+def purgeActualPublished(auth: Dict):
+    animals = getActualPublishedAnimals(auth)
```
bobintetley, line 105: needs pydoc to explain what the function does and that it's not called as part of normal publishing

src/asm3/publishers/petcolovelost.py
```diff
@@ -0,0 +1,309 @@
...
+    """
+    def __init__(self, dbo: Database, publishCriteria: PublishCriteria) -> None:
+        AbstractPublisher.__init__(self, dbo, publishCriteria)
+        self.initLog("petcolovelost", "Petcolovelost Publisher")
+        self.dbo = dbo
+    
+    def getAnimalData(self) -> str:
+        return self.dbo.query(
+            getAnimalDataQuery() +
+            "WHERE a.Archived = 0 AND ( s.SpeciesName = 'Dog' OR s.SpeciesName = 'Cat' ) AND et.ID = 2 AND a.CrueltyCase = 0" ## Include only archived, non case stray, cats and dogs
```
bobintetley: Always use SpeciesID=1 (Dog) and SpeciesID=2 (Cat) - people quite often rename their species to Canine and Feline for example, so this would break.

Also, the lint tools get really angry when you start comments with ## - just one will do, and always run "make compilepy" before committing to run the linter

src/asm3/publishers/petcolovelost.py
```diff
@@ -0,0 +1,309 @@
...
+
+                # Do we already have a PetcoLoveLost ID for this animal?
+                # This function returns empty string for no animalid
+                pcllid = asm3.animal.get_extra_id(self.dbo, an, IDTYPE_PETCOLOVELOST)
+
+                if pcllid:
+                    # Update existing post
+                    self.log("Updating: %s: %s (%d of %d)" % ( an["SHELTERCODE"], an["ANIMALNAME"], anCount, len(animals)))
+                    del payload["shelterId"]
+                    del payload["species"]
```
bobintetley, line 194: not necessarily a problem, but why remove these 2 elements before sending an update?

src/asm3/publishers/petcolovelost.py
```diff
@@ -0,0 +1,309 @@
...
+    )
+
+def purgeActualPublished(auth: Dict):
+    animals = getActualPublishedAnimals(auth)
+    animalsjson = asm3.utils.json_parse(animals)
+    for animal in animalsjson["pets"]:
+        removeAnimal(auth, animal["id"])
+
+def purgeImages(auth: Dict, pcllaid: str):
+    headers = {
```
bobintetley, line 112: document this function. Also, it looks like this gets called every time a listing is updated, seems like that might be unnecessary given that this publisher could run every couple of hours.

src/asm3/publishers/petcolovelost.py
```diff
@@ -0,0 +1,309 @@
...
+                    # Create new post
+                    self.log("Creating: %s: %s (%d of %d)" % ( an["SHELTERCODE"], an["ANIMALNAME"], anCount, len(animals)))
+                    r = asm3.utils.post_json(f"{auth["url"]}/v2/animals", asm3.utils.json(payload), headers)
+
+                if r["status"] in (200, 201):
+                    responsejson = asm3.utils.json_parse(r["response"])
+                    if not pcllid:
+                        pcllid = responsejson["id"]
+                    asm3.animal.set_extra_id(self.dbo, "pub::petcolovelost", an, IDTYPE_PETCOLOVELOST, pcllid)
+                    purgeImages(auth, pcllid)
```
bobintetley, line 206: just so I'm clear, we remove all the photos every time, even if they haven't changed?

There is a value in the normal animal query that tells you how many new photos have been added in the last X days that you might want to copy.

src/asm3/publishers/petcolovelost.py
```diff
@@ -0,0 +1,309 @@
...
+            removalcount += 1
+        self.log("Removed %s obsolete animals" % ( str(removalcount)))
+
+        self.saveLog()
+        self.setPublisherComplete()
+    
+    def processAnimal(self, an: ResultRow, shelterid: str, weightunit: str) -> Dict:
+        """ Generate a dictionary of data to post from an animal record """
+        
+        ## Weight must be an integer number of ounces
```
bobintetley, line 252: Use a single # for comments :)

src/asm3/publishers/petcolovelost.py
adamdspencer77: Changed.

src/asm3/publishers/petcolovelost.py
adamdspencer77: Done.

src/asm3/publishers/petcolovelost.py
adamdspencer77: Done.

src/asm3/publishers/petcolovelost.py
adamdspencer77: Done.

src/asm3/publishers/petcolovelost.py
adamdspencer77: It's a Petco thing. Animal creation required shelterID and species but animal updating doesn't need them and barfs if you include them :)

src/asm3/publishers/petcolovelost.py
adamdspencer77: Done and reworked so images only get updated if animal record has changed or a media record has been recently created/updated.

src/asm3/publishers/petcolovelost.py
adamdspencer77: Reworked so images only get updated if animal record has changed or a media record has been recently created/updated.

src/asm3/publishers/petcolovelost.py
adamdspencer77: Done.

src/asm3/publishers/petcolovelost.py
```diff
@@ -0,0 +1,333 @@
...
+                # This function returns empty string for no animalid
+                pcllid = asm3.animal.get_extra_id(self.dbo, an, IDTYPE_PETCOLOVELOST)
+                if lastpublished and an.LASTCHANGEDDATE < lastpublished:
+                    self.log("Skipping: %s: %s (%d of %d) as not changed since last publish" % ( an["SHELTERCODE"], an["ANIMALNAME"], anCount, len(animals)))
+                    anchanged = False
+                    processed_animalids.append(an.ID)
+                elif pcllid:
+                    # Update existing post
+                    self.log("Updating: %s: %s (%d of %d)" % ( an["SHELTERCODE"], an["ANIMALNAME"], anCount, len(animals)))
+                    del payload["shelterId"]
```
bobintetley, line 210: I'd add a comment to say why these elements need to be removed

src/asm3/publishers/petcolovelost.py
```diff
@@ -0,0 +1,333 @@
...
+                        else:
+                            self.logError("Error: HTTP %d, headers: %s, response: %s" % (r["status"], r["headers"], r["response"]))
+                    purgeImages(auth, pcllid)
+                    photourls = self.getPhotoUrls(an["ID"])
+                    if len(photourls):
+                        imagepayload = {"photos": []}
+                        for photourl in photourls:
+                            # Tweak to allow photos through to Petco Love Lost on dev server
+                            if PETCO_LOVELOST_BASE_URL == "https://api-dev.petcolove.org":
+                                imagepayload["photos"].append({"url": photourl.replace("sheltermanager.com/service", "sheltermanager.com/dev/service")})
```
bobintetley: Not keen on the domain being hardcoded here, this whole if block should not be necessary as getPhotoUrls should already be returning the URL pointing to the correct service endpoint.  If it isn't, we should figure out why that is, and either way this isn't necessary for production code.

src/static/js/publish_options.js
```diff
@@ -449,7 +460,90 @@ $(function() {
...
+                    // $(".pcllbutton").hide();
+                    // $("#pclldata").html('<p><img src="/static/images/wait/rolling_3a87cd.svg" style="height: 15px;vertical-align: middle;"/> Waiting for Petco Love Lost..</p>');
+                    header.show_loading(_("Waiting for Petco Love Lost"));
+                    let formdata = "mode=pcllpurge";
+                    await common.ajax_post("publish_options", formdata);
+                    $("#pclldata").html("");
+                    header.show_info("Petco Love Lost purged");
+                    // $(".pcllbutton").show();
+                    header.hide_loading();
+                });
```
bobintetley: I would put these click events in their own function just to keep petco related stuff away from normal bind stuff and call it from bind. Eg: bind_petcolovelost

src/asm3/publishers/petcolovelost.py
adamdspencer77: Done.

src/static/js/publish_options.js
adamdspencer77: Done.

src/asm3/publishers/petcolovelost.py
adamdspencer77: Sorted.

adamdspencer77:
Something odd is happening with the image processing. Petco appear to accept image URLS on the test server in this format:
`https://eur05e.sheltermanager.com/dev/service?account=robin&method=media_image&mediaid=785&ts=1773075883.0`
But not this format:
`https://eur05e.sheltermanager.com/service?account=robin&method=media_image&mediaid=785&ts=1773075883.0`

src/asm3/sitedefs.py
```diff
@@ -284,6 +285,9 @@ def get_dict(k: str, dv: Dict = {}) -> Dict:
 ANIBASE_API_USER = get_string("anibase_api_user", "")
 ANIBASE_API_KEY = get_string("anibase_api_key", "")
 AVID_US_BASE_URL = get_string("avid_us_base_url", "")
+PETCO_LOVELOST_BASE_URL = get_string("petco_lovelost_base_url", "")
+PETCO_LOVELOST_API_KEY = get_string("petco_lovelost_api_key", "")
+PETCO_LOVELOST_EXTRAS = get_boolean("petco_lovelost_extras", False)
```
bobintetley: put these in alphabetical order with the other publishers!

src/main.py
```diff
@@ -7588,11 +7144,28 @@ def controller(self, o):
             "hassavourlife": SAVOURLIFE_URL != "",
             "entryreasons": asm3.lookups.get_entryreasons(dbo),
             "logtypes": asm3.lookups.get_log_types(dbo),
+            "pcllextras": PETCO_LOVELOST_EXTRAS,
```
bobintetley: call this haspetcolovelostdebug instead and put it next to the haspetcolovelost variable so they're together

src/main.py
```diff
@@ -70,9 +68,10 @@
     AKC_REUNITE_BASE_URL, AVID_US_BASE_URL, BUDDYID_BASE_URL, FINDPET_BASE_URL, HOMEAGAIN_BASE_URL, \
     LARGE_FILES_CHUNKED, LOCALE, JQUERY_UI_CSS, \
     LEAFLET_CSS, LEAFLET_JS, MULTIPLE_DATABASES, \
-    ADMIN_EMAIL, EMAIL_ERRORS, MADDIES_FUND_TOKEN_URL, HTMLFTP_PUBLISHER_ENABLED, HTML_TO_PDF, \
+    ADMIN_EMAIL, EMAIL_ERRORS, MADDIES_FUND_TOKEN_URL, HTMLFTP_PUBLISHER_ENABLED, \
     MANUAL_HTML_URL, MANUAL_PDF_URL, MANUAL_FAQ_URL, MANUAL_VIDEO_URL, MAP_LINK, MAP_PROVIDER, \
-    MAP_PROVIDER_KEY, MAX_DOCUMENT_TEMPLATE_SIZE, OSM_MAP_TILES, FOUNDANIMALS_FTP_USER, PETCADEMY_FTP_HOST, \
+    MAP_PROVIDER_KEY, MAX_DOCUMENT_TEMPLATE_SIZE, OSM_MAP_TILES, FOUNDANIMALS_FTP_USER, \
+    PETCO_LOVELOST_BASE_URL, PETCO_LOVELOST_EXTRAS, PETCADEMY_FTP_HOST, \
```
bobintetley: PETCO_LOVELOST_DEBUG rather than EXTRAS is a better name I think

src/static/js/publish_options.js
```diff
@@ -420,6 +412,25 @@ $(function() {
...
+                        info: 'Find out more at <a target="_blank" href="https://petcolove.org/lost/">petcolove.org/lost/</a>', 
+                        fields: [
+                            { id: "enabledll", label: _("Enabled"), type: "check", classes: 'enablecheck' }, 
+                            { id: "pcllemail", post_field: "PetCoLoveLostEmail", label: "Email Address", type: "text", doublesize: true }, 
+                            { id: "pcllpassword", post_field: "PetCoLoveLostPassword", label: "Password", type: "text", doublesize: true },
+                            { id: "pcllshelterid", post_field: "PetCoLoveLostShelterID", label: "Shelter ID", type: "text", doublesize: true, readonly: true, 
+                                callout: _("The information from Options > Shelter Details and the email address from Options > Email will be used. Once committed, the Shelter ID cannot be changed."),
+                                xmarkup: ' <button id="button-pcllgetshelterid">' + _("Generate a Petco Love Lost shelter ID. Once committed, this cannot be changed.") + '</button>' +
```
bobintetley: Petco only operate in the US, so no need for this string to be translatable

src/static/js/publish_options.js
```diff
@@ -420,6 +412,25 @@ $(function() {
...
                         ]}, 
+                    { id: "tab-petcolovelost", title: "Petco Love Lost", classes: 'localeus localeca haspetcolovelost',
+                        info: 'Find out more at <a target="_blank" href="https://petcolove.org/lost/">petcolove.org/lost/</a>', 
+                        fields: [
+                            { id: "enabledll", label: _("Enabled"), type: "check", classes: 'enablecheck' }, 
+                            { id: "pcllemail", post_field: "PetCoLoveLostEmail", label: "Email Address", type: "text", doublesize: true }, 
+                            { id: "pcllpassword", post_field: "PetCoLoveLostPassword", label: "Password", type: "text", doublesize: true },
+                            { id: "pcllshelterid", post_field: "PetCoLoveLostShelterID", label: "Shelter ID", type: "text", doublesize: true, readonly: true, 
+                                callout: _("The information from Options > Shelter Details and the email address from Options > Email will be used. Once committed, the Shelter ID cannot be changed."),
```
bobintetley: Doesn't need to be translatable

src/static/js/publish_options.js
```diff
@@ -420,6 +412,25 @@ $(function() {
...
+                                xmarkup: ' <button id="button-pcllgetshelterid">' + _("Generate a Petco Love Lost shelter ID. Once committed, this cannot be changed.") + '</button>' +
+                                '<img id="pcllgetshelteridwait" src="/static/images/wait/rolling_3a87cd.svg" style="height: 15px;vertical-align: middle;display: none;"/>'
+                            },
+                            { type: "raw", markup: '<button id="button-pcllpublished" class="pcllbutton">Show Published</button> ' +
+                                '<button id="button-pcllpurge" class="pcllbutton">Purge</button>',
+                                hideif: function() { return !controller.pcllextras; }
```
bobintetley: controller.haspetcolovelostdebug

src/static/js/publish_options.js
```diff
@@ -420,6 +412,25 @@ $(function() {
...
+                            },
+                            { type: "raw", markup: '<div id="pclldata" style="padding-top: 3px;"></div>',
+                                hideif: function() { return !controller.pcllextras; }
```
bobintetley: controller.haspetcolovelostdebug

src/asm3/sitedefs.py

src/main.py

src/main.py
adamdspencer77: Changed.

src/static/js/publish_options.js
adamdspencer77: Changed.

src/static/js/publish_options.js
adamdspencer77: Changed.

src/static/js/publish_options.js
adamdspencer77: Changed.

src/static/js/publish_options.js
adamdspencer77: Changed.

adamdspencer77:
Looks like there may have been changed to the API since we started this ticket. It now looks like a "taxId" is now required to create a shelter.

The documentation I have shows this field as not required but I guess this has changed:
Screenshot from 2026-07-07 14-04-49

I could add a Tax ID field to the Petco publishing options unless you think it might prove useful to add it to the Shelter Details or elsewhere. What do you think?

src/asm3/animal.py
```diff
@@ -2418,6 +2409,18 @@ def get_extra_id(dbo: Database, a: ResultRow, idtype: str) -> str:
...
+def remove_extra_id(dbo: Database, user: str, a: ResultRow, idtype: str):
+    if a.EXTRAIDS:
+        ids = []
+        for x in a.EXTRAIDS.split("|"):
+            if x.find("=") != -1:
+                k, v = x.split("=")
+                if k != idtype: ids.append( "%s=%s" % (k, v))
+        extraids = "|".join(ids)
+        a.EXTRAIDS = extraids
+        dbo.update("animal", a.ID, { "ExtraIDs": extraids }, user, setLastChanged=False)
```
bobintetley: This whole function feels a bit unnecessary. I think we could just have the existing set_extra_id function remove the id if you pass idvalue as an empty string?

src/static/js/publish_options.js
```diff
@@ -406,11 +403,6 @@ $(function() {
                         'Contact clientcare@pethealthinc.com to get a folder for automatic batch registrations of microchips.', 
                         fields: [
                             { id: "enabledfa", label: _("Enabled"), type: "check", classes: 'enablecheck' }, 
-                            { id: "fafolder", post_field: "FoundAnimalsFolder", label: "Folder name", type: "text", doublesize: true }, 
-                            { id: "faemail", post_field: "FoundAnimalsEmail", label: "Rescue group email", type: "text", doublesize: true,
```
bobintetley, line 410: Why were these lines removed? They're key info for that publisher?

adamdspencer77: No idea. I've put them back and retested :)

PR #2027, opened 2026-03-16
json_microchip_registration service API call and API key mechanism

adamdspencer77:
We've had a few enquiries from microchip registration companies who don't have their own API available for us to push data to us, and instead want to pull from us. It should be fairly straightforward to make this work. It also highlights how our method for using user accounts to access the service API seems a bit dated and annoying to manage in 2025, so we should do something about that.

    Add a new tab to publish_options for "API Keys"

This new tab allows you to generate upto 10 API keys and control which service API methods they can call. It will just show an API Key box on the left, a button to the right to generate it (which generates a UUID) and then a multi-select with the list of methods available for that key. All repeated 10 times and stored in the configuration table as APIKey1/APIMethods1, APIKey2/APIMethods2, etc.

    All methods that currently require authentication with a username and password, can accept a key instead from this list. This keeps everything backwards compatible for people using user accounts to access authenticated methods of the service API

    Add new methods for json_microchip_registration and xml_microchip_registration

    These methods accept parameters for days (default is 30), and prefix (default is empty for any chip, if one is set only chips beginning with the prefix are returned)

    The methods call a new function in publishers.base named get_microchip_data_export - it returns a much smaller dataset than get_microchip_data since it's being viewed by 3rd parties. It should cover all the contact info for the person, and most things from the details slider of animal records. Instead of using the animalpublished table (which is useless in this context), it just returns all chips with registration events that occurred in the last X days. Like get_microchip_data it should include EVENTDATE and EVENTTYPE columns to indicate what caused the registration event. EVENTTYPE will be either a movement type name or "Intake" if that was the event.

Note: The service proxy smcom_service may need to be made aware of the new key parameter.

The manual should be updated so as to document the use of keys in place of usernames and passwords with service calls.


src/asm3/service.py
```diff
@@ -613,12 +648,6 @@ def handler(post: PostedData, path: str, remoteip: str, referer: str, useragent:
     dbo.timezone_dst = asm3.configuration.timezone_dst(dbo)
     asm3.al.info("call @%s --> %s [%s]" % (username, method, querystring), "service.handler", dbo)
 
-    # If the user does not have VIEW_PERSON permissions, force stripping of personal info from 
-    # methods that support it
-    if user is not None and user.SUPERUSER is not None:
-        if not asm3.users.check_permission_map_bool(user.SUPERUSER, securitymap, asm3.users.VIEW_PERSON):
-            strip_personal = True
```
bobintetley, line 625: I don't think this code needs to be removed? You aren't actually testing strip_personal or calling strip_personal_data() in your new service endpoint? And your fake api_key user is a superuser so it wouldn't be an issue if people used api keys in place of user accounts for endpoints that previously stripped personal details?

src/asm3/service.py
```diff
@@ -600,11 +616,30 @@ def handler(post: PostedData, path: str, remoteip: str, referer: str, useragent:
...
+            validkey = False
+            keys = asm3.configuration.api_keys(dbo)
+            for k in keys:
+                if k[0] == apikey and method in k[1]:
+                    validkey = True
+                    break
+            if not validkey:
+                asm3.al.error("Invalid API Key (%s)" % method, "service.handler", dbo)
+                return ("text/plain", 0, 0, "ERROR: Invalid API Key")
+            user = api_key()
```
bobintetley: You could keep the type the same by making it a ResultRow instead of using this fake api_key class. 

```
user = dbo.first_row(dbo.query("SELECT 1 AS SuperUser, 'api_key' AS UserName"))
```

src/asm3/service.py
```diff
@@ -132,6 +136,10 @@
     "online_form_post": [ 1, 15, 15 ]
 }
 
+class api_key:
+    SUPERUSER = 1
+    USERNAME = _("API_Key")
+
```
bobintetley: suggest using a fake query further down so that this isn't needed and the types are the same for anything checking

src/asm3/configuration.py
```diff
@@ -812,6 +812,18 @@ def anonymise_personal_data(dbo: Database) -> bool:
 def anonymise_after_years(dbo: Database) -> int:
     return cint(dbo, "AnonymiseAfterYears", DEFAULTS["AnonymiseAfterYears"])
 
+def api_keys(dbo: Database) -> List:
+    keys = []
+    for a in range(1, 11):
+        if cstring(dbo, f"APIKey{str(a)}") and cstring(dbo, f"APIMethods{str(a)}"):
```
bobintetley: str() is not necessary when using f strings as the output of your {} expression is coerced to a string afterwards.

Think this would be more readable if cstring was only called the once and assigned.

```
key = cstring(dbo, f"APIKey{a}")
methods = cstring(dbo, f"APIMethods{a}")
if key != "" and methods != "":
    keys.append( [ key, methods.split(",") ] )
```

src/asm3/service.py
```diff
@@ -528,6 +536,13 @@ def strip_personal_data(rows: Results) -> Results:
     """ Shorthand to save typing the module name repeatedly """
     return asm3.publishers.base.strip_sensitive_data(rows)
 
+# def validate_api_key(key: str, method: str):
+#     registeredkeys = []
+#     for a in range(1, 11):
+#         if 
+#     appropriatekeys = [registeredkey for registeredkey in registeredkeys if "a" in x]
```
bobintetley: Should this have been removed? I think this would be a nicer pattern than having the code iterate the api key config in handler.

src/asm3/service.py
```diff
@@ -600,11 +616,30 @@ def handler(post: PostedData, path: str, remoteip: str, referer: str, useragent:
...
-            asm3.al.error("auth failed - %s/%s is not a valid username/password from %s" % (username, password, remoteip), "service.handler", dbo)
-            return ("text/plain", 0, 0, "ERROR: Invalid username and password")
-        securitymap = asm3.users.get_security_map(dbo, user.ID)
+        if apikey:
+            validkey = False
+            keys = asm3.configuration.api_keys(dbo)
+            for k in keys:
+                if k[0] == apikey and method in k[1]:
+                    validkey = True
+                    break
```
bobintetley: Put this loop in your validate_api_key function instead and call it - makes the code much more readable. Also, when you're iterating over a list or tuple of something that itself is a list or tuple, you can use destructuring in the for loop to make it much easier to read.

Eg, instead of ``` for k in keys ``` and doing k[0] and k[1], you could write it like this:

```
for key, methods in keys:
    if key == apikey and method in methods: return True
```

src/asm3/service.py
```diff
@@ -721,6 +750,16 @@ def handler(post: PostedData, path: str, remoteip: str, referer: str, useragent:
         hotlink_protect("extra_image", referer)
         image_protect(title)
         return set_cached_response(cache_key, account, "image/jpeg", 86400, 86400, asm3.dbfs.get_string(dbo, title, "/reports"))
+    
+    elif method in ("json_microchip_registrations", "xml_microchip_registrations", "csv_microchip_registrations"):
+        if not post["days"]:
+            days = 30
+        else:
+            days = post.integer("days")
```
bobintetley: this seems a bit weird. post.integer will return 0 if given invalid input, so you can do:

```
days = post.integer("days")
if days == 0: days = 30
```

src/asm3/service.py
```diff
@@ -721,6 +750,16 @@ def handler(post: PostedData, path: str, remoteip: str, referer: str, useragent:
...
+        animals = asm3.publishers.base.get_microchip_data_export(dbo, post["prefix"], days)
+        if len(animals) == 0:
+            return ("text/plain", 0, 0, "ERROR: No microchips found to register")
```
bobintetley: We're going to be returning inconsistent types here. Since the callers of this function are going to be programs rather than humans, there shouldn't be a need to return an error message here - if there are no results, they will get an empty JSON list, CSV file or XML document. method_output has checks to make sure it returns something valid if there's no results.

src/asm3/users.py
```diff
@@ -253,7 +253,7 @@ def check_permission_bool(session: Session, flag: str) -> bool:
     if has_security_flag(session.securitymap, flag): return True
     return False
 
-def check_permission_map(l: str, superuser: int, securitymap: str, flag: str) -> None:
+def check_permission_map(l: str, superuser: int, securitymap: str, flag: str, apikey: bool = False) -> None:
```
bobintetley: unless I missed something, I don't think you're passing this apikey parameter anywhere, and it's not doing anything anyway?

src/main.py
```diff
@@ -7614,6 +7594,9 @@ def controller(self, o):
         }
         asm3.al.debug("loaded lookups", "main.publish_options", dbo)
         return c
+    
+    def post_uuid(self, o):
+        return asm3.utils.uuid_str()
```
bobintetley: there are functions to generate UUIDs on the client, you don't need to do an ajax call to get the backend to do it. 

``` common.generate_uuid() ``` is available in the javascript.

File: src/static/js/publish_options.js [hunk 11]
```diff
@@ -521,6 +595,19 @@ $(function() {
                 }); 
             });
 
+            $(".asm-apikeygenbutton").button({ icons: { primary: "ui-icon-arrowrefresh-1-s" }, text: false })
+                .click(async function() {
+                    let formdata = "mode=uuid";
+                    let uuid = await common.ajax_post("publish_options", formdata);
+                    $(this).parent().find("input").val(uuid);
```
bobintetley: Use ``` common.generate_uuid() ``` - no need to go to the backend.

src/static/css/asm.css
```diff
@@ -1127,6 +1127,10 @@ select:disabled, input:disabled, textarea:disabled {
     width: 113px;
 }
 
+.asm-apimethods td{
+    padding-bottom: 10px;
+}
```
bobintetley: shouldn't be necessary to specify the td part. 

Also, this is quite a generic thing that we might want to use elsewhere. We already have classes at the top of asm.css for things like .bottomborder

I would change the name of this to .bottompadding so that we can use it elsewhere and move it to the top of the file.

src/asm3/service.py
adamdspencer77: It's just moved so it's only executing when not using api keys.

src/asm3/service.py
adamdspencer77: Done.

src/asm3/service.py
adamdspencer77: Done.

src/asm3/configuration.py
adamdspencer77: Done.

Issue #1911, opened 2025-10-28
json_microchip_registration service API call and API key mechanism


bobintetley:
We've had a few enquiries from microchip registration companies who don't have their own API available for us to push data to us, and instead want to pull from us. It should be fairly straightforward to make this work. It also highlights how our method for using user accounts to access the service API seems a bit dated and annoying to manage in 2025, so we should do something about that.

1. Add a new tab to publish_options for "API Keys"

This new tab allows you to generate upto 10 API keys and control which service API methods they can call. It will just show an API Key box on the left, a button to the right to generate it (which generates a UUID) and then a multi-select with the list of methods available for that key. All repeated 10 times and stored in the configuration table as APIKey1/APIMethods1, APIKey2/APIMethods2, etc.

2. All methods that currently require authentication with a username and password, can accept a key instead from this list. This keeps everything backwards compatible for people using user accounts to access authenticated methods of the service API

3. Add new methods for json_microchip_registration and xml_microchip_registration

4. These methods accept parameters for days (default is 30), and prefix (default is empty for any chip, if one is set only chips beginning with the prefix are returned)

5. The methods call a new function in publishers.base named get_microchip_data_export - it returns a much smaller dataset than get_microchip_data since it's being viewed by 3rd parties. It should cover all the contact info for the person, and most things from the details slider of animal records. Instead of using the animalpublished table (which is useless in this context), it just returns all chips with registration events that occurred in the last X days. Like get_microchip_data it should include EVENTDATE and EVENTTYPE columns to indicate what caused the registration event. EVENTTYPE will be either a movement type name or "Intake" if that was the event.

Note: The service proxy smcom_service may need to be made aware of the new key parameter.

The manual should be updated so as to document the use of keys in place of usernames and passwords with service calls.

PR #2028, opened 2026-03-16
service API call: json_monthly_stats

adamdspencer77:
There are other data services besides SAC out there, such as SPDA (Best Friends). We've had enquiries about integration, but they don't have an API and would prefer to pull data from us.

Add new service methods for json/xml/csv_monthly_stats

These methods accept a month, year and species parameter (mandatory) and will return basically the same data that we send to SAC (see sacmetrics.py). In fact, it should be possible to simply call sacmetrics.py with the parameters to retrieve the data we want.

Use the method_output pattern that the other json/xml/csv methods in service.py use.

These methods should be authenticated and it would be best if https://github.com/sheltermanager/asm3/issues/1911 is done before this ticket so that API key auth is available.

### DIFF
--- a/doc/manual/service.rst
+++ b/doc/manual/service.rst
@@ -1164,6 +1164,20 @@ determines whether the format returned is JSON, CSV or XML::
 
     http://localhost:5000/service?method=json_held_animals&username=user&password=letmein
 
+json_monthly_stats, xml_monthly_stats, csv_monthly_stats
+--------------------------------------------------------
+
+.. rubric:: Cache time: 1 hour 
+.. rubric:: Permissions required: VIEW_ANIMAL, VIEW_MOVEMENT
+.. rubric:: Requires username/password: YES
+
+Requires month (integer), year (integer) and species (string) parameters. 
+Returns a dataset containing monthly statistics. 
+The method determines whether the format returned is JSON, CSV or XML::
+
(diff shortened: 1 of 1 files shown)


src/asm3/service.py
```diff
@@ -857,6 +861,16 @@ def handler(post: PostedData, path: str, remoteip: str, referer: str, useragent:
...
+    
+    elif method in ("json_monthly_stats", "xml_monthly_stats", "csv_monthly_stats"):
+        asm3.users.check_permission_map(l, user.SUPERUSER, securitymap, asm3.users.VIEW_ANIMAL)
+        asm3.users.check_permission_map(l, user.SUPERUSER, securitymap, asm3.users.VIEW_MOVEMENT)
+        month = post.integer("month")
+        year = post.integer("year")
+        speciesname = post["species"]
+        pc = asm3.publish.PublishCriteria(asm3.configuration.publisher_presets(dbo))
+        ms = asm3.publishers.sacmetrics.SACMetricsPublisher(dbo, pc).processStats(month, year, speciesname)
+        return set_cached_response(cache_key, account, method_mimetype(method), 3600, 3600, method_output(method, l, [ms,]))
```
bobintetley, line 873: Code all looks good, manual needs updating with info on these 3 calls.

adamdspencer77: Done.

PR #2029, opened 2026-03-16
Declining weight alert

adamdspencer77:
Add new fields - animal.Weight1 and animal.Weight2, which store the previous two values of the weight field (for a total of 3 quick to read samples).

When Weight1 is non-null and lower than Weight, show an alert for "X animals are losing weight" on the front screen - needs matching search.


src/asm3/dbupdates/35119.py
```diff
@@ -0,0 +1,4 @@
+from asm3.dbupdate import add_column
+
+add_column(dbo, "animal", "Weight1", dbo.type_float)
+add_column(dbo, "animal", "Weight2", dbo.type_float)
```
bobintetley, line 4: we should be adding indexes for both of these fields since the alert query is going to be checking them for every animal record on every visit to the home page

(and don't forget to add them to the animal table definition in dbupdate.py as well as this update file)

src/static/js/header.js
```diff
@@ -720,7 +720,7 @@ header = {
         const keywords = [ "activelost", "activefound", "donors", "deceased", "hold", "holdtoday", 
             "notforadoption", "onshelter", "quarantine", "forpublish", "reservenohomecheck", "notmicrochipped",
             "aco", "banned", "donors", "drivers", "homechecked", "homecheckers", 
-            "fosterers", "homecheckers", "members", "people", "retailers", "shelters", "staff", 
+            "fosterers", "homecheckers", "members", "lostweight", "people", "retailers", "shelters", "staff",
```
bobintetley: it's kind of a mess, not really as a result of your changes, but these keywords should all be in alphabetical order

src/asm3/dbupdates/35119.py
adamdspencer77: Hopefully sorted.

src/static/js/header.js
adamdspencer77: Alphabetised and duplicate "homecheckers" removed.

src/asm3/search.py
```diff
@@ -463,6 +464,12 @@ def cp(p):
         explain = _("Found animals reported in the last 30 days.", l)
         if cp(asm3.users.VIEW_FOUND_ANIMAL):
             ar(asm3.lostfound.get_foundanimal_find_simple(dbo, "", limit=limit, siteid=siteid), "FOUNDANIMAL", fasort)
+    
+    elif q == "lostweight":
+        explain = _("Animals that lost weight on their last 2 consecutive weighings.", l)
+        if cp(asm3.users.VIEW_ANIMAL):
+            rlist = dbo.query(f"{asm3.animal.get_animal_query(dbo)} WHERE a.Weight2 > a.Weight1 AND a.Weight1 > a.Weight AND a.Archived = 0")
```
bobintetley: This query should be in animal.py and called from search.py - try to keep actual queries in their modules rather than scattering them about the place!

Also, this should be using get_animal_brief_query for performance I think rather than the full animal_query (we should always prefer to use brief query when retrieving more than one animal record)

adamdspencer77: Done.

Tetzor:
Change icon from medical icon to health icon.

PR #2031, opened 2026-03-17
Online form: PDF upload field

adamdspencer77:
More than one user has asked if they can have the ability to upload PDFs via an online form. This is probably acceptable with a size limitation to prevent abuse.

### DIFF
--- a/src/asm3/onlineform.py
+++ b/src/asm3/onlineform.py
@@ -585,7 +585,7 @@ def get_onlineformincoming_html(dbo: Database, collationid: int,
         elif v.startswith("data:application/pdf"):
             h.append('<tr>')
             h.append('<td>%s</td>' % label )
-            h.append('<td>%s</td>' % asm3.i18n._("Uploaded PDF file"))
+            h.append('<td>%s</td>' % asm3.i18n._("Uploaded PDF file", dbo.locale))
             h.append('</tr>')
         elif f.FIELDNAME == "useragent":
             # Some user agent strings can be huge and without wrappable characters,
@@ -1051,7 +1051,7 @@ def insert_onlineformincoming_from_form(dbo: Database, post: PostedData, remotei
                     # Remove prefix of data:image/jpeg;base64, and decode
                     images.append( ("%s.jpg" % fieldname, "image/jpeg", asm3.utils.base64decode(v[v.find(",")+1:])) )
                 if fieldtype == FIELDTYPE_PDF and v.startswith("data:application/pdf"):
-                    # Remove prefix of data:image/jpeg;base64, and decode
+                    # Remove prefix of data:application/pdf;base64, and decode
                     pdfs.append( ("%s.pdf" % fieldname, "application/pdf", asm3.utils.base64decode(v[v.find(",")+1:])) )
 
             # Do the insert
@@ -1421,7 +1421,7 @@ def attach_form(dbo: Database, username: str, linktype: int, linkid: int, collat
                 if linktype == 0:
                     d["excludefrompublish"] = "1" # auto exclude images for animals to prevent them going to adoption websites
                 asm3.media.attach_file_from_form(dbo, username, linktype, linkid, asm3.media.MEDIASOURCE_ONLINEFORM, asm3.utils.PostedData(d, dbo.locale))
-            elif f.VALUE.startswith("data:application/pdf"):
+            elif f.VALUE.startswith("data:application/pdf") and len(f.VALUE.length) <= 2097152:
                 d = {
                     "retainfor":    str(retainfor),
                     "filename":     "document.pdf",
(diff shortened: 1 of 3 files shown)


src/static/js/onlineform_extra.js
```diff
@@ -105,6 +105,23 @@ $(document).ready(function() {
         imreader.readAsDataURL(file);
     };
 
+    // Loads and scales a pdf into an pdf form field for upload
+    const process_pdf = function(field) {
+        let filedata = null;
+        let imreader = new FileReader();
+        imreader.onload = function(e) { 
+            filedata = e.target.result;
+            if (filedata.length > 384000) { alert("File is too large"); field.val(""); return; }
```
adamdspencer77: Need to check if this size limit is practical.

src/asm3/onlineform.py
```diff
@@ -572,11 +577,16 @@ def get_onlineformincoming_html(dbo: Database, collationid: int,
...
-        elif v.startswith("data:"):
+        elif v.startswith("data:image"):
             h.append('<tr>')
             h.append('<td>%s</td>' % label )
             h.append('<td><img src="%s" border="0" /></td>' % v)
             h.append('</tr>')
+        elif v.startswith("data:application/pdf"):
+            h.append('<tr>')
+            h.append('<td>%s</td>' % label )
+            h.append('<td>%s</td>' % asm3.i18n._("Uploaded PDF file"))
```
bobintetley: without the ``` , l)``` parameter to i18n._ it doesn't actually translate and always returns English. That parameter tells it what locale you want the translation in.

``` asm3.i18n._("Uploaded PDF file", l) ```

File: src/asm3/onlineform.py [hunk 3]
```diff
@@ -1039,6 +1050,9 @@ def insert_onlineformincoming_from_form(dbo: Database, post: PostedData, remotei
                 if fieldtype == FIELDTYPE_IMAGE and v.startswith("data:image/jpeg"):
                     # Remove prefix of data:image/jpeg;base64, and decode
                     images.append( ("%s.jpg" % fieldname, "image/jpeg", asm3.utils.base64decode(v[v.find(",")+1:])) )
+                if fieldtype == FIELDTYPE_PDF and v.startswith("data:application/pdf"):
+                    # Remove prefix of data:image/jpeg;base64, and decode
```
bobintetley: not sure this comment is right

src/asm3/onlineform.py
```diff
@@ -1407,6 +1421,16 @@ def attach_form(dbo: Database, username: str, linktype: int, linkid: int, collat
                 if linktype == 0:
                     d["excludefrompublish"] = "1" # auto exclude images for animals to prevent them going to adoption websites
                 asm3.media.attach_file_from_form(dbo, username, linktype, linkid, asm3.media.MEDIASOURCE_ONLINEFORM, asm3.utils.PostedData(d, dbo.locale))
+            elif f.VALUE.startswith("data:application/pdf"):
```
bobintetley: We can't purely trust the frontend to validate the file size for us. We should be checking it here just in case.

src/static/js/onlineform_extra.js
bobintetley: it's definitely a bit on the small side. I'd go with at least a megabyte  1048576

adamdspencer77: Changed to 1048576.

src/asm3/onlineform.py

src/asm3/onlineform.py

src/asm3/onlineform.py

src/static/js/onlineform_extra.js
adamdspencer77: Upped to 2097152 after discussing.

src/static/js/onlineform_extra.js
```diff
@@ -105,6 +105,23 @@ $(document).ready(function() {
...
+            filedata = e.target.result;
+            if (filedata.length > 2097152) { alert("File is too large"); field.val(""); return; }
+            $("input[name='" + field.attr("data-name") + "']").val(filedata);
+        };
+        let file = field[0].files[0];
+
+        // Is this a pdf? If not, stop now
+        if (!file.type.match('application.pdf')) { alert("File is not a pdf"); field.val(""); return; }
+        
+        imreader.readAsDataURL(file);
```
bobintetley: Very minor thing but variable name should be something else, like pdfreader instead of imreader since it's not reading images :)

src/static/js/onlineform_extra.js
```diff
@@ -105,6 +105,23 @@ $(document).ready(function() {
         imreader.readAsDataURL(file);
     };
 
+    // Loads and scales a pdf into an pdf form field for upload
+    const process_pdf = function(field) {
+        let filedata = null;
+        let imreader = new FileReader();
+        imreader.onload = function(e) { 
+            filedata = e.target.result;
+            if (filedata.length > 2097152) { alert("File is too large"); field.val(""); return; }
```
bobintetley: I'd make the message "PDF file is too large", otherwise it might confuse the user if there are image uploads in the form. Also, maybe we could set the focus to this field so that the browser scrolls to it and the user can see which one is the issue.

src/static/js/onlineform_extra.js
adamdspencer77: Changed.

src/static/js/onlineform_extra.js
adamdspencer77: Done.

PR #2042, opened 2026-03-30
Stock levels created by Product movements do not inherit price information from the Product

adamdspencer77:
Found when working on another ticket.

### DIFF
--- a/src/asm3/stock.py
+++ b/src/asm3/stock.py
@@ -386,8 +386,8 @@ def insert_productmovement_from_form(dbo: Database, post: PostedData, username: 
             slpost["low"] = 0
             slpost["expiry"] = post["expiry"]
             slpost["batchnumber"] = post["batch"]
-            slpost["cost"] = post.integer("COSTPRICE")
-            slpost["unitprice"] = post.integer("RETAILPRICE")
+            slpost["cost"] = post.integer("costprice")
+            slpost["unitprice"] = post.integer("retailprice")
             slpost["usagedate"] = python2display(dbo.locale, dbo.today())
             slpost["usagetype"] = usagetypeid
             slpost["comments"] = post["comments"]
--- a/src/asm3/stock.py
+++ b/src/asm3/stock.py
@@ -336,9 +336,11 @@ def insert_productmovement_from_form(dbo: Database, post: PostedData, username: 
     unitratio = post.integer("unitratio")
 
     # Get current stock levels of the selected product
-    stocklevels = dbo.query("SELECT ID, BatchNumber, Balance, Cost, UnitPrice, Total, Low, Expiry, StockLocationID FROM stocklevel " \
-        "WHERE ProductID = ? AND BatchNumber = ? AND StockLocationID = ? " \
-        "ORDER BY Balance", [ post.integer("productid"), post["batch"], fromlocation ])
+    stocklevels = dbo.query("SELECT *, " \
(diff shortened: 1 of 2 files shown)


src/asm3/stock.py
```diff
@@ -341,62 +336,72 @@ def insert_productmovement_from_form(dbo: Database, post: PostedData, username:
...
-                    remaining = stocklevel["BALANCE"] - quantity
-                    quantity = quantity - stocklevel["BALANCE"]
+    stocklevels = dbo.query("SELECT *, " \
+        "CASE WHEN BatchNumber = ? THEN 1 ELSE 0 END AS BatchMatch " \
+        "FROM stocklevel " \
+        "WHERE ProductID = ? " \
+        "ORDER BY BatchMatch desc, Balance", [ post["batch"], post.integer("productid") ])
+    startingstockbalance = 0
+    for stocklevel in stocklevels:
+        startingstockbalance += stocklevel["BALANCE"]
```
bobintetley: sorry, same thing again with accessing result values via stocklevel.BALANCE etc. Lots of it below :)

src/asm3/stock.py
```diff
@@ -341,62 +336,72 @@ def insert_productmovement_from_form(dbo: Database, post: PostedData, username:
...
+    if fromlocation and startingstockbalance < quantity:
+        raise asm3.utils.ASMValidationError(_("Impossible stock movement.", l))
+    else:
+        if fromlocation != 0:
+            batchexhausted = False # batchexhausted = True when we have run out stock of the correct batch at the current location
+            for stocklevel in stocklevels:
+                if quantity == 0:
+                    break
+                if not stocklevel["BATCHMATCH"]:
```
bobintetley: this seems an odd indirection and the fact that it's negated made me think about it for a minute. Why not just do the comparison in python? Couldn't this have been ``` if stocklevel.BATCHNUMBER != post["batch"] ``` ?

src/asm3/stock.py
```diff
@@ -341,62 +336,72 @@ def insert_productmovement_from_form(dbo: Database, post: PostedData, username:
...
-                if stocklevel["BALANCE"] < 0:
-                    remaining = stocklevel["BALANCE"] - quantity
-                    quantity = 0
-                elif quantity >= stocklevel["BALANCE"]:
-                    remaining = 0
-                    quantity = quantity - stocklevel["BALANCE"]
-                else:
-                    remaining = stocklevel["BALANCE"] - quantity
-                    quantity = quantity - stocklevel["BALANCE"]
+    stocklevels = dbo.query("SELECT *, " \
```
bobintetley, line 339: I try not to write different select statements whenever I do a query as it means that code reading the results may find things are missing.

You should be using get_stocklevel_query(dbo) to get the select query and any joins and adding an extra joins and your WHERE and ORDER to it like we do everywhere else.

src/asm3/stock.py
adamdspencer77: Sorted.

src/asm3/stock.py
adamdspencer77: The BATCHMATCH variable/column created in the query was necessary to order the rows correctly and with it already existing made for the more readable if statement.

src/asm3/stock.py
adamdspencer77: I can't in this case because I need the CASE statement I added to order the rows correctly.

PR #2045, opened 2026-04-02
clinic interface: stock

adamdspencer77:
Future expansion for the clinic interface - allow deducting items from stock and printing medicine labels for frequency, dosage and animal name (this could be free form text).


src/asm3/clinic.py
```diff
@@ -238,6 +238,22 @@ def update_invoice_from_form(dbo: Database, username: str, post: PostedData) ->
...
+    Adds a comma seperated string of associated stockusage records to a clinicinvoiceitem.
+    If updateprice = True, updates clinicinvoiceitem.Amount with the combined value of the 
+    associated stock usage ids.
+    """
+    payload = {
+        "StockUsageIDs":    ",".join([str(stockusageid) for stockusageid in stockusageids])
+    }
+    if updateprice:
+        price = dbo.query_float(
+            f'SELECT SUM(su.Quantity * sl.UnitPrice * -1) AS Price FROM stockusage su INNER JOIN stocklevel sl ON su.StockLevelID = sl.ID WHERE su.ID IN ({",".join([str(stockusageid) for stockusageid in stockusageids])})'
```
bobintetley: I'm not keen on long expressions in f-strings. I'd do something like ``` inclause = ",".join([str(stockusageid) for stockusageid in stockusageids]) ``` and then just put {inclause} in the query

src/asm3/stock.py
bobintetley: always use property members when accessing columns in results, eg: stocklevel.BALANCE, stocklevel.BATCHNUMBER - it's easier to read and it means we can tell at a glance that it's a database result rather than a dictionary

src/asm3/clinic.py
```diff
@@ -238,6 +238,22 @@ def update_invoice_from_form(dbo: Database, username: str, post: PostedData) ->
     }, username)
     update_appointment_total(dbo, post.integer("appointmentid"))
 
+def update_invoice_stock_movements(dbo: Database, username: str, invoiceid: int, stockusageids: List[int], updateprice: bool = False):
+    """
+    Adds a comma seperated string of associated stockusage records to a clinicinvoiceitem.
+    If updateprice = True, updates clinicinvoiceitem.Amount with the combined value of the 
+    associated stock usage ids.
+    """
+    payload = {
```
bobintetley: weird variable name, this for parameters to an update query - these are usually called "values"

src/asm3/clinic.py
adamdspencer77: Done.

src/asm3/stock.py
adamdspencer77: Done.

src/asm3/clinic.py
adamdspencer77: Changed.

Tetzor:
Get rid of label button in clinic invoice screen.

PR #2049, opened 2026-04-09
MedicalType[Type]Name wordprocessor token

adamdspencer77:
We currently only have two tokens for MedicalType[Type]Given and Due. A customer asked if we could have MedicalType[Type]Name as well so that their document can indicate the specific treatment administered for that medical type.

Should be fairly straightforward, update medical.py to add another subquery to get_medical_types_animal to retrieve the TreatmentName column from animalmedical, and have the code in wordprocessor.py use it.

### DIFF
--- a/src/asm3/medical.py
+++ b/src/asm3/medical.py
@@ -337,11 +337,11 @@ def get_batch_for_vaccination_types(dbo: Database) -> Results:
 def get_medical_types_animal(dbo: Database, animalid: int) -> Results:
     """
     Returns a recordset of medicaltypes for an animal:
-    MEDICALTYPE, DATEREQUIRED, LASTGIVEN
+    MEDICALTYPE, TREATMENTNAME, DATEREQUIRED, LASTGIVEN
     """
     sql = "SELECT mt.MedicalTypeName, " \
         "(" \
-            "SELECT DateRequired " \
+            f"SELECT {dbo.sql_concat(['DateRequired', '"$$"', 'TreatmentName'])} " \
             "FROM animalmedicaltreatment " \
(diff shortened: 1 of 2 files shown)


adamdspencer77:
Turned out to be a little more complicated. The treatment name for the due treatment can be different from the treatment name for the given treatment. Have created tokens for MedicalType[Type]GivenName and MedicalType[Type]DueName.

src/asm3/medical.py
```diff
@@ -337,19 +337,19 @@ def get_batch_for_vaccination_types(dbo: Database) -> Results:
 def get_medical_types_animal(dbo: Database, animalid: int) -> Results:
     """
     Returns a recordset of medicaltypes for an animal:
-    MEDICALTYPE, DATEREQUIRED, LASTGIVEN
+    MEDICALTYPE, DATEREQUIRED$$TREATMENTNAME, LASTGIVEN$$TREATMENTNAME
     """
     sql = "SELECT mt.MedicalTypeName, " \
         "(" \
-            "SELECT DateRequired " \
+            f"SELECT {dbo.sql_concat(['DateRequired', '"$$"', 'TreatmentName'])} " \
```
bobintetley: Apart from being seriously smelly, concat-ing different types of varchar and timestamp like this will break on postgres.

src/asm3/medical.py
bobintetley: I think keep DateRequired and DateGiven subqueries as they were. Add 2 more subqueries for DateRequiredTreatmentName and DateGivenTreatmentName. There's no need to do any concat-ing or splitting then.

adamdspencer77: Done.

src/asm3/medical.py
adamdspencer77: Sorted.

PR #2054, opened 2026-04-14
Funded Payments: recording how payments have been spent (DBU 35120)

adamdspencer77:
This is a ticket to consider how we might record how donations have been spent.

We had a user who received large donations from people. When poorer folks would adopt animals, they would charge them a token amount like $25 and take the rest from one of these donations. They were doing it by breaking the payment up into smaller payments so they could see where that donation had been spent. Obviously, by doing this though, the original receipt of the donation was lost, and it looked like that person had made lots of payments for adoptions fees on different animals.

From an auditing/data loss point of view, this is pretty bad. But it would be nice if there were a way to record what a payment had been spent on.

As a first stab/thought at it:

    Use a simple self-join in the ownerdonation table for a new FundedByOwnerDonationID column. This would allow you to pick a payment that the current payment was funded from.
    In order to stop double counting, all of the payment reports would need to exclude payments where this field was set.
    Choosing a funding payment would make the screen hide the payment method, check no, tax and other fields that don't make sense for a funded payment.
    Funded payments would also not create matching transactions in the accounts.

Given how little of the payment info is needed, would a new table/screen be better? Probaby not, but worth considering.

These funded payments should be displayed differently in the list so it's clear what they are.

Consideration needs to be given as to how funded payments can be chosen. I'm thinking an autocomplete widget where the user types the receipt number of the funded payment and the dropdown list has some basic info (the receipt number, payment amount and payee)

Since this might potentially cause confusion, this feature should be gated behind a config option and off by default.

Needs documenting clearly in the manual and a callout next to the "funded by" field.


reports/payments_funded_by_receipt_number.rep
```diff
@@ -0,0 +1,56 @@
...
+en
+###
+SELECT od.ReceiptNumber, od.Date, od.Donation, o.OwnerName 
+FROM ownerdonation od 
+INNER JOIN ownerdonation fd ON od.FundedByOwnerDonationID = fd.ID 
+INNER JOIN owner o ON od.OwnerID = o.ID 
+WHERE fd.ReceiptNumber = '$ASK STRING Enter the receipt number$';
+###
+$$HTMLHEADER
+<DOCTYPE html>
```
bobintetley: Did you copy the payment receipt report? There shouldn't be any need to be setting a HTMLHEADER. 

The report should be including information about the top level/funding payment as well as all the payments it has funded. I would expect joins to animal as well as owner so you can see what payments related to which animal. You can use $$GROUP to separate the funded payment from the sub payments.

reports/payments_funded_by_receipt_number.rep
```diff
@@ -0,0 +1,56 @@
+Funded Payments by Fund
```
bobintetley: I think just "Funded Payments" is ok?

src/asm3/financial.py
```diff
@@ -851,32 +855,35 @@ def insert_donation_from_form(dbo: Database, username: str, post: PostedData) ->
...
+        "VATRate":                  post.floating("vatrate"),
+        "VATAmount":                post.integer("vatamount"),
+        "Comments":                 post["comments"]
     }, username)
 
-    if asm3.configuration.donation_trx_override(dbo):
-        update_matching_donation_transaction(dbo, username, donationid, post.integer("destaccount"))
-    else:
-        update_matching_donation_transaction(dbo, username, donationid)
+    if post.integer("funding") == 0:
```
bobintetley, line 883: include a comment about not creating a matching transaction for funded payments

src/static/js/donations.js
```diff
@@ -56,7 +60,8 @@ $(function() {
                     { json_field: "ANIMALID", post_field: "animal", label: _("Animal"), type: "animal" },
                     { json_field: "OWNERID", post_field: "person", label: _("Person"), type: "person", validation: "notzero" },
                     { json_field: "MOVEMENTID", post_field: "movement", label: _("Movement"), type: "select", options: "" },
-                    { json_field: "COMMENTS", post_field: "comments", label: _("Comments"), type: "textarea" }
+                    { json_field: "COMMENTS", post_field: "comments", label: _("Comments"), type: "textarea" },
+                    { json_field: "ISFUNDINGSOURCE", post_field: "isfundingsource", label: _("Available for Funding"), type: "check" }
```
bobintetley: I think the label should be "Funding Source" too. Have a callout on it that says something like "This payment can be used as a source of funds for other payments".

reports/payments_funded_by_receipt_number.rep
adamdspencer77: Changed.

src/static/js/donations.js
adamdspencer77: Done.

src/asm3/financial.py
adamdspencer77: Done.

reports/payments_funded_by_receipt_number.rep
adamdspencer77: Report tweaked and second report added to show funded payments in a date range.

src/static/js/donations.js
```diff
@@ -25,6 +25,10 @@ $(function() {
                 fields: [
                     { json_field: "DONATIONTYPEID", post_field: "type", label: _("Type"), type: "select", options: { displayfield: "DONATIONNAME", valuefield: "ID", rows: controller.donationtypes }},
                     { json_field: "DONATIONPAYMENTID", post_field: "payment", label: _("Method"), type: "select", options: { displayfield: "PAYMENTNAME", valuefield: "ID", rows: controller.paymentmethods }},
+                    { json_field: "FUNDEDBYOWNERDONATIONID", post_field: "funding", label: _("Fund"), type: "select", options: { displayfield: "FUNDNAME", valuefield: "ID",
```
bobintetley, line 28: I can see you've put it a toggle in onload, but I think it's more readable to use a hideif function on this element itself

src/static/js/donations.js
```diff
@@ -56,7 +60,10 @@ $(function() {
                     { json_field: "ANIMALID", post_field: "animal", label: _("Animal"), type: "animal" },
                     { json_field: "OWNERID", post_field: "person", label: _("Person"), type: "person", validation: "notzero" },
                     { json_field: "MOVEMENTID", post_field: "movement", label: _("Movement"), type: "select", options: "" },
-                    { json_field: "COMMENTS", post_field: "comments", label: _("Comments"), type: "textarea" }
+                    { json_field: "COMMENTS", post_field: "comments", label: _("Comments"), type: "textarea" },
+                    { json_field: "ISFUNDINGSOURCE", post_field: "isfundingsource", label: _("Funding Source"), type: "check",
```
bobintetley, line 64: this should have a hideif instead

src/static/js/donations.js
```diff
@@ -89,6 +96,8 @@ $(function() {
                             // Only allow destination account to be overridden when the received date
                             // hasn't been set yet.
                             $("#destaccountrow").toggle( config.bool("DonationTrxOverride") && !row.DATE );
+                            $("#fundingrow").toggle(config.bool("FundedPaymentsEnabled"));
+                            $("#isfundingsourcerow").toggle(config.bool("FundedPaymentsEnabled"));
```
bobintetley, line 100: these 2 lines are redundant with the hideifs ...

src/static/js/donations.js
```diff
@@ -173,6 +184,8 @@ $(function() {
                                 donations.create_semaphore = true;
                                 // Editing can hide the destination account field - show it if the option is on
                                 $("#destaccountrow").toggle(config.bool("DonationTrxOverride"));
+                                $("#fundingrow").toggle(config.bool("FundedPaymentsEnabled"));
+                                $("#isfundingsourcerow").toggle(config.bool("FundedPaymentsEnabled"));
```
bobintetley, line 188: also redundant with hideif

src/static/js/donations.js
adamdspencer77: Changed.

src/static/js/donations.js
adamdspencer77: Done.

src/static/js/donations.js
adamdspencer77: Removed.

src/static/js/donations.js
adamdspencer77: Removed.

bobintetley:
Add an icon next to the payment type in the list to indicate funded payments rather than showing them as greyed out. Choose wisely from existing icons or create a new one.

When editing funding payments in the dialog, hide the due date field too.

reports/payment_breakdown_by_date.rep
```diff
@@ -2,7 +2,7 @@ Payment Breakdown By Date
 ###
 Financial
 ###
-34204/Any rev02
+34204/Any rev03
```
bobintetley: all of the database version numbers on these reports need increasing to 35120 or they will fail on earlier databases that do not have the FundedByOwnerDonationID column

adamdspencer77: Done.

PR #2060, opened 2026-04-20
Online form: create logs

adamdspencer77:
Some users would like the ability for online forms to create a log entry. I think this is only really useful for animals so make it contingent on an animalname field being supplied and it referencing an existing animal. There should be some way of specifying the log type in the form as well.

The log entry created should use the same format as daily obs, with field=value pairs in the string for all of the fields in the form (except for the built in ones).

One possible use for this is for fosterers recording daily obs for the animals in their care. By making a form that has the same fields as your daily obs configuration, they can enter them without having to have access to the daily obs screen or us adding fosterers as locations to daily obs (which makes no sense since it's about batching entering obs for lots of animals in one location).

### DIFF
--- a/src/asm3/configuration.py
+++ b/src/asm3/configuration.py
@@ -349,6 +349,7 @@ DEFAULTS = {
     "MoveAdoptGeneratePaperwork": "Yes",
     "MoveAdoptDonationsEnabled": "No",
     "JSWindowPrint": "Yes",
+    "OnlineFormDefaultLogType": "3",
     "OnlineFormSpamHoneyTrap": "Yes",
     "OnlineFormSpamUACheck": "No",
     "OnlineFormSpamFirstnameMixCase": "Yes",
@@ -1033,6 +1034,9 @@ def default_media_notes_from_file(dbo: Database) -> bool:
 
 def default_nonsheltertype(dbo: Database) -> int:
     return cint(dbo, "AFNonShelterType", 40)
+
+def default_onlineformlogtype(dbo: Database) -> int:
+    return cint(dbo, "OnlineFormDefaultLogType", 3)
 
 def default_payment_method(dbo: Database) -> int:
     return cint(dbo, "AFDefaultPaymentMethod", int(DEFAULTS["AFDefaultPaymentMethod"]))
--- a/src/asm3/configuration.py
+++ b/src/asm3/configuration.py
@@ -1029,14 +1029,14 @@ def default_location(dbo: Database) -> int:
 def default_log_filter(dbo: Database) -> int:
     return cint(dbo, "AFDefaultLogFilter", 0)
 
+def default_log_type(dbo: Database) -> int:
+    return cint(dbo, "AFDefaultLogType", 0)
+
 def default_media_notes_from_file(dbo: Database) -> bool:
     return cboolean(dbo, "DefaultMediaNotesFromFile", DEFAULTS["DefaultMediaNotesFromFile"] == "Yes")
 
 def default_nonsheltertype(dbo: Database) -> int:
     return cint(dbo, "AFNonShelterType", 40)
-
-def default_onlineformlogtype(dbo: Database) -> int:
-    return cint(dbo, "OnlineFormDefaultLogType", 3)
 
 def default_payment_method(dbo: Database) -> int:
     return cint(dbo, "AFDefaultPaymentMethod", int(DEFAULTS["AFDefaultPaymentMethod"]))
--- a/src/asm3/configuration.py
+++ b/src/asm3/configuration.py
@@ -349,7 +349,6 @@ DEFAULTS = {
     "MoveAdoptGeneratePaperwork": "Yes",
     "MoveAdoptDonationsEnabled": "No",
     "JSWindowPrint": "Yes",
-    "OnlineFormDefaultLogType": "3",
     "OnlineFormSpamHoneyTrap": "Yes",
     "OnlineFormSpamUACheck": "No",
     "OnlineFormSpamFirstnameMixCase": "Yes",
--- a/src/asm3/dbupdates/35121.py
+++ b/src/asm3/dbupdates/35121.py
@@ -1,0 +1,8 @@
+from asm3.dbupdate import add_column, add_index, execute
+add_column(dbo, "onlineform", "LogTypeID", dbo.type_integer)
+add_index(dbo, "onlineform_LogTypeID", "onlineform", "LogTypeID")
+execute(dbo, "UPDATE onlineform SET LogTypeID = 0")
+
+add_column(dbo, "onlineformincoming", "FormID", dbo.type_integer)
+add_index(dbo, "onlineformincoming_FormID", "onlineformincoming", "FormID")
+execute(dbo, "UPDATE onlineformincoming SET FormID = 0")
--- a/src/asm3/dbupdates/35121.py
+++ b/src/asm3/dbupdates/35121.py
@@ -1,8 +1,0 @@
-from asm3.dbupdate import add_column, add_index, execute
-add_column(dbo, "onlineform", "LogTypeID", dbo.type_integer)
-add_index(dbo, "onlineform_LogTypeID", "onlineform", "LogTypeID")
-execute(dbo, "UPDATE onlineform SET LogTypeID = 0")
-
-add_column(dbo, "onlineformincoming", "FormID", dbo.type_integer)
-add_index(dbo, "onlineformincoming_FormID", "onlineformincoming", "FormID")
-execute(dbo, "UPDATE onlineformincoming SET FormID = 0")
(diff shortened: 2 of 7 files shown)


src/asm3/dbupdates/35121.py
```diff
@@ -0,0 +1,8 @@
+from asm3.dbupdate import add_column, add_index, execute
+add_column(dbo, "onlineform", "LogTypeID", dbo.type_integer)
+add_index(dbo, "onlineform_LogTypeID", "onlineform", "LogTypeID")
+execute(dbo, "UPDATE onlineform SET LogTypeID = 0")
+
+add_column(dbo, "onlineformincoming", "FormID", dbo.type_integer)
+add_index(dbo, "onlineformincoming_FormID", "onlineformincoming", "FormID")
+execute(dbo, "UPDATE onlineformincoming SET FormID = 0")
```
bobintetley, line 8: This isn't how I imagined this going!

I was expecting the logtype to be a field that is submitted as part of the form itself. We would use some kind of guess method to try and match it to a log type in the database, but if we can't find one or the field isn't included in the form, we just use the default log type from the configuration.

It should not have required any changes to the database!

src/asm3/onlineform.py
```diff
@@ -1958,3 +1967,18 @@ def auto_remove_old_incoming_forms(dbo: Database) -> None:
     for r in rows:
         delete_onlineformincoming(dbo, "system", r.COLLATIONID)
     asm3.al.debug("removed %s incoming forms older than %s days" % (len(rows), removeafter), "onlineform.auto_remove_old_incoming_forms", dbo)
+
+def log_animal_form(dbo: Database, username: str, animalid: int, logtypeid: int, collationid: int):
```
bobintetley, line 1971: I imagined this being a create_animal_log function, similar to create_animal, which uses the animalname field to find the animal to create the log message on.

src/asm3/dbupdates/35121.py
adamdspencer77: DB Update removed and mechanism reworked.

src/asm3/onlineform.py

adamdspencer77:
Think I got confused with the branch numbers on some of these commits. I hope it doesn't cause any problems.

src/asm3/onlineform.py
```diff
@@ -92,7 +92,7 @@
 SPAMBOT_TXT = 'a_emailaddress'
 
 # Fields that are added to forms by the system and are not user enterable
-SYSTEM_FIELDS = [ "useragent", "ipaddress", "retainfor", "formreceived", "mergeperson", "processed" ]
+SYSTEM_FIELDS = [ "useragent", "ipaddress", "retainfor", "formreceived", "mergeperson", "processed", "logtype" ]
```
bobintetley: logtype is not a system field? It should not be in this list.

src/asm3/onlineform.py
```diff
@@ -1958,3 +1958,33 @@ def auto_remove_old_incoming_forms(dbo: Database) -> None:
...
+    logtypeid = 0
+    animalid = 0
+    animalname, dummy, dumy = get_onlineformincoming_animalperson(dbo, collationid)
+    if animalname:
+        sheltercode = animalname.split("::")[1]
+        animalid = dbo.query_int("SELECT ID FROM animal WHERE ShelterCode = '%s'" % sheltercode)
+    logcontent = []
+    fields = get_onlineformincoming_detail(dbo, collationid)
+    for f in fields:
+        if f.FIELDNAME not in SYSTEM_FIELDS:
```
bobintetley: Fair enough using SYSTEM_FIELDS to omit them from the log, but you should add logtype to system fields here rather than putting it in SYSTEM_FIELDS.
```if f.FIELDNAME != "logtype" and f.FIELDNAME not in SYSTEM_FIELDS:```

File: src/asm3/onlineform.py [hunk 5]
```diff
@@ -1958,3 +1958,33 @@ def auto_remove_old_incoming_forms(dbo: Database) -> None:
...
+            logcontent.append(f"{f.FIELDNAME}={f.VALUE}")
+        if not logtypeid and f.FIELDNAME == "logtype":
+            logtypename = f.VALUE
+            logtypeid = dbo.query_int("SELECT ID FROM logtype WHERE LogTypeName = '%s'" % logtypename)
```
bobintetley: This is wide open for a SQL injection attack!

Always use parameterised queries where you can, particularly when the value coming in has been submitted by a user. 

```dbo.query_int("SELECT ID FROM logtype WHERE LogTypeName = ?", [ logtypename ])```

File: src/asm3/onlineform.py (hunk 3 above)
Comment (adamdspencer77) [reply src/asm3/onlineform.py]: Removed.

File: src/asm3/onlineform.py (hunk 4 above)

File: src/asm3/onlineform.py (hunk 5 above)

File: src/asm3/onlineform.py [hunk 6]
```diff
@@ -1958,3 +1958,33 @@ def auto_remove_old_incoming_forms(dbo: Database) -> None:
...
         delete_onlineformincoming(dbo, "system", r.COLLATIONID)
     asm3.al.debug("removed %s incoming forms older than %s days" % (len(rows), removeafter), "onlineform.auto_remove_old_incoming_forms", dbo)
+
+def create_animal_log(dbo: Database, username: str, collationid: int):
+    logtypeid = 0
+    animalid = 0
+    animalname, dummy, dumy = get_onlineformincoming_animalperson(dbo, collationid)
+    if animalname:
+        sheltercode = animalname.split("::")[1]
+        animalid = dbo.query_int("SELECT ID FROM animal WHERE ShelterCode = ?", [sheltercode])
```
bobintetley: use the get_animal_id_from_field function instead of doing it yourself

src/static/js/options.js
```diff
@@ -898,6 +898,9 @@ $(function() {
                     ]}, 
                     { id: "tab-onlineforms", title: _("Online Forms"), fields: [
                         { id: "autoremoveforms", post_field: "AutoRemoveIncomingFormsDays", label: _("Remove incoming forms after"), type: "number", min: 1, max: 56, prelabel: "hcb", halfsize: true, xmarkup: _(" days.") }, 
+                        { id: "defaultonlineformlogtype", post_field: "OnlineFormDefaultLogType", label: _("If no logtype can be determined, use this type"), type: "select", 
+                            options: html.list_to_options(controller.logtypes, "ID", "LOGTYPENAME"), prelabel: "hcb"
+                        },
```
bobintetley, line 903: I don't think this is necessary - there's already a systemwide default log type to fall back to. If they wanted to set one for that form, there's nothing to stop them putting it in a lookup with one choice or something.

src/asm3/configuration.py
```diff
@@ -1034,6 +1035,9 @@ def default_media_notes_from_file(dbo: Database) -> bool:
 def default_nonsheltertype(dbo: Database) -> int:
     return cint(dbo, "AFNonShelterType", 40)
 
+def default_onlineformlogtype(dbo: Database) -> int:
+    return cint(dbo, "OnlineFormDefaultLogType", 3)
```
bobintetley: Not necessary, use the regular default log type - AFDefaultLogType

src/asm3/onlineform.py

src/static/js/options.js
adamdspencer77: Changed.

src/asm3/configuration.py
adamdspencer77: Done.

src/asm3/onlineform.py
```diff
@@ -1958,3 +1958,32 @@ def auto_remove_old_incoming_forms(dbo: Database) -> None:
...
+        logtypeid = asm3.configuration.default_log_type(dbo)
+    if animalid:
+        data = {
+            "type":     logtypeid,
+            "logdate":  asm3.i18n.format_date(asm3.i18n.today()),
+            "logtime":  asm3.i18n.format_time(asm3.i18n.now()),
+            "entry":    ", ".join(logcontent)
+        }
+        logpost = asm3.utils.PostedData(data, dbo.locale)
+        asm3.log.insert_log_from_form(dbo, username, asm3.log.ANIMAL, animalid, logpost)
```
bobintetley: I think you can use asm3.log.add_log instead? It has the log elements as parameters so you don't need to construct a posteddata object.

src/asm3/configuration.py
```diff
@@ -349,6 +349,7 @@
     "MoveAdoptGeneratePaperwork": "Yes",
     "MoveAdoptDonationsEnabled": "No",
     "JSWindowPrint": "Yes",
+    "OnlineFormDefaultLogType": "3",
```
bobintetley: Think this is now redundant and wants removing since we're using default log type

src/asm3/onlineform.py

src/asm3/configuration.py

bobintetley:
Needs an autoprocess option for create animal log too!

bobintetley:
logtype is missing from the list of fields at the top of onlineform.py 

strip animalname field from the logs when adding

PR #2074, opened 2026-04-29
PetFinder: Send YouTube video link

adamdspencer77:
PetFinder can now accept YouTube or Vimeo URLs in the photo1-6 fields of the data file.

People have been asking for this for a long time.

We currently allow one preferred video per animal. If that preferred video media for animal is a YouTube URL link, send it in the first empty photo slot. If there isn't one, replace photo6.

This means the PetFinder limit is now 6 pieces of media rather than 6 photos.

### DIFF
--- a/src/asm3/publishers/petfinder.py
+++ b/src/asm3/publishers/petfinder.py
@@ -409,12 +409,13 @@ class PetFinderPublisher(FTPPublisher):
         else:
             # Adoptable - include all available upto a max of 6 photos
             urls = self.getPhotoUrls(an.ID)
-            line.append(self.pfImageUrl(an.ID, urls, 0, cikeys)) # photo1
-            line.append(self.pfImageUrl(an.ID, urls, 1, cikeys)) # photo2
-            line.append(self.pfImageUrl(an.ID, urls, 2, cikeys)) # photo3
-            line.append(self.pfImageUrl(an.ID, urls, 3, cikeys)) # photo4
-            line.append(self.pfImageUrl(an.ID, urls, 4, cikeys)) # photo5
-            line.append(self.pfImageUrl(an.ID, urls, 5, cikeys)) # photo6
+            videolinkincluded = False
+            for a in range(0, 6):
+                url = self.pfImageUrl(an.ID, urls, 0, cikeys)
+                if url:
+                    line.append(self.pfImageUrl(an.ID, urls, 0, cikeys))
+                if an.WEBSITEVIDEOURL and not videolinkincluded and (not url or a == 5):
+                    line.append(an.WEBSITEVIDEOURL)
         # Arrival Date
         line.append(self.pfDate(an.MOSTRECENTENTRYDATE))
         # Birth Date
--- a/src/asm3/publishers/petfinder.py
+++ b/src/asm3/publishers/petfinder.py
@@ -418,12 +418,12 @@ class PetFinderPublisher(FTPPublisher):
             urls = self.getPhotoUrls(an.ID)
             videolinkincluded = False
             for a in range(0, 6):
-                url = self.pfImageUrl(an.ID, urls, 0, cikeys)
+                url = self.pfImageUrl(an.ID, urls, a, cikeys)
                 if an.WEBSITEVIDEOURL and not videolinkincluded and (not url or a == 5):
                     line.append(an.WEBSITEVIDEOURL)
                     videolinkincluded = True
                 else:
-                    line.append(self.pfImageUrl(an.ID, urls, 0, cikeys))
+                    line.append(url)
         # Arrival Date
         line.append(self.pfDate(an.MOSTRECENTENTRYDATE))
         # Birth Date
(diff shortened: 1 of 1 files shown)


adamdspencer77:
Not tested.

src/asm3/publishers/petfinder.py
```diff
@@ -408,13 +408,22 @@ def processAnimal(self, an: ResultRow, status: str = "A",
...
             urls = self.getPhotoUrls(an.ID)
-            line.append(self.pfImageUrl(an.ID, urls, 0, cikeys)) # photo1
-            line.append(self.pfImageUrl(an.ID, urls, 1, cikeys)) # photo2
-            line.append(self.pfImageUrl(an.ID, urls, 2, cikeys)) # photo3
-            line.append(self.pfImageUrl(an.ID, urls, 3, cikeys)) # photo4
-            line.append(self.pfImageUrl(an.ID, urls, 4, cikeys)) # photo5
-            line.append(self.pfImageUrl(an.ID, urls, 5, cikeys)) # photo6
+            videolinkincluded = False
+            for a in range(0, 6):
+                url = self.pfImageUrl(an.ID, urls, 0, cikeys)
```
bobintetley: shouldn't the 0 in this call to pfImageUrl be a?

src/asm3/publishers/petfinder.py
```diff
@@ -408,13 +408,22 @@ def processAnimal(self, an: ResultRow, status: str = "A",
...
+                if an.WEBSITEVIDEOURL and not videolinkincluded and (not url or a == 5):
+                    line.append(an.WEBSITEVIDEOURL)
+                    videolinkincluded = True
+                else:
+                    line.append(self.pfImageUrl(an.ID, urls, 0, cikeys))
```
bobintetley: you already got the url in a variable above, no need to call pfImageUrl again?

src/asm3/publishers/petfinder.py
```diff
@@ -408,13 +408,22 @@ def processAnimal(self, an: ResultRow, status: str = "A",
...
+            if not validvideourl:
+                an.WEBSITEVIDEOURL = ""
             urls = self.getPhotoUrls(an.ID)
-            line.append(self.pfImageUrl(an.ID, urls, 0, cikeys)) # photo1
-            line.append(self.pfImageUrl(an.ID, urls, 1, cikeys)) # photo2
-            line.append(self.pfImageUrl(an.ID, urls, 2, cikeys)) # photo3
-            line.append(self.pfImageUrl(an.ID, urls, 3, cikeys)) # photo4
-            line.append(self.pfImageUrl(an.ID, urls, 4, cikeys)) # photo5
-            line.append(self.pfImageUrl(an.ID, urls, 5, cikeys)) # photo6
+            videolinkincluded = False
```
bobintetley: videolinkincluded doesn't seem to be used for anything?

src/asm3/publishers/petfinder.py
adamdspencer77: Yep. Changed.

src/asm3/publishers/petfinder.py
adamdspencer77: Changed.

src/asm3/publishers/petfinder.py
adamdspencer77: It's used on line 422.

src/asm3/publishers/petfinder.py
```diff
@@ -408,13 +408,22 @@ def processAnimal(self, an: ResultRow, status: str = "A",
...
+            for a in range(0, 6):
+                url = self.pfImageUrl(an.ID, urls, a, cikeys)
+                if an.WEBSITEVIDEOURL and not videolinkincluded and (not url or a == 5):
+                    line.append(an.WEBSITEVIDEOURL)
+                    videolinkincluded = True
+                else:
+                    line.append(url)
```
bobintetley: I'm not that keen on using a loop like this, particularly when it the meat of it only happens in one situation. I think the original 5 lines the condition swapping video for photo 6 would be more readable?

```
line.append(self.pfImageUrl(an.ID, urls, 0, cikeys)) # photo1
line.append(self.pfImageUrl(an.ID, urls, 1, cikeys)) # photo2
line.append(self.pfImageUrl(an.ID, urls, 2, cikeys)) # photo3
line.append(self.pfImageUrl(an.ID, urls, 3, cikeys)) # photo4
line.append(self.pfImageUrl(an.ID, urls, 4, cikeys)) # photo5
line.append(asm3.utils.iif(validvideourl, an.WEBSITEVIDEOURL, self.pfImageUrl(an.ID, urls, 5, cikeys))) # photo 6 or preferred video
```

src/asm3/publishers/petfinder.py
```diff
@@ -408,13 +408,22 @@ def processAnimal(self, an: ResultRow, status: str = "A",
             line.append("")
         else:
             # Adoptable - include all available upto a max of 6 photos
+            validvideourl = False
+            for marker in [ "youtube.com/", "youtu.be/", "vimeo.com/"]:
+                if marker in an.WEBSITEVIDEOURL:
```
bobintetley, line 413: I didn't realise this was something they changed in python3! You used to have to use the .find() method to test for substrings, this is maybe more readable.

src/asm3/publishers/petfinder.py
adamdspencer77: Changed.

PR #2077, opened 2026-05-07
personchoosermulti widget

adamdspencer77:
We have a widget for choosing multiple animals, it would be really useful if we had a similar widget for choosing multiple people.

This widget could then be used to implement a bulk change screen for people (on a future ticket).

The widget itself would be very similar to animalchoosermulti in the page (ie. that you get a long list of the selected elements and no ability to add), but with a popup dialog very similar to personchooser. In addition to the search text at the top, it should also have a multiselect dropdown that allows you to filter people by flag. Selecting multiple flags requires that you get people who have all of the flags selected.

The person results would have checkboxes at the side to enable the selection.

### DIFF
--- a/src/main.py
+++ b/src/main.py
@@ -7104,6 +7104,14 @@ class person_embed(ASMEndpoint):
         asm3.al.debug("find '%s' got %d rows" % (self.query(), len(rows)), "main.person_embed", o.dbo)
         return asm3.utils.json(rows)
 
+    def post_multiselect(self, o):
+        self.content_type("application/json")
+        dbo = o.dbo
+        rows = asm3.person.get_person_find_simple(dbo, "", classfilter="all", limit=asm3.configuration.record_search_limit(dbo))
+        flags = asm3.lookups.get_person_flags(dbo)
+        rv = { "rows": rows, "flags": flags }
+        return asm3.utils.json(rv)
+
     def post_id(self, o):
         self.check(asm3.users.VIEW_PERSON)
         self.content_type("application/json")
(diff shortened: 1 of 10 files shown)


src/main.py
```diff
@@ -7104,6 +7104,14 @@ def post_find(self, o):
         asm3.al.debug("find '%s' got %d rows" % (self.query(), len(rows)), "main.person_embed", o.dbo)
         return asm3.utils.json(rows)
 
+    def post_multiselect(self, o):
+        self.content_type("application/json")
+        dbo = o.dbo
+        rows = asm3.person.get_person_find_simple(dbo, "", classfilter="all", limit=asm3.configuration.record_search_limit(dbo))
```
bobintetley: Unless I'm missing something, this looks like it will always return the first 1,000 person records on file and send them. Is this actually necessary? They get assigned to o.rows, but it doesn't look like they actually get used for anything since the user still has to do a search to find what they're interested in, which then replaces this set.

src/main.py
```diff
@@ -7124,14 +7132,17 @@ def post_personwarn(self, o):
...
-        p = asm3.person.get_person_embedded(dbo, pid)
-        if not p:
-            asm3.al.error("get person by id %d found no records." % pid, "main.person_embed", dbo)
-            raise web.notfound()
-        else:
-            asm3.person.embellish_adoption_warnings(dbo, p)
-            return asm3.utils.json((p,))
+        pids = o.post["ids"]
+        rows = []
+        for pid in [int(pid) for pid in pids.split(",")]:
```
bobintetley: There's a helper method that does this. You can just do ``` for pid in post.integer_list("ids"): ```

src/main.py
adamdspencer77: You're not missing anything. I made this widget by starting with the animalmulti widget and adapting. That widget starts populated so this one did originally :) Sorted.

src/main.py
adamdspencer77: Not actually needed. Restored post_personwarn to previous code.

src/static/js/common_personchoosermulti.js
```diff
@@ -0,0 +1,426 @@
...
+                    let checkbox = $(e.currentTarget);
+                    let rowid = checkbox.attr("data");
+                    $.each(o.rows, function(i, r) {
+                        if (r.ID == rowid) {
+                            let rowpresent = false;
+                            $.each(o.selectedrows, function(i, v) {
+                                if (v.ID == r.ID) {
+                                    rowpresent = true;
+                                }
+                            });
```
bobintetley: You can use common.get_row() to return a row from a set with a particular ID. 

``` let rowpresent = common.get_row(o.selectedrows, r.ID) != null; ```

File: src/static/js/common_personchoosermulti.js [hunk 4]
```diff
@@ -0,0 +1,426 @@
...
+                            if (checkbox.prop("checked") && !rowpresent) {
+                                o.selectedrows.push(r);
+                            } else if (!checkbox.prop("checked") && rowpresent) {
+                                $.each([...o.selectedrows], function(c, s) {
```
bobintetley, line 324: the spread operator is quite alien to me, I would put a comment above this line to explain that we're iterating a copy of selectedrows

src/static/js/common_personchoosermulti.js
adamdspencer77: Changed.

src/static/js/common_personchoosermulti.js
adamdspencer77: Done.

PR #2082, opened 2026-05-20
Recent incidents minimap

adamdspencer77:
Add a new option under the Animal Control for "Map of recent incidents". It should show pins for:

    Animals recently reclaimed (last 3 months)
    Owned non-shelter animals in the area (use city and last 12 months as filter)
    Recent incidents (last 3 months)

They'll have clickable links in their description box.

We're using a more upto date leaflet API now, see if there's an option to use different pin colours. I did find this on StackOverflow:
Image

it looks like we could add some pin styles to asm.css to have other coloured pins and incorporate that into the data we pass to the mapping functions in common_map.js

### DIFF
--- a/src/asm3/animalcontrol.py
+++ b/src/asm3/animalcontrol.py
@@ -244,6 +244,36 @@ def get_animalcontrol_find_advanced(dbo: Database, criteria: dict, username: str
     sql = "%s WHERE %s ORDER BY ac.ID DESC" % (get_animalcontrol_query(dbo), " AND ".join(ss.ands))
     return reduce_find_results(dbo, username, dbo.query(sql, ss.values, limit=limit, distincton="ID"))
 
+def get_recent_incidents(dbo: Database):
+    now = dbo.now()
+    sql = "SELECT ac.ID, t.IncidentName, ac.DispatchLatLong AS LatLong, ac.DispatchAddress AS Address, o.OwnerName, 0 AS IncidentType FROM " \
+    "animalcontrol ac " \
+    "LEFT JOIN owner o ON ac.OwnerID = o.ID " \
+    "LEFT JOIN incidenttype t ON ac.IncidentTypeID = t.ID " \
+    "WHERE ac.IncidentDateTime BETWEEN ? AND ? " \
+    f"UNION SELECT a.ID, {dbo.sql_concat(("a.ShelterCode", "' '", "a.AnimalName"))} AS IncidentName, o.LatLong, o.OwnerAddress AS Address, o.OwnerName, 1 AS IncidentType " \
+    "FROM animal a " \
+    "INNER JOIN adoption m ON a.ID = m.AnimalID AND m.MovementType = ? " \
+    "INNER JOIN owner o ON m.OwnerID = o.ID " \
+    "WHERE m.MovementDate BETWEEN ? AND ? " \
+    f"UNION SELECT a.ID, {dbo.sql_concat(("a.ShelterCode", "' '", "a.AnimalName"))} AS IncidentName, o.LatLong, o.OwnerAddress AS Address, o.OwnerName, 2 AS IncidentType " \
+    "FROM animal a " \
+    "INNER JOIN owner o ON a.OwnerID = o.ID " \
+    "WHERE a.NonShelterAnimal = 1 AND a.DateBroughtIn BETWEEN ? AND ? " \
+    "AND o.OwnerTown = ? "
+    return dbo.query(sql,
+        (
+            dbo.sql_date(dbo.today(offset=-90)),
+            now,
+            asm3.movement.RECLAIMED,
+            dbo.sql_date(dbo.today(offset=-90)),
+            now,
+            dbo.sql_date(dbo.today(offset=-365)),
+            now,
+            asm3.configuration.organisation_town(dbo)
+        )
+    )
+
 def reduce_find_results(dbo: Database, username: str, rows: Results) -> Results:
     """
     Given the results of a find operation, goes through the results and removes 
--- a/src/asm3/animalcontrol.py
+++ b/src/asm3/animalcontrol.py
@@ -263,8 +263,8 @@ def get_recent_incidents(dbo: Database):
     f"UNION SELECT a.ID, {dbo.sql_concat(("a.ShelterCode", "' '", "a.AnimalName"))} AS IncidentName, o.LatLong, o.OwnerAddress AS Address, o.OwnerName, 'asm-nonshelterpin' AS PinStyle " \
     "FROM animal a " \
     "INNER JOIN owner o ON a.OwnerID = o.ID " \
-    "WHERE a.NonShelterAnimal = 1 AND a.DateBroughtIn BETWEEN ? AND ? " \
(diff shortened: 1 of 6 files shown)


adamdspencer77:
Tweaked the map layout to use available viewport height.

src/asm3/animalcontrol.py
```diff
@@ -244,6 +244,36 @@ def get_animalcontrol_find_advanced(dbo: Database, criteria: dict, username: str
     sql = "%s WHERE %s ORDER BY ac.ID DESC" % (get_animalcontrol_query(dbo), " AND ".join(ss.ands))
     return reduce_find_results(dbo, username, dbo.query(sql, ss.values, limit=limit, distincton="ID"))
 
+def get_recent_incidents(dbo: Database):
+    now = dbo.now()
```
bobintetley, line 252: Add a docstring to indicate what the function is returning.

src/static/css/asm.css
```diff
@@ -1131,6 +1131,17 @@ select:disabled, input:disabled, textarea:disabled {
     width: 113px;
 }
 
+/** Leaflet pins */
+/* .asm-incidentpin {
+    filter: hue-rotate(120deg);
+} */
+.asm-nonshelterpin {
```
bobintetley, line 1138: add comments to indicate what colour this filter makes the pin for info

src/static/js/common_map.js
```diff
@@ -91,6 +91,13 @@ const mapping = {
                 if (!v.latlong || v.latlong.indexOf("0,0") == 0) { return; }
                 ll = v.latlong.split(",");
                 var marker = L.marker([ll[0], ll[1]]).addTo(map);
+                if (v.INCIDENTTYPE == 1) {
```
bobintetley: Not keen on the map library having to know anything about incidenttype.

I would have it look for a "pinstyle" element. That element could contain any extra class to add for asm-reclaimedpin, asm-nonshelterpin etc. That way the caller can influence the style and the map function doesn't need to do this. This if/else block then just becomes:

``` if (v.PINSTYLE) { marker._icon.classList.add(v.PINSTYLE); } ```

File: src/asm3/animalcontrol.py [hunk 4]
```diff
@@ -244,6 +244,36 @@ def get_animalcontrol_find_advanced(dbo: Database, criteria: dict, username: str
...
+    sql = "SELECT ac.ID, t.IncidentName, ac.DispatchLatLong AS LatLong, ac.DispatchAddress AS Address, o.OwnerName, 0 AS IncidentType FROM " \
```
bobintetley: Rather than using IncidentType in the mapping library, add a new column to the query called pinstyle to control how the pin looks and include the CSS class you want to add to the pin (see comment below)

``` 'asm-incidentpin' AS PinStyle FROM ```

File: src/asm3/animalcontrol.py [hunk 5]
```diff
@@ -244,6 +244,36 @@ def get_animalcontrol_find_advanced(dbo: Database, criteria: dict, username: str
...
+    "animalcontrol ac " \
+    "LEFT JOIN owner o ON ac.OwnerID = o.ID " \
+    "LEFT JOIN incidenttype t ON ac.IncidentTypeID = t.ID " \
+    "WHERE ac.IncidentDateTime BETWEEN ? AND ? " \
+    f"UNION SELECT a.ID, {dbo.sql_concat(("a.ShelterCode", "' '", "a.AnimalName"))} AS IncidentName, o.LatLong, o.OwnerAddress AS Address, o.OwnerName, 1 AS IncidentType " \
+    "FROM animal a " \
+    "INNER JOIN adoption m ON a.ID = m.AnimalID AND m.MovementType = ? " \
+    "INNER JOIN owner o ON m.OwnerID = o.ID " \
+    "WHERE m.MovementDate BETWEEN ? AND ? " \
+    f"UNION SELECT a.ID, {dbo.sql_concat(("a.ShelterCode", "' '", "a.AnimalName"))} AS IncidentName, o.LatLong, o.OwnerAddress AS Address, o.OwnerName, 2 AS IncidentType " \
```
bobintetley: ``` 'asm-reclaimpin' AS PinStyle ```

src/main.py
```diff
@@ -4573,7 +4573,24 @@ def controller(self, o):
...
         return {
-            "rows": rows
+            "rows": rows,
+            "name": "active_incident_map"
+        }
+
+class recent_incident_map(JSONEndpoint):
+    url = "recent_incident_map"
+    js_module = "incident_map"
+    get_permissions = asm3.users.VIEW_INCIDENT
```
bobintetley: needs VIEW_MOVEMENT as well

src/asm3/animalcontrol.py
adamdspencer77: Done.

src/static/css/asm.css
adamdspencer77: Done. Changed the pink to orange.

src/static/js/common_map.js
adamdspencer77: Changed.

src/asm3/animalcontrol.py
adamdspencer77: Done.

src/asm3/animalcontrol.py
adamdspencer77: Done.

src/main.py

Tetzor:
Review no labels showing for the pins.

src/asm3/animalcontrol.py
```diff
@@ -244,6 +244,40 @@ def get_animalcontrol_find_advanced(dbo: Database, criteria: dict, username: str
...
+    f"UNION SELECT a.ID, {dbo.sql_concat(("a.ShelterCode", "' '", "a.AnimalName"))} AS IncidentName, o.LatLong, o.OwnerAddress AS Address, o.OwnerName, 'asm-reclaimedpin' AS PinStyle " \
+    "FROM animal a " \
+    "INNER JOIN adoption m ON a.ID = m.AnimalID AND m.MovementType = ? " \
+    "INNER JOIN owner o ON m.OwnerID = o.ID " \
+    "WHERE m.MovementDate BETWEEN ? AND ? " \
+    f"UNION SELECT a.ID, {dbo.sql_concat(("a.ShelterCode", "' '", "a.AnimalName"))} AS IncidentName, o.LatLong, o.OwnerAddress AS Address, o.OwnerName, 'asm-nonshelterpin' AS PinStyle " \
+    "FROM animal a " \
+    "INNER JOIN owner o ON a.OwnerID = o.ID " \
+    "WHERE a.NonShelterAnimal = 1 AND a.DateBroughtIn BETWEEN ? AND ? "
+    # "AND o.OwnerTown = ? "
```
adamdspencer77, line 267: It took me a while to realise when testing that the organisation town field on my test DB had been reset since I last worked on this branch and this was preventing any non-shelter animals appearing on my map.

I reasoned that if a shelter has created an non-shelter animal then they must think that it's relevant and it is likely to be local anyway. I've left the previous code commented out in case you disagree.

PR #2083, opened 2026-05-21
Person records: Active fosters and Owned animals rows in the banner

adamdspencer77:
A user was asking about the last movement shown in the banner currently. They were trying to make it skip the latest if it was deceased. Digging into what they actually want, they'd like animals that are being actively fostered by that person to appear in the banner.

I suggested a row for owned animals as well, as it would be convenient when viewing people with non-shelter animals in particular.

So, rows in the banner for both, but they only appear if there is active data for them.

I think trying to add this info to get_person_query would heavily impact performance for searching and in other unnecessary areas that use person query.

If we are going to do it, we should make this a function in person.py/get_active_animals(personid: int), it should use a UNION query to return a resultset of AnimalID, ShelterCode, ShortCode, AnimalName, LinkType animals connected to this person. LinkType is an int to indicate the relationship to the person. Use 1 for adopter (ie. person is on an open adoption movement for that animal), 2 for fosterer and 3 for owner.

This function can then be called in all of the person/person_x endpoints and the data returned in the controller so that person_edit_header can access it and output the animal links in the new banner rows described above.

The two new rows can replace the last movement row in the banner. Assuming nothing else is using those fields (and it doesn't look like it), the latestmoveanimal/latestfoster columns can be removed from get_person_query.

### DIFF
--- a/src/asm3/person.py
+++ b/src/asm3/person.py
@@ -92,7 +92,6 @@ def get_person(dbo: Database, personid: int) -> ResultRow:
     """
     p = dbo.first_row( dbo.query(get_person_query(dbo) + "WHERE o.ID = ?", [personid]) )
     if p is None: return None
-    p = embellish_latest_movement(dbo, p)
     p = asm3.users.embellish_vieweditroles(dbo, "ownerrole", "OwnerID", personid, p)
     return p
 
@@ -127,27 +126,6 @@ def embellish_adoption_warnings(dbo: Database, p: ResultRow) -> ResultRow:
     for r in dbo.query("SELECT AnimalID FROM adoption WHERE OwnerID=? AND MovementType=0 AND ReservationCancelledDate Is Null", [p.ID]):
         reserves.append(str(r.ANIMALID))
     p.RESERVEDANIMALIDS = ",".join(reserves)
-    return p
-
-def embellish_latest_movement(dbo: Database, p: ResultRow) -> ResultRow:
-    """ Adds the latest movement info to a person record p and returns it.
-        The query already does this and 99% of the time it will work fine and makes these columns available
-        in v_person for the query builder. BUT if we have a data import where the movements were created out of
-        order, MAX(ID) will fail and return the wrong movement. """
-    if p is None: return p
-    lm = dbo.first_row(dbo.query("SELECT m.ID AS LatestMoveAnimalID, a.ID AS LatestMoveAnimalID, a.AnimalName AS LatestMoveAnimalName, " \
-        "a.ShelterCode AS LatestMoveShelterCode, a.DeceasedDate AS LatestMoveDeceasedDate, mt.MovementType AS LatestMoveTypeName " \
-        "FROM adoption m "
-        "INNER JOIN animal a ON m.AnimalID = a.ID " \
-        "INNER JOIN lksmovementtype mt ON mt.ID = m.MovementType " \
-        "WHERE m.MovementType > 0 AND m.OwnerID = ? AND (ReturnDate Is Null OR ReturnDate > ?)" \
-        "ORDER BY m.MovementDate DESC", [p.ID, dbo.today()]))
-    if lm is not None:
-        p.LATESTMOVEANIMALID = lm.LATESTMOVEANIMALID
-        p.LATESTMOVEANIMALNAME = lm.LATESTMOVEANIMALNAME
-        p.LATESTMOVESHELTERCODE = lm.LATESTMOVESHELTERCODE
-        p.LATESTMOVEDECEASEDDATE = lm.LATESTMOVEDECEASEDDATE
-        p.LATESTMOVETYPENAME = lm.LATESTMOVETYPENAME
     return p
 
 def get_homechecked(dbo: Database, personid: int) -> Results:
--- a/src/asm3/person.py
+++ b/src/asm3/person.py
@@ -23,7 +23,7 @@ from datetime import datetime
 ASCENDING = 0
 DESCENDING = 1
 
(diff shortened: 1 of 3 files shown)


adamdspencer77:
I removed person.embellish_latest_movement, I didn't appear to be used anywhere except in person headers and was no longer needed.

src/static/js/header_edit_header.js
```diff
@@ -588,12 +588,44 @@ edit_header = {
...
+                fosterlinks.push('<b><a href="animal?id=' + v.ANIMALID + '">' + v.ANIMALNAME + '</a></b>');
+            });
+            fostershtml += fosterlinks.join(", ");
+            fostershtml += '</td></tr>';
+        }
+        if (ownedanimals.length) {
+            ownedanimalshtml = '<tr><td>' + _("Owned Animals") + ':</td>';
+        }
+        if (ownedanimals.length > 5) {
+            ownedanimalshtml += '<td><b><a href="animal?id=' + ownedanimals[0].ANIMALID + '">' + ownedanimals[0].ANIMALNAME + '</a> plus <a href="person_movements?id=' + controller.person.ID + '">' + (ownedanimals.length - 1) + ' more...</a></b></td></tr>';
```
bobintetley: "more ..." should be translatable

src/static/js/header_edit_header.js
```diff
@@ -588,12 +588,44 @@ edit_header = {
...
+                ownedanimals.push(v);
+            }
+        });
+        let fostershtml = "";
+        let ownedanimalshtml = "";
+        if (fosters.length) {
+            fostershtml = '<tr><td>' + _("Active Fosters") + ':</td>';
+        }
+        if (fosters.length > 5) {
+            fostershtml += '<td><b><a href="animal?id=' + fosters[0].ANIMALID + '">' + fosters[0].ANIMALNAME + '</a> plus <a href="person_movements?id=' + controller.person.ID + '">' + (fosters.length - 1) + ' more...</a></b></td></tr>';
```
bobintetley: another translation for more ...

src/asm3/person.py
```diff
@@ -23,6 +23,20 @@
 ASCENDING = 0
 DESCENDING = 1
 
+def get_active_animals(dbo: Database, personid: int):
```
bobintetley: Minor thing, I think "get_owned_animals" would be a slightly more accurate function name :)

src/static/js/header_edit_header.js
adamdspencer77: Changed.

src/static/js/header_edit_header.js
adamdspencer77: Changed.

src/asm3/person.py

src/static/js/header_edit_header.js
```diff
@@ -588,12 +588,44 @@ edit_header = {
...
+                ownedanimals.push(v);
+            }
+        });
+        let fostershtml = "";
+        let ownedanimalshtml = "";
+        if (fosters.length) {
+            fostershtml = '<tr><td>' + _("Active Fosters") + ':</td>';
+        }
+        if (fosters.length > 5) {
+            fostershtml += '<td><b><a href="animal?id=' + fosters[0].ANIMALID + '">' + fosters[0].ANIMALNAME + '</a> plus <a href="person_movements?id=' + controller.person.ID + '">' + (fosters.length - 1) + ' ' + _("more...") + '</a></b></td></tr>';
```
bobintetley: The text string should be tokenised so that it can be translated in other languages. I also would not make the animal's name a link to the animal, just make the whole thing a link to the movement tab.

```
fostershtml += '<td><a href="person_movements?id=' + controller.person.ID + '">' + _("{0} plus {1} more...").replace("{0}", fosters[0].ANIMALNAME).replace("{1}",  (fosters.length - 1)) + '</a></td></tr>';
```

src/static/js/header_edit_header.js
```diff
@@ -588,12 +588,44 @@ edit_header = {
...
+                fosterlinks.push('<b><a href="animal?id=' + v.ANIMALID + '">' + v.ANIMALNAME + '</a></b>');
+            });
+            fostershtml += fosterlinks.join(", ");
+            fostershtml += '</td></tr>';
+        }
+        if (ownedanimals.length) {
+            ownedanimalshtml = '<tr><td>' + _("Owned Animals") + ':</td>';
+        }
+        if (ownedanimals.length > 5) {
+            ownedanimalshtml += '<td><b><a href="animal?id=' + ownedanimals[0].ANIMALID + '">' + ownedanimals[0].ANIMALNAME + '</a> plus <a href="person_movements?id=' + controller.person.ID + '">' + (ownedanimals.length - 1) + ' ' + _("more...") + '</a></b></td></tr>';
```
bobintetley: Same as above

src/static/js/header_edit_header.js
adamdspencer77: Changed.

adamdspencer77: Realised that translatable string needs to contain HTML so the the link the the most recent animal and the link the the movement tab go to their respective targets so tweaked.

src/static/js/header_edit_header.js
adamdspencer77: Kind of done, as above.

src/static/js/header_edit_header.js
adamdspencer77: Put back.

src/static/js/header_edit_header.js
adamdspencer77: Put back.

Issue #2069, opened 2026-04-28
personchoosermulti widget


bobintetley:
We have a widget for choosing multiple animals, it would be really useful if we had a similar widget for choosing multiple people.

This widget could then be used to implement a bulk change screen for people (on a future ticket).

The widget itself would be very similar to animalchoosermulti in the page (ie. that you get a long list of the selected elements and no ability to add), but with a popup dialog very similar to personchooser. In addition to the search text at the top, it should also have a multiselect dropdown that allows you to filter people by flag. Selecting multiple flags requires that you get people who have all of the flags selected.

The person results would have checkboxes at the side to enable the selection.

bobintetley:
Use the person_embed endpoint for doing the search and getting the results for the widget.

PR #2086, opened 2026-05-28
Bulk change person screen

adamdspencer77:
Add a bulk change person screen. This is dependent on https://github.com/sheltermanager/asm3/issues/2069 being done first.

It's up in the air what fields make the most sense to bulk change on people. Adding or removing flags is a definite, but consider others.


src/asm3/person.py
```diff
@@ -23,6 +23,32 @@
...
+            delete_person(dbo, username, personid, remove_movements=True)
+            deleted.append(personid)
+        except asm3.utils.ASMValidationError as error:
+            skippedids.append(personid)
+            skippeddict[personid] = error.msg
+            asm3.utils.web_context().status = "200 OK"
+    if len(skippedids):
+        skippedids = tuple(skippedids)
+        skipped = dbo.query(
+            f"SELECT ID, OwnerName FROM owner WHERE ID IN {str(skippedids)}"
```
adamdspencer77: Wanted to discuss this query. I tried to parameterise this but couldn't find a way and my solution seems scrappy.

src/static/js/common_tableform.js
```diff
@@ -2399,14 +2399,15 @@ const tableform = {
         // If this dialog has already been created, destroy it first
         common.widget_destroy(selector, "dialog", false);
 
-        if (!o) { o = { notblank: [], notzero: [], okclick: null, redbutton: null, redclick: null, cancelclick: null }; }
+        if (!o) { o = { notblank: [], notzero: [], okclick: null, redbutton: null, redclick: null, cancelclick: null, width: "auto" }; }
```
adamdspencer77: I made this change to allow my dialog a little extra width. I don't think that this change should affect anything already using this function.

src/asm3/person.py
bobintetley: Use dbo.sql_placeholders function. It outputs enough ? for each element in a given list.

```
skipped = dbo.query(f"SELECT ID, OwnerName FROM owner WHERE ID IN ({dbo.sql_placeholders(skippedids)})", skippedids)
```

src/static/js/common_tableform.js
bobintetley: I think it will break everywhere that was already manually setting a width for the dialog!

If you want your dialog wider, you can set width in the object that your caller is passing to show_okcancel_dialog instead rather than making it auto for every caller ever.

src/asm3/person.py
adamdspencer77: Cheers. Sorted.

src/static/js/common_tableform.js
adamdspencer77: Hopefully sorted.

PR #2093, opened 2026-06-01
DOA not marked as DiedOffShelter

adamdspencer77:
Animals booked in as DOA are not currently marked as DiedOffShelter. Make changes so that they are.


src/asm3/animal.py
```diff
@@ -3338,12 +3338,14 @@ def insert_animal_from_form(dbo: Database, post: PostedData, username: str) -> i
...
     # Handle deceased date and doa being set via entry type
+    diedoffshelter = 0
     deceaseddate = post.date("deceaseddate")
     deadonarrival = post.integer("deadonarrival")
     deathcategory = post.integer("deathcategory")
     if post.integer("entrytype") == 9:
         deceaseddate = dbo.today()
         deadonarrival = 1
+        diedoffshelter = 1
         deathcategory = asm3.configuration.default_death_reason(dbo)
```
bobintetley: I don't think we need to update diedoffshelter at all here, so both these lines could be removed?

When we get to the end of insert_animal_from_form (this function), it calls update_animal_status which you've now fixed to check isdoa and set diedoffshelter accordingly.

src/asm3/animal.py
```diff
@@ -5626,6 +5628,10 @@ def b2i(x):
     # Non-shelter animals who are deceased have by definition died off the shelter
     if a.nonshelteranimal == 1 and a.deceaseddate:
         diedoffshelter = True
+    
+    # DOA animals have by definition died off the shelter
+    if a.isdoa == 1:
```
bobintetley, line 5631: While this code/check is perfectly fine, you need to change get_animal_status_query to include the ISDOA column or it won't be present when a bunch of functions call update_animal_status and this line will cause an exception.

src/asm3/animal.py

src/asm3/animal.py
adamdspencer77: Done.

bobintetley:
Coming back to this with fresh eyes, why did we need to set diedoffshelter for DOA? DOA is currently an entry category and has a boolean column in the DB, any reports and figures that count deaths can already separate DOA from other types of death, can you remember what was driving the need to set diedoffshelter?

Perhaps the naming of the column has caused us confusion - diedoffshelter doesn't literally mean died off the shelter, it means "this death should not be counted in figures"

bobintetley:
Closing for now, I'm not sure this makes sense any more.

PR #2096, opened 2026-06-04
Automatic email filing

adamdspencer77:
This is probably an smcom specific feature, but backed by code that open source users could use if necessary.

Allow users to copy in an email address when they are replying to emails they receive from adopters, volunteers, etc. This email address should be something like accountnumber/[alias@filing.sheltermanager.com](mailto:alias@filing.sheltermanager.com)

This box will call a cron task for that account that finds the person based on the TO email address and auto files the message in the log. Will need quite a few specifics figuring out.


src/asm3/service.py
```diff
@@ -971,6 +972,20 @@ def handler(post: PostedData, path: str, remoteip: str, referer: str, useragent:
         asm3.media.attach_file_from_form(dbo, username, asm3.media.ANIMAL, int(animalid), asm3.media.MEDIASOURCE_SERVICEUPLOAD, post)
         return ("text/plain", 0, 0, "OK")
 
+    elif method == "email_received_log":
+        if not asm3.configuration.email_log(dbo):
+            return ("text/plain", 0, 0, "ERROR - Email logging disabled")
+        emailaddress = post["emailaddress"]
+        logtypeid = asm3.configuration.email_log_type(dbo)
+        logtext = post["logtext"]
+        linktype = 1
```
bobintetley: use the constant for readability, presume this is log.PERSON

src/asm3/sitedefs.py
```diff
@@ -413,3 +413,4 @@ def get_dict(k: str, dv: Dict = {}) -> Dict:
 VIDEO_ENABLED = get_boolean("video_enabled", False)
 VIDEO_SIZE_LIMIT = get_integer("video_size_limit", 10485760)
 VIDEO_THUMBNAIL_CMD = get_string("video_thumbnail_cmd", "ffmpeg -v 0 -y -an -i %(input)s -ss 00:00:2.000 -vframes 1 %(output)s")
+EMAIL_SERVER_IP_WHITELIST = get_string("email_server_ip_whitelist", "")
```
bobintetley: Put this up near the SMTP_SERVER value in order to keep mail related config together.

I'd put a comment near it to indicate what this config value is for. Eg: "Whitelist of mail server IPs that are allowed to call the email_received_log method of the service API"

If you make the default value "127.0.0.1 localhost" then it also indicates that it can be IP addresses or hostnames and they should be whitespace separated.

src/asm3/smcom.py
```diff
@@ -184,6 +184,9 @@ def vacuum_full(dbo: Database) -> None:
     """ Performs a full vacuum on the database via command line (transaction problems via db.py) """
     os.system("psql -U %s -c \"VACUUM FULL;\"" % dbo.database)
 
+def validate_mail_server_ip(ipaddress):
+    return (ipaddress in EMAIL_SERVER_IP_WHITELIST)
```
bobintetley: Using in against a str like this will be doing a partial match, which in certain edge cases could produce unpredictable results. Split the string first.

Eg: return ipaddress in EMAIL_SERVER_IP_WHITELIST.split()

src/asm3/service.py
adamdspencer77: Done.

src/asm3/sitedefs.py

src/asm3/smcom.py
adamdspencer77: Done.

src/asm3/service.py
```diff
@@ -971,6 +972,20 @@ def handler(post: PostedData, path: str, remoteip: str, referer: str, useragent:
         asm3.media.attach_file_from_form(dbo, username, asm3.media.ANIMAL, int(animalid), asm3.media.MEDIASOURCE_SERVICEUPLOAD, post)
         return ("text/plain", 0, 0, "OK")
 
+    elif method == "email_received_log":
```
bobintetley, line 976: Add a unit test for this service method, should hopefully be able to call service.handler directly and include everything necessary in post data.

adamdspencer77: Done.

PR #2098, opened 2026-06-05
Online Forms: Limit how often the same email address can submit a form

adamdspencer77:
Shelters often have issues with the same people spamming them with multiple adoption applications over and over again.

Add a new integer column to onlineform for EmailSubmissionLimitDays

This column determines how many submissions we will accept from the same email address within X days. If the value is 0 then there is no limit (the default behaviour and what will be applied to existing forms).

To make this work, when we receive a form submission, we will go through the following process:

    If the EmailSubmissionLimitDays value for this form is 0, do nothing.
    Check if there's an emailaddress field in the form and it has a valid email in it, if there isn't do nothing.
    Check the disk cache for an entry named formemail_{formid}_{emailaddress}. This cache entry should contain the last date and time we received a form submission from that email address.
    If we have a value, add the EmailSubmissionLimitDays to it in days and see if we are within the limit. If we are, raise an exception (ASMValidationError).
    If we're not within the limit, accept the form submission as normal
    Set the disk cache entry for formemail_{formid}_{emailaddress} to dbo.now()

I would put this in a separate function in onlineform.py called check_submission_limit that can be called from insert_onlineformincoming_from_form


src/asm3/configuration.py
```diff
@@ -1171,6 +1171,12 @@ def email_licence_reminder(dbo: Database) -> bool:
 def email_licence_reminder_days(dbo: Database) -> int:
     return cint(dbo, "EmailLicenceReminderDays", DEFAULTS["EmailLicenceReminderDays"])
 
+def email_log(dbo: Database) -> bool:
+    return cboolean(dbo, "LogEmailByDefault")
+
+def email_log_type(dbo: Database) -> int:
+    return cint(dbo, "EmailLogType", DEFAULTS["EmailLogType"])
+
```
bobintetley: These are from branch 952 and will cause conflicts during merge :)

src/asm3/onlineform.py
```diff
@@ -195,6 +200,7 @@ def get_onlineform_html(dbo: Database, formid: int, completedocument: bool = Tru
     h.append('<input type="hidden" name="flags" value="%s" />' % form.SETOWNERFLAGS)
     h.append('<input type="hidden" name="mediaflags" value="%s" />' % form.SETMEDIAFLAGS)
     h.append('<input type="hidden" name="formname" value="%s" />' % asm3.html.escape(form.NAME))
+    h.append('<input type="hidden" name="emailsubmissionlimit" value="%s" />' % form.EMAILSUBMISSIONLIMITDAYS)
```
bobintetley: I don't think this needs to be here - passing it through the client would let a malicious actor change it. There are quite a few form fields that are just looked up from the name and this should be one of them.

src/asm3/onlineform.py
```diff
@@ -95,7 +96,7 @@
 SPAMBOT_TXT = 'a_emailaddress'
 
 # Fields that are added to forms by the system and are not user enterable
-SYSTEM_FIELDS = [ "useragent", "ipaddress", "retainfor", "formreceived", "mergeperson", "processed" ]
+SYSTEM_FIELDS = [ "useragent", "ipaddress", "retainfor", "formreceived", "mergeperson", "processed", "emailsubmissionlimitdays" ]
```
bobintetley: Shouldn't be necessary to make this a form field at all.

src/asm3/onlineform.py
```diff
@@ -888,10 +896,28 @@ def insert_onlineformincoming_from_form(dbo: Database, post: PostedData, remotei
...
         if k.startswith("lastname") or k.startswith("surname"): lastname = v
         if k.startswith("zipcode") or k.startswith("postcode"): postcode = v
+        if k.startswith("emailaddress"): emailaddress = v
+        if k == "emailsubmissionlimit": emailsubmissionlimit = int(v)
+
+    if emailsubmissionlimit:
+        if not asm3.utils.is_valid_email_address(emailaddress):
+            pass
+        else:
+            formid = dbo.query_int("SELECT ID FROM onlineform WHERE Name = ?", [post["formname"]])
```
bobintetley: You could look up emailsubmissionlimitdays on the onlineform record here instead of passing it through the client. Use ```dbo.first_row(dbo.query("SELECT ID, EmailSubmissionLimitDays FROM onlineform ...```

src/asm3/onlineform.py
```diff
@@ -888,10 +896,28 @@ def insert_onlineformincoming_from_form(dbo: Database, post: PostedData, remotei
...
+    emailaddress = ""
+    emailsubmissionlimit = 0
     for k, v in post.data.items():
         if k.startswith("firstname") or k.startswith("forenames"): firstname = v
         if k.startswith("lastname") or k.startswith("surname"): lastname = v
         if k.startswith("zipcode") or k.startswith("postcode"): postcode = v
+        if k.startswith("emailaddress"): emailaddress = v
+        if k == "emailsubmissionlimit": emailsubmissionlimit = int(v)
+
+    if emailsubmissionlimit:
```
bobintetley: Check the submission limit after we've done the spam checks instead of first

src/asm3/service.py
```diff
@@ -20,19 +20,21 @@
...
 import asm3.publishers.html
 import asm3.reports
+import asm3.smcom
 import asm3.users
 import asm3.utils
 from asm3.i18n import _, now, add_seconds, format_currency, format_time, python2display, subtract_seconds
 from asm3.sitedefs import BOOTSTRAP_JS, BOOTSTRAP_CSS, BOOTSTRAP_ICONS_CSS
 from asm3.sitedefs import JQUERY_JS, JQUERY_UI_JS, SIGNATURE_JS, JQUERY_UI_CSS, MOUSETRAP_JS
-from asm3.sitedefs import BASE_URL, SERVICE_URL, MULTIPLE_DATABASES, CACHE_SERVICE_RESPONSES, IMAGE_HOTLINKING_ONLY_FROM_DOMAIN
+from asm3.sitedefs import BASE_URL, SERVICE_URL, MULTIPLE_DATABASES, CACHE_SERVICE_RESPONSES, IMAGE_HOTLINKING_ONLY_FROM_DOMAIN, EMAIL_SERVER_IP_WHITELIST
```
bobintetley: from branch 952

src/asm3/service.py
```diff
@@ -970,6 +972,20 @@ def handler(post: PostedData, path: str, remoteip: str, referer: str, useragent:
...
+        emailaddress = post["emailaddress"]
+        logtypeid = asm3.configuration.email_log_type(dbo)
+        logtext = post["logtext"]
+        linktype = 1
+        pids = asm3.person.get_person_ids_for_email(dbo, emailaddress)
+        if not asm3.smcom.validate_mail_server_ip(remoteip):
+            return ("text/plain", 0, 0, "ERROR - Invalid IP Address")
+        for pid in pids:
+            asm3.log.add_log(dbo, "system", linktype, pid, logtypeid, logtext)
+        return ("text/plain", 0, 0, "OK")
```
bobintetley: This is from branch 952 as well

src/asm3/smcom.py
```diff
@@ -3,7 +3,7 @@
 import asm3.cachedisk
 import asm3.db
 
-from asm3.sitedefs import MULTIPLE_DATABASES, MULTIPLE_DATABASES_TYPE
+from asm3.sitedefs import MULTIPLE_DATABASES, MULTIPLE_DATABASES_TYPE, EMAIL_SERVER_IP_WHITELIST
```
bobintetley: branch 952

src/asm3/smcom.py
```diff
@@ -184,6 +184,9 @@ def vacuum_full(dbo: Database) -> None:
     """ Performs a full vacuum on the database via command line (transaction problems via db.py) """
     os.system("psql -U %s -c \"VACUUM FULL;\"" % dbo.database)
 
+def validate_mail_server_ip(ipaddress):
+    return (ipaddress in EMAIL_SERVER_IP_WHITELIST)
```
bobintetley: branch 952

src/static/js/onlineforms.js
```diff
@@ -64,6 +64,10 @@ $(function() {
                         type: "select", classes: "asm-doubleselectbox",
                         callout: _("Process submissions of this form automatically and bypass the incoming forms queue"),
                         options: { displayfield: "NAME", valuefield: "ID", rows: onlineforms.auto_process_options } },
+                    { json_field: "EMAILSUBMISSIONLIMITDAYS", post_field: "emailsubmissionlimitdays", label: _("Email submission limit (days)"),
+                        type: "number", min: 0, max: 30, validation: "notblank",
+                        callout: "Do not allow multiple submissions of this form the same email address within this number of days (0 removes any limit)"
```
bobintetley: missing the word "from" - Do not allow multiple submissions of this form from the same email address

src/asm3/configuration.py
adamdspencer77: Removed :)

src/asm3/onlineform.py
adamdspencer77: Removed.

src/asm3/onlineform.py

src/asm3/onlineform.py
adamdspencer77: Moved.

src/asm3/service.py
adamdspencer77: Removed.

src/asm3/service.py
adamdspencer77: Removed.

src/asm3/smcom.py
adamdspencer77: Removed.

src/asm3/smcom.py
adamdspencer77: Removed.

src/static/js/onlineforms.js

src/asm3/onlineform.py
adamdspencer77: Removed. Also removed retainfor from this section and reworked attach_form to use the formname to get the retainfor value.

src/asm3/configuration.py
```diff
@@ -1171,6 +1171,9 @@ def email_licence_reminder(dbo: Database) -> bool:
 def email_licence_reminder_days(dbo: Database) -> int:
     return cint(dbo, "EmailLicenceReminderDays", DEFAULTS["EmailLicenceReminderDays"])
 
+def email_log(dbo: Database) -> bool:
+    return cboolean(dbo, "LogEmailByDefault")
```
bobintetley: Code is fine, but it's not in alphabetical order with the rest of the functions!

src/asm3/onlineform.py
```diff
@@ -969,7 +977,24 @@ def insert_onlineformincoming_from_form(dbo: Database, post: PostedData, remotei
...
+            dbo.query("SELECT ID, EmailSubmissionLimitDays FROM onlineform WHERE Name = ?", 
+            [post["formname"]]
+            )
+        )
+        formid = formdata.ID
+        key = f"formemail_{str(formid)}_{emailaddress}"
+        if asm3.cachedisk.exists(key, dbo.name()):
+            lastreceived = asm3.cachedisk.get(key, dbo.name())
+            if lastreceived and (dbo.now() - lastreceived).days < formdata.EMAILSUBMISSIONLIMITDAYS:
+                raise asm3.utils.ASMValidationError("Form already submitted recently from %s" % emailaddress)
```
bobintetley: Since the user will see this error message, we should make it translatable, and also include how many days we expect them to wait between attempts. Use the {0} tokens again.

``` _("Form recently submitted by {0}, please wait {1} days between submissions.", l).format(emailaddress, formdata.EMAILSUBMISSIONLIMITDAYS) ```

File: src/asm3/sitedefs.py [hunk 13]
```diff
@@ -413,3 +413,5 @@ def get_dict(k: str, dv: Dict = {}) -> Dict:
 VIDEO_ENABLED = get_boolean("video_enabled", False)
 VIDEO_SIZE_LIMIT = get_integer("video_size_limit", 10485760)
 VIDEO_THUMBNAIL_CMD = get_string("video_thumbnail_cmd", "ffmpeg -v 0 -y -an -i %(input)s -ss 00:00:2.000 -vframes 1 %(output)s")
+
+EMAIL_SERVER_IP_WHITELIST = get_string("email_server_ip_whitelist", "")
```
bobintetley: I think this was part of another branch?

src/asm3/configuration.py
adamdspencer77: Moved.

src/asm3/onlineform.py

src/asm3/sitedefs.py
adamdspencer77: Removed.

src/asm3/onlineform.py
```diff
@@ -969,7 +977,26 @@ def insert_onlineformincoming_from_form(dbo: Database, post: PostedData, remotei
...
+    if asm3.utils.is_valid_email_address(emailaddress):
+        # Check if a limit has been set for days between submissions for this form
+        formdata = dbo.first_row(
+            dbo.query("SELECT ID, EmailSubmissionLimitDays FROM onlineform WHERE Name = ?", 
+            [post["formname"]]
+            )
+        )
+        formid = formdata.ID
+        key = f"formemail_{str(formid)}_{emailaddress}"
+        if asm3.cachedisk.exists(key, dbo.name()):
```
bobintetley: asm3.cachedisk.get() returns None if the value doesn't exist, so you can just call that with having to test exists first

src/asm3/onlineform.py
```diff
@@ -969,7 +977,26 @@ def insert_onlineformincoming_from_form(dbo: Database, post: PostedData, remotei
...
+            lastreceived = asm3.cachedisk.get(key, dbo.name())
+            if lastreceived and (dbo.now() - lastreceived).days < formdata.EMAILSUBMISSIONLIMITDAYS:
```
bobintetley: If this condition fails, execution will fall out of the conditions and do nothing, when I think we should be updating the cache entry?

src/static/js/onlineforms.js
```diff
@@ -64,6 +64,10 @@ $(function() {
                         type: "select", classes: "asm-doubleselectbox",
                         callout: _("Process submissions of this form automatically and bypass the incoming forms queue"),
                         options: { displayfield: "NAME", valuefield: "ID", rows: onlineforms.auto_process_options } },
+                    { json_field: "EMAILSUBMISSIONLIMITDAYS", post_field: "emailsubmissionlimitdays", label: _("Email submission limit (days)"),
+                        type: "number", min: 0, max: 30, validation: "notblank",
+                        callout: "Do not allow multiple submissions of this form from the same email address within this number of days (0 removes any limit)"
```
bobintetley: This string should be translatable (missing _(""))

src/asm3/onlineform.py
adamdspencer77: Don't I need the if so that if it doesn't exist, it gets created in the else?

adamdspencer77: Hopefully sorted now.

src/asm3/onlineform.py
adamdspencer77: Hopefully sorted.

src/static/js/onlineforms.js

PR #2100, opened 2026-06-09
Add additional fields for animals in adoption events.

saggineumann:
Introduce the Animal in Event link type for eventanimal records with admin support, configurable list columns, and edit dialog integration. Includes DB update 35119, a maint_db_update fix for runs without a logged-in session, Hebrew translations, and tests.
(I also changed the hardcoded Make test's 5000 port to 5001 because 5000 is in use on Macs).


src/asm3/dbupdates/35123.py
```diff
@@ -0,0 +1,3 @@
+from asm3.dbupdate import execute
```
bobintetley, line 1: This file 35119 already exists in another pull request - rename to 35123 please

Makefile
```diff
@@ -138,7 +138,7 @@ smcomreports:
 
 test: version
 	@echo "[test] ========================="
-	cd src && python3 main.py 5000
+	cd src && python3 main.py 5001
```
bobintetley, line 141: Please undo, there's far too many internal processes here that rely on port 5000 and we don't use Macs. What service is Mac running on 5000 anyway?

src/asm3/additional.py
```diff
@@ -31,6 +31,7 @@
 MOVEMENT_RELEASED = 28
 MOVEMENT_RETAILER = 29
 MOVEMENT_RESERVATION = 30
+EVENT_ANIMAL = 32
```
bobintetley: Move up so it's below EVENT

bobintetley (commented):
Thanks, couple of minor things before merge

Makefile
saggineumann: Airplay receiver 🤪

src/asm3/dbupdates/35123.py
saggineumann: renamed

src/asm3/additional.py
saggineumann: moved

Makefile
saggineumann: reverted

saggineumann:
@bobintetley this should be good to go now 🤞

PR #2103, opened 2026-06-11
AreaLatLong column for animallost and animalfound

adamdspencer77:
If we had an arealatlong column on the lost and found tables, we could populate it by getting a geocode from the AreaLost/Found and AreaPostcode fields.

This would allow us to show lost and found maps based on where animals are actually lost and found rather than the address of the person that lost/found them.

Add a third column to the details tab on lost/found screens to show a minimap based on this geocode (same as person and dispatch). As with other screens, it should be possible to move the pin, etc.

Also added an interactive lost and found animal map.


src/asm3/animalcontrol.py
```diff
@@ -554,7 +520,9 @@ def update_dispatch_geocode(dbo: Database, incidentid: int, latlon: str = "", ad
     # If someone has deleted the values, a latlon of ,,HASH is returned so
     # we allow the geocode to be regenerated in that case.
     if asm3.configuration.show_lat_long(dbo) and latlon is not None and latlon != "" and not latlon.startswith(",,"):
-        return latlon
+        # Has the address changed? If so do nothing
```
adamdspencer77, line 568: Put in so that latlong updates when addresses are changed even when they are hidden.

src/asm3/person.py
```diff
@@ -1697,7 +1697,9 @@ def update_geocode(dbo: Database, personid: int, latlon: str = "", address: str
     # If someone has deleted the values, a latlon of ,,HASH is returned so
     # we allow the geocode to be regenerated in that case.
     if asm3.configuration.show_lat_long(dbo) and latlon is not None and latlon != "" and not latlon.startswith(",,"):
-        return latlon
+        # Has the address changed? If so do nothing
```
adamdspencer77, line 1727: Put in so that latlong updates when addresses are changed even when they are hidden.

src/static/js/common_map.js
```diff
@@ -16,6 +16,7 @@ const mapping = {
      * latlong: A lat,long string to mark the center of the map (or empty string for current location)
      * markers: A list of marker objects to draw { latlong: "", popuptext: "", popupactive: false }
      */
+    markers: [],
```
adamdspencer77: Moved markers array to here so it can be accessed from outside to manipulate markers later.

src/static/js/common_map.js
```diff
@@ -78,27 +79,37 @@ const mapping = {
         $("head").append('<link rel="stylesheet" href="' + asm.leafletcss + '" />');
         mapping._get_script(asm.leafletjs, function() {
             var ll = latlong.split(",");
-            var map = L.map(divid).setView([ll[0], ll[1]], 15);
+            mapping.map = L.map(divid).setView([ll[0], ll[1]], 15);
```
adamdspencer77, line 192: Made map available from outside to allow restyling of markers in response to changing filters.

src/asm3/dbupdates/35124.py
```diff
@@ -0,0 +1,7 @@
+from asm3.dbupdate import add_column, execute
+
+add_column(dbo, "animallost", "LatLong", dbo.type_shorttext)
+execute(dbo, "UPDATE animallost SET LatLong = ''")
```
bobintetley: AreaLatLong rather LatLong :(

src/static/js/common_map.js
bobintetley: that's fine, but it should be down below nearer leaflet, have its own comments to indicate what it's for and the name should indicate that it belongs to the leaflet mapping functions rather than google and it should indicate private with an underscore prefix. Eg: ``` _leaflet_markers: [] ```

src/static/js/common_map.js
bobintetley: It would be better for encapsulation purposes to have functions that can be called from outside to manipulate the map, rather than exposing the map itself since this object is leaflet specific. Besides technically supporting Google as well, we might want to support other client side map libraries in future.

src/asm3/dbupdates/35124.py
adamdspencer77: Changed.

src/asm3/animal.py
```diff
@@ -3138,6 +3138,27 @@ def get_recent_changes(dbo: Database, months: int = 1, include_additional_fields
...
+        "FROM animal a " \
+        "INNER JOIN species s ON a.SpeciesID = s.ID " \
+        "INNER JOIN owner o ON a.OwnerID = o.ID " \
+        "WHERE a.NonShelterAnimal = 1 AND a.CreatedDate >= ?",
+        [floor]
+    )
+
+def get_recent_reclaimed_animals(dbo: Database, floor: datetime):
+    return dbo.query(
+        "SELECT a.ID, a.ShelterCode, a.ShortCode, a.AnimalName, a.SpeciesID, s.SpeciesName, o.LatLong, o.OwnerAddress, m.MovementDate " \
```
bobintetley, line 3160: Should be using SELECT DISTINCT since there's the capacity here for duplicate animals to be returned if they were reclaimed multiple times in the same period.

src/asm3/lostfound.py
```diff
@@ -665,7 +666,36 @@ def get_found_person_name(dbo: Database, aid: int) -> str:
...
+    fromdate = dbo.sql_date(dbo.today(offset=offset))
+    return dbo.query(
+        "SELECT al.ID, 'lost' AS LostOrFound, DateLost, s.SpeciesName, s.ID AS SpeciesID, al.AreaLatLong, al.AreaLost, 'asm-lostanimalpin' AS PinStyle FROM animallost al INNER JOIN species s ON al.AnimalTypeID = s.ID WHERE al.DateLost >= ? " \
+        "UNION SELECT af.ID, 'found' AS LostOrFound, DateFound AS LFDate, s.SpeciesName, s.ID AS SpeciesID, af.AreaLatLong, af.AreaFound AS Area, 'asm-foundanimalpin' AS PinStyle FROM animalfound af INNER JOIN species s ON af.AnimalTypeID = s.ID WHERE af.DateFound >= ? ",
+        [fromdate, fromdate]
+    )
+
+def get_recent_found_animals(dbo: Database, floor: datetime) -> Results:
+    """
+    Returns rows of unresolved found animal data that were found on or after floor
```
bobintetley, line 683: For consistency with other functions, they usually call this date "fromdate" or sometimes "cutoff" rather than floor. Very minor thing :)

src/static/js/common_map.js
```diff
@@ -30,6 +31,10 @@ const mapping = {
         if (latlong != "") {
             _draw_map(latlong);
         }
+        // No https connection - assuming test environment, providing fallback location
+        else if (!document.location.href.startsWith("https://")) {
+            _draw_map("38.536777,-90.4471,415AxminsterDriveFentonMO63026");
+        }
```
bobintetley, line 70: I think this might break for open source users who aren't hosting over SSL? Maps display fine for me on my local machine without this code so I think it's probably not necessary?

src/static/js/lostfound.js
```diff
@@ -270,6 +282,21 @@ $(function() {
             if (this.mode == "found") { validate.indicator([ "datefound", "datereported", "owner" ]); }
         },
 
+        delay: function() {
+            let arealostfound = "";
+            if (controller.animal.AREALOST) {
```
bobintetley: should be using lostfound.mode == "lost"

src/main.py
```diff
@@ -5456,6 +5445,106 @@ def content(self, o):
...
+
+    def post_getmarkers(self, o):
+        markers = []
+        mk = o.post["mk"]
+        dbo = o.dbo
+        floor = o.post.date("floor")
+        shortcodes = asm3.configuration.use_short_shelter_codes(dbo)
+        if "l" in mk and self.checkb(asm3.users.VIEW_LOST_ANIMAL):
+            for m in asm3.lostfound.get_recent_lost_animals(dbo, floor):
+                popuptext = f'<a href="lostanimal?id={str(m.ID)}">{_("Lost {0} {1}").replace("{0}", m.SPECIESNAME.lower()).replace("{1}", str(m.ID))}</a>' \
```
bobintetley: It was this kind of delimiter soup that blew up the older Pythons on the servers. I think probably best to assign these expressions to variables before substituting for readability and to prevent upsetting older Python! 

Also, no need to do str() on anything in {} because Python does it automatically. When you have tokens {0}, {1} present in a string, you can use the Python function "format" to pass the arguments for those tokens in order.

```
linktext = _("Lost {0} {1}").format(m.SPECIESNAME.lower(), m.ID)
popupdesc = _("Lost {0}").format(python2display(o.locale, m.DATELOST)
popuptext = f'<a href="lostanimal?id={m.ID}>{linktext}</a><br>{popupdesc}<br>{m.AREALOST}<br>'
```

src/asm3/animal.py
adamdspencer77: Done.

adamdspencer77: Undone.

src/asm3/lostfound.py
adamdspencer77: Changed.

src/static/js/common_map.js
adamdspencer77: Removed.

src/static/js/lostfound.js
adamdspencer77: Changed.

src/main.py
adamdspencer77: Tidied up.

adamdspencer77:
Tested the new code on the dev server and hit some issues. Code to manipulate the map markers were needing access to leaflet libraries that weren't available yet. Put in a loop that checks that leaflet/google maps is available before continuing, not sure how you feel about my recursive loop, I wondered if there's a cleaner way to do this.

adamdspencer77:
Not tested on the dev server with google maps.

src/static/js/common_map.js
```diff
@@ -74,31 +138,41 @@ const mapping = {
         return fv;
     },
 
-    _leaflet_draw_map: function(divid, zoom, latlong, markers) {
+    _leaflet_draw_map: async function(divid, zoom, latlong, markers) {
```
bobintetley: I might have missed something here, but it looks like you've made this function async, and added await elsewhere, but this function has not been changed to do anything with promises? If that's the case, your async/await is not actually doing anything!

You should be creating a $.Deferred object at the beginning of the function, returning the deferred.promise() object at the end and resolving your promise when the map related stuff is completed (which will typically be based on some kind of event/callback from the mapping library).

Have a look at something like tableform.dialog_add_show for an example of how to use a promise.

adamdspencer77:
Tested google maps on dev server. Tidied UI.

src/static/js/mapview.js
```diff
@@ -0,0 +1,186 @@
...
+                $("#datefloor").datepicker("setDate", fl);
+            } else {
+                let datefloor = new Date();
+                datefloor.setDate(datefloor.getDate() - 90);
+                $("#datefloor").datepicker("setDate", datefloor);
+            }
+        },
+
+        _check_leaflet_loaded: function() {
+            try {
```
bobintetley: mapview should not have any knowledge of leaflet or google, it should be talking to the mapping module to do everything, certainly if it's something low level like making sure everything is initialised

src/static/js/mapview.js
```diff
@@ -0,0 +1,186 @@
...
+$(function() {
+
+    "use strict";
+
+    const mapview = {
+
+        update_markers_from_checkboxes: async function() {
+            let ticklength = 500;
+            let tickmessage = "Mapping library unavailable, waiting " + ticklength + " milliseconds";
+            if (asm.mapprovider == "osm") {
```
bobintetley: mapview should not have any knowledge of this stuff, this is the whole point of abstraction/encapsulation, it just calls the mapping module and the map gets drawn.

If you need to wait for the map to be initialised in some way, we should be doing it all inside the mapping module, and if it's internal functions in mapping that need to ensure the map is initialised, they should be the ones calling it too.

I would make a promise based function inside mapping called ready(). It resolves the promise when whatever condition is satisfied that the map is fully loaded. 

I'd also suggest that the mapping.redraw_markers function starts by calling this new ``` await mapping.ready() ``` function you've just added - then you don't need any of this stuff in mapview at all and it doesn't have to know or care what mapping library we're using. Then, your code in _update_markers_from_checkboxes can move into update_markers_from_checkboxes, replacing all of this timing/checking code.

PR #2107, opened 2026-06-22
Update the renewal column in ownerlicence when the new licence is of a different type to previous

adamdspencer77:
This is not a bug. In this scenario, the organisation is using licencing for dog licences (which many people do). However, there are other types of licences that can't just be renewed by "any" licence.

Since it is the most common circumstance of use, what I suggest we do is add a new option under a new "Animal Control" tab for "Only allow licenses to renew licenses of the same type" and make it unchecked by default. This makes the new option effectively the default, but people can turn this on to enable the old behaviour if they want.

### DIFF
--- a/src/static/js/licence.js
+++ b/src/static/js/licence.js
@@ -136,6 +136,9 @@ $(function() {
                                     controller.rows.push(row);
                                     tableform.table_update(table);
                                     tableform.dialog_close();
+                                    if (!common.current_url().includes("/licence")) {
+                                        common.route_reload();
+                                    }
                                 }
                                 finally {
                                     tableform.dialog_enable_buttons();
@@ -173,9 +176,13 @@ $(function() {
                                     controller.rows.push(row);
                                     tableform.table_update(table);
                                     tableform.dialog_close();
+                                    if (!common.current_url().includes("/licence")) {
+                                        common.route_reload();
+                                    }
                                 }
                                 finally {
                                     tableform.dialog_enable_buttons();
+                                    
                                 }
                             },
                             onload: function() {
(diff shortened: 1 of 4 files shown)


src/static/js/licence.js
```diff
@@ -136,6 +136,9 @@ $(function() {
                                     controller.rows.push(row);
                                     tableform.table_update(table);
                                     tableform.dialog_close();
+                                    if (!common.current_url().includes("/licence")) {
```
bobintetley: shouldn't need to inspect the current url. The pattern for most of these forms is to set controller.name to the endpoint, so you can just do if (controller.name == 'licence')

src/static/js/licence.js
```diff
@@ -173,9 +176,13 @@ $(function() {
...
```
bobintetley: Same again

src/static/js/licence.js
```diff
@@ -199,6 +206,9 @@ $(function() {
                         tableform.buttons_default_state(buttons);
                         let ids = tableform.table_ids(table);
                         await common.ajax_post("licence", "mode=delete&ids=" + ids);
+                        if (!common.current_url().includes("/licence")) {
```
bobintetley: and again ...

src/static/js/licence.js
adamdspencer77: Done.

src/static/js/licence.js
adamdspencer77: Done.

src/static/js/licence.js
adamdspencer77: Done.

PR #2111, opened 2026-06-23
Update address bar URL when filters are changed on calendar view

adamdspencer77:


src/static/js/calendarview.js
```diff
@@ -15,6 +15,7 @@ $(function() {
             $("#calendar").fullCalendar("removeEvents");
             $("#calendar").fullCalendar("removeEventSource", calendarview.calendar_events);
             $("#calendar").fullCalendar("addEventSource", calendarview.calendar_events);
+            window.history.pushState({}, '', '/calendarview?ev=' + ev);
```
bobintetley: Prefer native functions like this to be encapsulated. Add common.route_push(path) to common.js and have it basically do exactly what you're doing here ``` window.history.pushState({}, '', path) ```. Comment it to indicate it will replace the URL in the address bar, but bypasses the path/route mechanism.

One other thing, when you're setting URLs anywhere, you should never be putting a forward slash on the front - if you do, it will break this for any open source installations that are not accessible at /

adamdspencer77: Done.

PR #2115, opened 2026-06-26
add python3-lxml to list of required dependencies

leestran1995:
First off, an introduction!

Hi! My name is Lee, I'm a professional software developer (just hit 7 years of industry experience this month, wow)! Today I volunteered at a cat shelter for the first time and they set me up with access to their shelter manager instance, I was curious about the website and was excited to find out it's open source and figured I could volunteer my time in another way, too!

While setting up the project locally, I wasn't able to get it started until I installed `python3-lxml`, so this adds it to the README.

Nice to meet you all, please feel free to point me towards any smaller outstanding issues that might be a good first bit for me to tackle, otherwise I'm going to get it running locally, play with it, and see if I can break anything >:)

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@ install all the software you need to run ASM. If you are using the
 sheltermanager3 deb package it already has dependencies set for these and will
 install them for you.
 
-* apt-get install make python3 python3-cheroot python3-pil python3-mysqldb python3-psycopg2 python3-memcache
+* apt-get install make python3 python3-cheroot python3-pil python3-mysqldb python3-psycopg2 python3-memcache python3-lxml
 
 Extra, non-mandatory packages:
 


leestran1995:
I ran into a handful of other issues that _might_ need to get added to the local startup sequence, but I want to do a completely fresh clone and build before I actually commit them to this branch (and potentially to the README). While I had Claude for writing code, I used it for helping get this set up and here's it's summary of what I had to do to get the test instance running properly. 

- Build tools and runtime dependencies need to be installed 
  - `apt-get install memcached flake8 exuberant-ctags nodejs npm`
- For sessions to work, memcached has to be running otherwise you just get redirected straight back to the login page
  - `service memcached start`
- Javascript dependencies must be installed without sudo
  - `npm install`
- Static assets need to be built
  - `make clean rollup schema`
- Then we're ready to start the test instance
  - `make test`

bobintetley:
Thanks for the feedback on the README. I'm not sure if it's in there (I'll check), but there should be a convenient "deps" target in the Makefile, so you can run ```make deps``` to install everything that's needed.

Issue #1326, opened 2023-09-10
Bulk create medical/vacc/test keep list after creation


bobintetley:
One customer said it will help them a lot if the animal multi-select tool wasn't cleared after creating bulk medical items. This is because they will deworm and flea groups of animals at the same time.

I can't see how it would hurt anything to do this as there is a button to clear the list right next to it if you don't want that.

bobintetley:
new_bulk_X / onload is manually clearing the list, so it would be easy to remove there. The medical screens reload with the default option under Settings->Options->Medical, so I think that would need to be turned off in order for the multiselect not to clear.

leestran1995:
I've been hacking at this and it's a little more involved than originally expected😅 

The issue isn't so much getting the selected animals to persist, it's getting them to do so while also displaying them in the vaccination table after submission.

### Single-Vaccination

The single-vaccination code (and the single animal picker) are clever. When an animal is selected, it gets sent in a "change" trigger. 

(As an aside, forgive my parlance here, I'm still very new to pure jQuery, so "trigger" might not be the right word)

Anyways, that then gets picked up by `vaccination.js` in the `bind` function: 

```$("#animal").on("change loaded", function(event, rec) { vaccination.lastanimal = rec; });```

And then, when `new_vacc` calls `vaccination.set_extra_fields`, it uses that `lastanimal` to fill in the rest of the data for the `row`. 

(Another aside, there is a minor bug in that code, it doesn't fill the "Species" or "GivenBy" fields in the table, those only populate on page refresh)

### Bulk Vaccination

The `animalchoosermulti`, is more spartan, functionality-wise. The animals get checkboxed, and once "select" is clicked it simply loads those animals' IDs...... somewhere (sorry, I'm not precisely sure where). But those IDs get sent by:

```
await tableform.fields_post(dialog.fields, "mode=createbulk", "vaccination");
```

and then the page simply gets reloaded, which fetches all the data from the backend to repopulate the table.

## Two possible solutions

### The Easy Way

Turning off the reload page setting, populating the table with _something_ (just the IDs that we have to work with), and not clearing the animal multichooser gets us the list persisting after submission, but leaves us with an awfully ugly table until the page gets refreshed:

<img width="1737" height="506" alt="Image" src="https://github.com/user-attachments/assets/0b134b0d-0355-4745-98f4-011437d8b355" />

### The Better (More Difficult Way)

[...]

PR #2116, opened 2026-06-27
persist bulk vaccination animal selection on submission

leestran1995:
Demo: https://www.loom.com/share/e737aa754b834881a2a770b720a70cbe

- Use session storage to save selected animals from multichooser to add to vaccination table on form submission
- Do not clear or close bulk vaccination form until the user clicks "cancel" 
- Add Species-Name and Given-By to table on single animal vaccination submission
- Return IDs in backend `createBulk` endpoint to match functionality of non-bulk endpoint

Not a lot of lines, but quite a bit of effort

### DIFF
--- a/src/static/js/vaccination.js
+++ b/src/static/js/vaccination.js
@@ -344,7 +344,7 @@ $(function() {
         },
 
         new_bulk_vacc: function() {
-            let dialog = vaccination.dialog;
+            let dialog = vaccination.dialog, table = vaccination.table;
             tableform.dialog_show_add(dialog, {
                 onvalidate: function() {
                     return validate.notblank([ "animals" ]);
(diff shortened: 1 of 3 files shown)


src/static/js/vaccination.js
```diff
@@ -344,28 +344,40 @@ $(function() {
...
+                oncancel: async function() {
+                    sessionStorage.removeItem("last_multi_animals");
+                },
                 onload: function() {
                     $("#animalrow").hide();
                     $("#animalsrow").show();
-                    $("#animals").animalchoosermulti("clear");
                     $("#dialog-tableform .asm-textbox, #dialog-tableform .asm-textarea").val("");
                     $("#type").select("value", config.str("AFDefaultVaccinationType"));
+                    $("#animals").animalchoosermulti("clear");
```
bobintetley, line 379: Simply removing this line would have kept the selected animals in the widget if that's what you were trying to do?

src/static/js/vaccination.js
```diff
@@ -344,28 +344,40 @@ $(function() {
...
-                        tableform.dialog_close();
-                        common.route_reload();
+                        let chosenanimalsstring = sessionStorage.getItem("last_multi_animals");
+                        let chosenanimals = JSON.parse(chosenanimalsstring);
+                        let chosentype = $("#type option:selected").text() 
+                        chosenanimals.forEach(animal => {
+                            animal.VACCINATIONTYPE = chosentype;
+                            tableform.fields_update_row(dialog.fields, animal);
+                            controller.rows.push(animal);
+                        })
```
bobintetley, line 362: The animalchoosermulti widget has a "value" method that allows you to retrieve or set its value as a string of comma separated ids. 

``` 
let v = $("#animals").animalchoosermulti("value");
$("#animals").animalchoosermulti("value", v);
```

Manually extracting values from checkboxes to lookup records and then serialising the entire records into JSON to go in sessionstorage seems way OTT. Did AI write this?

bobintetley (commented):
Many thanks for your efforts and contributing to ASM. I'm really sorry but I don't approve of the implementation of this feature. I really hope this doesn't put you off contributing in future.

src/static/js/vaccination.js
leestran1995: >Did AI write this?

Your opening comment doesn't put me off from contributing in the future, but this sure does. I take some amount of pride in specifically _not_ using AI for writing code, if I continue contributing in the future I do hope you'll keep that in mind.

I detailed my investigations, and the reason I settled on this approach over here: https://github.com/sheltermanager/asm3/issues/1326

src/static/js/vaccination.js
leestran1995: The bulk chooser refreshes the page on submission, so the animals don't remain in the widget.

bobintetley: Fair point, I'd forgotten it did that. I think the approach to persisting should be using the value method of animalchoosermulti to get the ids and storing them in the module, vaccination.lastanimals. They can be brought back when the dialog is shown by setting the value. 

When the screen does common.route_reload() after the bulk insert, this doesn't do a page transition, it only reloads the data from the backend via ajax.

src/static/js/vaccination.js
bobintetley: You did, and I apologise, I didn't see your previous explanation on that issue. 

I'm also strongly against the use of AI for writing code and we've had a few low quality vibe-coded submissions recently, which has made me unfairly view new contributions in a negative light. Learning our codebase isn't easy and I do really appreciate you wanting to contribute your time and skills to helping with ASM. Thank you.

PR #2118, opened 2026-06-29
persistence bulk validation (take 2)!

leestran1995:
Following @bobintetley's directions [here](https://github.com/sheltermanager/asm3/issues/1326#issuecomment-4825367806)

In vaccination, medical, and test:

- on bulk form submission, persist IDs and rows in parent elements 
  - IDs are "invisible" but are what gets sent to the backend, rows are required for the animals to be displayed on element reload
- On bulk form reload, if IDs and rows were saved, load them back into the bulk form
- On bulk selection clear, clear the persisted IDs and rows in parent elements

### DIFF
--- a/src/static/js/common_animalchoosermulti.js
+++ b/src/static/js/common_animalchoosermulti.js
@@ -131,6 +131,8 @@ $.fn.animalchoosermulti = asm_widget({
         o.results.find(":checked").prop("checked", false);
         this.update_status(t);
         if (fireclearedevent) { t.trigger("cleared"); }
+
+        // Used by bulk-selection persistence regardless of fireclearedevent
         t.trigger("bulk_clear");
     },
 
--- a/src/static/js/common_animalchoosermulti.js
+++ b/src/static/js/common_animalchoosermulti.js
@@ -164,6 +164,10 @@ $.fn.animalchoosermulti = asm_widget({
         return t.data("o").rows;
     }, 
 
+    set_rows: function(t, newrows) {
+        t.data("o").rows = newrows;
+    },
+
     get_selected_rows: function(t) {
         let rows = [];
         let self = this;
@@ -192,6 +196,7 @@ $.fn.animalchoosermulti = asm_widget({
      */
     select: function(t) {
         let self = this, o = t.data("o"), selval = [];
+        console.log(self.get_rows(t));
         o.display.html("");
         o.results.find(":checked").each(function() {
             let aid = $(this).attr("data"), rec = self.get_row(t, aid);
@@ -378,6 +383,7 @@ $.fn.animalchoosermulti = asm_widget({
         let o = t.data("o");
         o.display.html("");
         $.each(newval.split(","), function(i, v) {
+            console.log(v);
             let rec = self.get_row(t, parseInt(v));
             let disp = "<a class=\"asm-embed-name\" href=\"animal?id=" + rec.ID + "\">" + rec.CODE + " - " + rec.ANIMALNAME + "</a>";
             o.display.append(disp);
--- a/src/static/js/common_animalchoosermulti.js
+++ b/src/static/js/common_animalchoosermulti.js
@@ -131,6 +131,7 @@ $.fn.animalchoosermulti = asm_widget({
         o.results.find(":checked").prop("checked", false);
         this.update_status(t);
         if (fireclearedevent) { t.trigger("cleared"); }
+        t.trigger("bulk_clear");
     },
 
     is_empty: function(t) {
@@ -196,7 +197,6 @@ $.fn.animalchoosermulti = asm_widget({
      */
     select: function(t) {
         let self = this, o = t.data("o"), selval = [];
-        console.log(self.get_rows(t));
         o.display.html("");
         o.results.find(":checked").each(function() {
             let aid = $(this).attr("data"), rec = self.get_row(t, aid);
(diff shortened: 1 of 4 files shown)


src/static/js/common_animalchoosermulti.js
```diff
@@ -131,6 +131,9 @@ $.fn.animalchoosermulti = asm_widget({
         o.results.find(":checked").prop("checked", false);
         this.update_status(t);
         if (fireclearedevent) { t.trigger("cleared"); }
+
+        // Used by bulk-selection persistence regardless of fireclearedevent
+        t.trigger("bulk_clear");
```
leestran1995, line 136: I wasn't sure if it was safe to set the default for `fireclearedevent` to `true`, so I made a second trigger that always fires to listen on

src/static/js/common_animalchoosermulti.js
```diff
@@ -164,6 +167,10 @@ $.fn.animalchoosermulti = asm_widget({
         return t.data("o").rows;
     }, 
 
+    set_rows: function(t, newrows) {
+        t.data("o").rows = newrows;
+    },
```
leestran1995, line 172: `value` is the set of integer IDs, but in order to get the animals' display information into the bulk chooser we need to save it back to rows.

src/static/js/medical.js
```diff
@@ -616,11 +621,14 @@ $(function() {
                 onload: function() {
                     $("#animalrow").hide();
                     $("#animalsrow").show();
-                    $("#animals").animalchoosermulti("clear");
+                    if (medical.lastanimalsrows != null && medical.lastanimals != null) {
+                        $("#animals").animalchoosermulti("set_rows", medical.lastanimalsrows);
+                        $("#animals").animalchoosermulti("value", medical.lastanimals);
```
leestran1995, line 626: The component renders the text into the animals row in the bulk chooser on the call to `value`

src/static/js/medical.js
```diff
@@ -85,6 +87,7 @@ $(function() {
                 edit_title: _("Edit medical regimen"),
                 edit_perm: 'mcam',
                 close_on_ok: false,
+                use_default_values: false,
```
leestran1995, line 90: If this is `true` the animal chooser multi gets cleared. This is the easy way (and also has parity with `vaccination` and `test`, but I wouldn't be surprised if we want this to be `true` and give some other sort of directive to not clear the animalchoosermulti further down the call stack

src/static/js/medical.js
```diff
@@ -603,6 +606,8 @@ $(function() {
                     return validate.notblank([ "animals" ]);
                 },
                 onadd: async function() {
+                    medical.lastanimals = $("#animals").animalchoosermulti("value");
+                    medical.lastanimalsrows = $("#animals").animalchoosermulti("get_rows");
```
bobintetley, line 610: I might be missing something, but why is lastanimalsrows being persisted at all? The chooser should already have rows for all shelter animals, so it should only be necessary to get and set the value of selected ids?

src/static/js/medical.js
```diff
@@ -616,7 +621,10 @@ $(function() {
                 onload: function() {
                     $("#animalrow").hide();
                     $("#animalsrow").show();
-                    $("#animals").animalchoosermulti("clear");
+                    if (medical.lastanimalsrows != null && medical.lastanimals != null) {
+                        $("#animals").animalchoosermulti("set_rows", medical.lastanimalsrows);
```
bobintetley, line 625: was there a problem that is solved by making and calling set_rows?

src/static/js/common_animalchoosermulti.js
bobintetley: this shouldn't be necessary? the rows containing all shelter animals should be loaded when the component initialises, so restoring the value should be ok

src/static/js/common_animalchoosermulti.js
bobintetley: It definitely should be safe to listen for the cleared event rather than needing to create a whole new event. If it's not then it's a bug that wants fixing as that event's entire reason for existing is for callers to find out when the widget has been manually cleared by the user clicking the clear button!

bobintetley:
I think I've just realised why you were adding set_rows. I guess there's a race condition, where if you set the value before the widget has finished loading, the rows aren't available.

The fix for this is to have the value setting code wait until the widget has finished initialising before actually setting the value, this can be done via monitoring a semaphore that gets set once the ajax call to get the rows completes or something.

src/static/js/common_animalchoosermulti.js
leestran1995: Clarifying question, which of the following two functionalities do you want?

1. When a user clicks "New Bulk Vaccination", the last set of animals they chose is already in the "Add Bulk Vaccination" window, without having to click the magnifying glass
2. When a user clicks "New Bulk Vaccination", the list of animals is empty, but when they click the magnifying glass the animals they had checked last time are still checked

I coded up #1 because it's the better user experience, in my opinion. The issue isn't that it's a race condition, it's that the animal chooser doesn't actually fetch the animals from the backend until the user clicks the magnifying glass.

bobintetley: ah, thanks for clarifying. Yes I agree option 1 is the preferred. The animalchoosermulti.value method should check/fetch the animal records before applying the new value.

leestran1995: Why bother with the unnecessary API call when we can just save it locally, the same way we do with the IDs, as I'm doing here?

PR #2119, opened 2026-07-06
Bootstrap online forms

adamdspencer77:
Integrate the snazzy bootstrap formatting to online forms. Must ensure that the system stays backwards compatible.


src/asm3/dbupdates/35125.py
```diff
@@ -0,0 +1,4 @@
+from asm3.dbupdate import add_column, execute
+
+add_column(dbo, "onlineform", "BootstrapStyle", dbo.type_integer)
+execute(dbo, "UPDATE onlineform SET BootstrapStyle = 0")
```
bobintetley: ooh, bad name! Make it FormRenderer, where legacy = 0 and bootstrap = 1, this means if something takes over from bootstrap in future we can make that an option

src/asm3/onlineform.py
```diff
@@ -156,7 +157,7 @@ def get_onlineforms(dbo: Database) -> Results:
     """ Return all online forms """
     return dbo.query("SELECT *, (SELECT COUNT(*) FROM onlineformfield WHERE OnlineFormID = onlineform.ID) AS NumberOfFields FROM onlineform ORDER BY Name")
 
-def get_onlineform_html(dbo: Database, formid: int, completedocument: bool = True) -> str:
+def get_onlineform_bootstrap_html(dbo: Database, formid: int, completedocument: bool = True) -> str:
```
bobintetley: I'd call these 2 methods _get_onlineform_html_bootstrap and _get_onlineform_html_legacy (note the underscore prefix) - can then make get_onlineform_html accept a "renderer" parameter that defaults to RENDERER_LEGACY that calls the correct implementation.

Callers outside this module should include the renderer parameter.

src/static/js/onlineforms.js
```diff
@@ -97,7 +97,12 @@ $(function() {
                         callout: _("The confirmation email message to send to the form submitter.") }, 
                     { json_field: "DESCRIPTION", post_field: "description", label: _("Description"), type: "htmleditor", height: "100px", width: "600px" },
                     { json_field: "HEADER", post_field: "header", label: _("Header"), type: "htmleditor", height: "100px", width: "600px" },
-                    { json_field: "FOOTER", post_field: "footer", label: _("Footer"), type: "htmleditor", height: "100px", width: "600px" }
+                    { json_field: "FOOTER", post_field: "footer", label: _("Footer"), type: "htmleditor", height: "100px", width: "600px" },
+                    { json_field: "BOOTSTRAPSTYLE", post_field: "bootstrapstyle", 
+                        label: _("Use Bootstrap styled form"), 
+                        type: "check",
+                        callout: _("If your header/footer is not expecting Bootstrap style, the results may be poor")
```
bobintetley: This should be a select box instead labeled "Renderer" that has options for legacy (0) and bootstrap (1)

src/asm3/onlineform.py
```diff
@@ -18,6 +18,7 @@
 
 from asm3.sitedefs import BASE_URL, SERVICE_URL
 from asm3.sitedefs import ASMSELECT_CSS, ASMSELECT_JS, JQUERY_JS, JQUERY_UI_JS, JQUERY_UI_CSS, SIGNATURE_JS, TIMEPICKER_CSS, TIMEPICKER_JS
+from asm3.sitedefs import BOOTSTRAP_JS, BOOTSTRAP_CSS
 from asm3.typehints import Any, datetime, Database, List, PostedData, ResultRow, Results, Tuple
 
 FIELDTYPE_YESNO = 0
```
bobintetley, line 25: add constants for 

```
RENDERER_LEGACY = 0 
RENDERER_BOOTSTRAP = 1
```

src/asm3/dbupdates/35125.py
adamdspencer77: Changed.

src/asm3/onlineform.py
adamdspencer77: Kind of done. FormRenderer is worked out using the formid. I guess that the FormRenderer could be worked out form the formid in main.py instead and then passed to get_form_html but I thought that this way main.py would be left cleaner.

src/static/js/onlineforms.js

src/asm3/onlineform.py
adamdspencer77: Done.

src/static/js/onlineforms.js
```diff
@@ -97,7 +97,11 @@ $(function() {
                         callout: _("The confirmation email message to send to the form submitter.") }, 
                     { json_field: "DESCRIPTION", post_field: "description", label: _("Description"), type: "htmleditor", height: "100px", width: "600px" },
                     { json_field: "HEADER", post_field: "header", label: _("Header"), type: "htmleditor", height: "100px", width: "600px" },
-                    { json_field: "FOOTER", post_field: "footer", label: _("Footer"), type: "htmleditor", height: "100px", width: "600px" }
+                    { json_field: "FOOTER", post_field: "footer", label: _("Footer"), type: "htmleditor", height: "100px", width: "600px" },
+                    { json_field: "FORMRENDERER", post_field: "formrenderer", label: _("Use Bootstrap styled form"), type: "select",
```
bobintetley: The label should just be "Renderer". You should also be able to set a defaultval attribute to "1" to have new forms default to bootstrap in future.

src/static/js/onlineforms.js
```diff
@@ -97,7 +97,11 @@ $(function() {
                         callout: _("The confirmation email message to send to the form submitter.") }, 
                     { json_field: "DESCRIPTION", post_field: "description", label: _("Description"), type: "htmleditor", height: "100px", width: "600px" },
                     { json_field: "HEADER", post_field: "header", label: _("Header"), type: "htmleditor", height: "100px", width: "600px" },
-                    { json_field: "FOOTER", post_field: "footer", label: _("Footer"), type: "htmleditor", height: "100px", width: "600px" }
+                    { json_field: "FOOTER", post_field: "footer", label: _("Footer"), type: "htmleditor", height: "100px", width: "600px" },
+                    { json_field: "FORMRENDERER", post_field: "formrenderer", label: _("Use Bootstrap styled form"), type: "select", 
+                        options: '<option value="0">' + _("Legacy") + '</option><option value="1">' + _("Bootstrap") + '</option>', 
+                        callout: _("If your header/footer is not expecting this style, the results may be poor. The default header/footer is compatible")
```
bobintetley: "The form renderer to use. Some renderers require specific styles to be set in the header. If you have issues use the Header/Footer button to revert to the default".

src/static/js/onlineforms.js

src/static/js/onlineforms.js

PR #2124, opened 2026-07-09
Conditions tab, include description

adamdspencer77:
Similar to what we do with medical types being shown below regimen names in the list on the medical tab, include the condition description from the lookup item below the condition name in the list.

### DIFF
--- a/src/static/js/animal_condition.js
+++ b/src/static/js/animal_condition.js
@@ -105,7 +105,7 @@ $(function() {
             this.model();
             let s = "";
             s += tableform.dialog_render(this.dialog);
-            s += edit_header.animal_edit_header(controller.animal, "condition", controller.tabcounts);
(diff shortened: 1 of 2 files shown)


src/static/js/animal_condition.js
```diff
@@ -105,7 +113,7 @@ $(function() {
             this.model();
             let s = "";
             s += tableform.dialog_render(this.dialog);
-            s += edit_header.animal_edit_header(controller.animal, "condition", controller.tabcounts);
+            s += edit_header.animal_edit_header(controller.animal, "conditions", controller.tabcounts);
```
adamdspencer77, line 117: Noticed that the Conditions tab wasn't being highlighted when it was active.

Tetzor:
The description doesn't appear after adding the condition until you refresh the screen or leave and return.

PR #2126, opened 2026-07-17
Built in flag lists

adamdspencer77:
Adding new person and animal flags is now a massive pain, in particular because there are now so many places where the built in list is duplicated throughout the code.

This needs to change, make the list of built in animal and person flags dictionaries at the top of lookups.py. Consider how we can do it in future without having to modify so much code.

These dictionaries should include the name of the field they are set by for ISACO etc, the translated text of the flag and the value of the flag for a dropdown and in the ADDITIONALFLAGS column.

These are places that are affected when a person or animal flag is added:

animal.py

insert_animal_from_form, update_animal_from_form

person.py

insert_person_from_form, update_person_from_form (set the field to 0)
update_flags

lookups.py

_merge_db_flags

common_html.js

animal_flag_options
person_flag_options

header_edit_header.js

animal_flags
person_flags

lookups.js

add_builtins

### DIFF
--- a/src/asm3/animal.py
+++ b/src/asm3/animal.py
@@ -3574,16 +3574,13 @@ def update_animal_from_form(dbo: Database, post: PostedData, username: str) -> N
             asm3.diary.complete_diary_notes_for_animal(dbo, username, aid)
 
     # Sort out any flags
-    def bi(b): 
+    def bi(b):
         return b and 1 or 0
 
+    def fb(v, flags):
+        return bi(v in flags)
+
     flags = post["flags"].split(",")
-    courtesy = bi("courtesy" in flags)
-    crueltycase = bi("crueltycase" in flags)
-    notforadoption = bi("notforadoption" in flags)
-    notforregistration = bi("notforregistration" in flags)
-    nonshelter = bi("nonshelter" in flags)
-    quarantine = bi("quarantine" in flags)
     flagstr = "|".join(flags) + "|"
 
     # If the option is on and the flags have changed, log it
@@ -3596,16 +3593,16 @@ def update_animal_from_form(dbo: Database, post: PostedData, username: str) -> N
     # day. Non shelter animals don't have visible movements and this prevents a bug where
     # an open foster/retailer movement on a non-shelter animal can make it publish for adoption
     # when the "include fosters/retailers" publishing options are on.
-    if nonshelter == 1:
+    if fb("nonshelter", flags):
         dbo.execute("UPDATE adoption SET ReturnDate = MovementDate WHERE MovementType IN (2,8) AND AnimalID = ?", [aid])
 
     dbo.update("animal", aid, {
-        "NonShelterAnimal":     nonshelter,
-        "IsNotAvailableForAdoption": notforadoption,
-        "IsNotForRegistration": notforregistration,
-        "IsQuarantine":         quarantine,
-        "IsCourtesy":           courtesy,
-        "CrueltyCase":          crueltycase,
+        "NonShelterAnimal":     fb("nonshelter", flags),
+        "IsNotAvailableForAdoption": fb("notforadoption", flags),
+        "IsNotForRegistration": fb("notforregistration", flags),
+        "IsQuarantine":         fb("quarantine", flags),
+        "IsCourtesy":           fb("courtesy", flags),
+        "CrueltyCase":          fb("crueltycase", flags),
         "AdditionalFlags":      flagstr,
         "ShelterCode":          post["sheltercode"],
         "ShortCode":            post["shortcode"],
--- a/src/asm3/animal.py
+++ b/src/asm3/animal.py
@@ -3577,7 +3577,7 @@ def update_animal_from_form(dbo: Database, post: PostedData, username: str) -> N
     def bi(b):
         return b and 1 or 0
 
-    def fb(v, flags):
+    def fb(v):
         return bi(v in flags)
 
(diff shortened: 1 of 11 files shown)


src/static/js/onlineforms.js
```diff
@@ -167,7 +167,28 @@ $(function() {
...
+                            $.each(row.SETOWNERFLAGS.split(","), function(i, v) {
+                                if ($.inArray(v, stock) > 0) {
+                                    flags.push(labels[v]);
+                                } else {
+                                    flags.push(v);
+                                }
+                            });
+                        }
+                        flags.sort();
+                        return flags.join(", ");
```
bobintetley: It's good that you want to resolve the built in tokens to their labels (something it did not do before), but I feel like this code should definitely be a function in common_html.js rather than here. Something like ``` html.substitute_person_flag_names(row.SETOWNERFLAGS) ```

I'm sure we show flags in places like the banner on animal and person records, and in the global search results, so this new function might replace anything those functions might be doing themselves to swap built in keys for labels

src/static/js/header_edit_header.js
```diff
@@ -691,23 +691,15 @@ edit_header = {
...
-            flags.push(_("Courtesy Listing"));
-        }
-        if (a.CRUELTYCASE == 1) {
-            flags.push(_("Cruelty Case"));
-        }
-        if (a.NONSHELTERANIMAL == 1) {
-            flags.push(_("Non-Shelter"));
-        }
-        if (a.ISNOTAVAILABLEFORADOPTION == 1) {
-            flags.push("<span style=\"color: red\">" + _("Not For Adoption") + "</span>");
```
bobintetley, line 704: I might have missed it, but is this style being retained?

src/static/js/header_edit_header.js
```diff
@@ -724,79 +716,15 @@ edit_header = {
...
-            flags.push("<span class=\"asm-flag-dangerous\">" + _("Dangerous") + "</span>");
-        }
-        if (p.INVESTIGATION > 0) {
-            flags.push("<span class=\"asm-flag-investigation\">" + _("Investigation") + "</span>");
-        }
-        if (p.INCIDENT > 0) {
-            flags.push("<span class=\"asm-flag-incident\">" + _("Incident") + "</span>");
-        }
-        if (p.ISDECEASED == 1) {
-            flags.push("<span class=\"asm-flag-deceased\">" + _("Deceased") + "</span>");
```
bobintetley, line 743: Are these classes being retained? I don't see them in the dict definitions in lookups.py but I guess they could be inferred from the flag key as asm-flag-X

src/static/js/onlineforms.js

src/static/js/header_edit_header.js
adamdspencer77: Nope. Fixed.

src/static/js/header_edit_header.js
adamdspencer77: Done.

src/asm3/animal.py
```diff
@@ -3596,16 +3593,16 @@ def bi(b):
...
-        "IsNotForRegistration": notforregistration,
-        "IsQuarantine":         quarantine,
-        "IsCourtesy":           courtesy,
-        "CrueltyCase":          crueltycase,
+        "NonShelterAnimal":     fb("nonshelter", flags),
+        "IsNotAvailableForAdoption": fb("notforadoption", flags),
+        "IsNotForRegistration": fb("notforregistration", flags),
+        "IsQuarantine":         fb("quarantine", flags),
+        "IsCourtesy":           fb("courtesy", flags),
+        "CrueltyCase":          fb("crueltycase", flags),
```
bobintetley: flags is already a local variable that your fb() function can see without you having to pass it every time. It could just be fb("nonshelter") etc.

src/asm3/person.py
```diff
@@ -1348,27 +1329,26 @@ def bi(b):
...
-        "IsShelter":                shelter,
-        "IsACO":                    aco,
-        "IsStaff":                  staff,
-        "IsFosterer":               fosterer,
-        "IsRetailer":               retailer,
-        "IsVet":                    vet,
-        "IsSponsor":                sponsor,
-        "IsGiftAid":                giftaid,
-        "IsSupplier":               supplier,
+        "ExcludeFromBulkEmail":     fb("excludefrombulkemail", flags),
```
bobintetley: Same as before, fb can see flags, no need to pass it.

src/asm3/animal.py
adamdspencer77: Changed.

src/asm3/person.py
adamdspencer77: Changed.

src/asm3/person.py
```diff
@@ -1348,27 +1329,26 @@ def bi(b):
...
-        "IsShelter":                shelter,
-        "IsACO":                    aco,
-        "IsStaff":                  staff,
-        "IsFosterer":               fosterer,
-        "IsRetailer":               retailer,
-        "IsVet":                    vet,
-        "IsSponsor":                sponsor,
-        "IsGiftAid":                giftaid,
-        "IsSupplier":               supplier,
+        "ExcludeFromBulkEmail":     fb("excludefrombulkemail"),
```
bobintetley, line 1333: It looks like the IDCheck/homechecked flag has been removed?

adamdspencer77: Yes it had. Put back in.

PR #2132, opened 2026-07-20
Record alert/warning system

adamdspencer77:
One customer has said they would like to be able to receive an alert or warning on the animal's record when it has a due vacc, test or medical item. This is a really good idea.

Need to experiment with what looks/works the best. Try the global html.warn function first that puts a box at the top of the screen.

I think the way to do this is to have an exclamation mark type icon shown in the tab.

We could extend animal.get_satellite_counts to include counts of overdue items for each tab as well. Make it a suffix called "due", so for example while we have a count of vacc records as "vaccination", we could also add a count of due vacces call "vaccinationdue".

The code in header_edit_header.js that outputs the tabs can then look for tab + "due" and if it exists and is non-zero, adds the exclamation mark. This could be done in the check_display_icon function at the top of animal_edit_header to make it work for any tab.

### DIFF
--- a/src/asm3/animal.py
+++ b/src/asm3/animal.py
@@ -3065,25 +3065,24 @@ def get_satellite_counts(dbo: Database, animalid: int) -> Results:
     """
     return dbo.query("SELECT a.ID, " \
         "(SELECT COUNT(*) FROM animalvaccination av WHERE av.AnimalID = a.ID) AS vaccination, " \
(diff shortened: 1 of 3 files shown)


adamdspencer77:
Warning icon should show for overdue:
- animal vaccinations
- animal tests
- animal medical treatments
- person payment
- person citations
- person traploans

PR #2133, opened 2026-07-21
Method of calculating days on shelter for a date range (DBU 35127)

adamdspencer77:
This is something that comes up fairly often. Some shelters want to be able to calculate how many days an animal was on shelter for a given period. I'm sure the RSPCA SAWA report wanted to do something like this as well.

I gave up on the making an SQL token or using code to generate this on the fly. I think the best way is a clever data structure:

Create a new table, animalfiguresonshelter with these columns animalid, monthmidpoint, month, year, daysonshelter. index on animalid

This table should be calculated by a function in animal.py called update_animal_figures_onshelter(animalid). The function first deletes any previous rows for animalid, then regenerates them from the animal's data and movements. Every month that the animal has ever been on shelter will be given a day count. Figuring out which months the animal was on shelter will be the tricky bit. The monthmidpoint field will contain a date of half way through that month. This will make it easy to filter this table with a date range for reports.

This function will be called from insert_animal_from_form, update_animal_from_form as well as movement.insert/update/delete_movement

There should be a new option under batch.js/trigger batch processes that allows you to call this function for all animals (which will take a while in some databases).

### DIFF
--- a/src/asm3/animal.py
+++ b/src/asm3/animal.py
@@ -3533,6 +3533,8 @@ def insert_animal_from_form(dbo: Database, post: PostedData, username: str) -> i
     
     update_animallocation(dbo, nextid, username)
 
+    update_animal_figures_onshelter(dbo, nextid, username)
+
     return (nextid, get_code(dbo, nextid))
 
 def update_animal_from_form(dbo: Database, post: PostedData, username: str) -> None:
@@ -3756,6 +3758,8 @@ def update_animal_from_form(dbo: Database, post: PostedData, username: str) -> N
         update_litter_count(dbo, post["litterid"])
     
     update_animallocation(dbo, aid, username)
+
+    update_animal_figures_onshelter(dbo, aid, username)
 
 def update_flags(dbo: Database, username: str, animalid: int, flags: List[str]) -> None:
     """
@@ -7332,15 +7336,15 @@ def update_animal_figures_onshelter(dbo: Database, animalid: int, username: str)
     figures = {}
     for osdate in onshelterdates:
         if osdate.month != month or osdate.year != year:
-            figureskey = f"{osdate.month}-{osdate.year}"
+            figureskey = (osdate.month, osdate.year)
             figures[figureskey] = 0
             month = osdate.month
             year = osdate.year
         figures[figureskey] += 1
 
     for f in figures.items():
-        month = int(f[0].split("-")[0])
-        year = int(f[0].split("-")[1])
+        month = int(f[0][0])
+        year = int(f[0][1])
         values = {
             "AnimalID": animalid,
             "MonthMidPoint": datetime(year, month, 15),
@@ -7349,5 +7353,3 @@ def update_animal_figures_onshelter(dbo: Database, animalid: int, username: str)
             "DaysOnShelter": f[1]
         }
         dbo.insert("animalfiguresonshelter", values, username, False)
-    
-    return figures
--- a/src/asm3/animal.py
+++ b/src/asm3/animal.py
@@ -7286,6 +7286,14 @@ def create_waitinglist(dbo: Database, username: str, aid: int) -> int:
     }
     return asm3.waitinglist.insert_waitinglist_from_form(dbo, asm3.utils.PostedData(data, l), username)
 
+def update_all_animal_figures_onshelter(dbo: Database, username: str) -> str:
+    animals = dbo.query_list("SELECT ID FROM animal")
+    asm3.asynctask.set_progress_max(dbo, len(animals))
+    for a in enumerate(animals):
+        update_animal_figures_onshelter(dbo, a[1], username)
+        asm3.asynctask.set_progress_value(dbo, a[0])
+    return "OK"
+
 def update_animal_figures_onshelter(dbo: Database, animalid: int, username: str):
     # Delete existing animalfiguresonshelter rows with this animalid
     dbo.delete("animalfiguresonshelter", "AnimalID = %s" % animalid)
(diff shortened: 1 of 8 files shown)


adamdspencer77:
I wondered if `update_animal_figures_onshelter` should be called when an animal is deleted.

src/asm3/animal.py
```diff
@@ -7284,3 +7286,78 @@ def create_waitinglist(dbo: Database, username: str, aid: int) -> int:
     }
     return asm3.waitinglist.insert_waitinglist_from_form(dbo, asm3.utils.PostedData(data, l), username)
 
+def update_all_animal_figures_onshelter(dbo: Database, username: str) -> str:
+    animals = dbo.query_list("SELECT ID FROM animal")
+    asm3.asynctask.set_progress_max(dbo, len(animals))
+    for a in enumerate(animals):
```
bobintetley: It's more readable if you destructure the 2 elements coming back from enumerate - ``` for i, a in enumerate(animals) ``` - you can then just refer to a and i rather than a[0] and a[1]

src/asm3/animal.py
```diff
@@ -7284,3 +7286,78 @@ def create_waitinglist(dbo: Database, username: str, aid: int) -> int:
...
+        update_animal_figures_onshelter(dbo, a[1], username)
+        asm3.asynctask.set_progress_value(dbo, a[0])
+    return "OK"
```
bobintetley: I think we normally return the number of elements affected with the OK ``` return f"OK {len(animals)}" ```

src/asm3/animal.py
```diff
@@ -7284,3 +7286,78 @@ def create_waitinglist(dbo: Database, username: str, aid: int) -> int:
...
+                        break
+        else:
+            nextinboundmovement = None
+            if inboundmovements:
+                nextinboundmovement = inboundmovements[0]
+                if date == nextinboundmovement:
+                    onshelterdates.append(date)
+                    onshelter = True
+                    inboundmovements.pop(0)
+        date = asm3.i18n.add_days(date, 1)
```
bobintetley, line 7356: If we're adding a list element for every single day that an animal is on shelter, won't this potentially make an absolutely huge list of elements while it's working?

Since we're only operating at the level of a month, we could add 28 days here instead of 1 and have far fewer entries in this list and it should still work the same I think?

src/asm3/movement.py
```diff
@@ -1159,3 +1162,19 @@ def send_movement_emails(dbo: Database, username: str, post: PostedData) -> bool
         for pid in post.integer_list("personids"):
             asm3.log.add_log_email(dbo, username, asm3.log.PERSON, pid, logtype, emailto, subject, body)
     return rv
+
+def is_exit_movement(dbo: Database, movemementtypeid: int) -> bool:    
+    if movemementtypeid in (0, 9, 10): # None, Reservation, Cancelled Reservation
+        return False
+    if movemementtypeid in (2, 12): # Foster, Permanent Foster
+        if asm3.configuration.foster_on_shelter(dbo):
```
bobintetley: couldn't these lines just be written as ``` return asm3.configuration.foster_on_shelter(dbo) ``` and ``` return asm3.configuration.retailer_on_shelter(dbo) ``` ?

src/asm3/animal.py

src/asm3/animal.py

src/asm3/movement.py
adamdspencer77: Yes they could. Done.

src/asm3/animal.py
adamdspencer77: Rewritten bypassing list of all on shelter days.

bobintetley:
Nobody likes writing them, but I'm not seeing any unit tests for this one!

adamdspencer77:
Added unit test. Am getting errors when I run all unit tests but checked on master and I get them there too. I don't think that these errors are related to my test.

src/asm3/animal.py
```diff
@@ -7284,3 +7287,85 @@ def create_waitinglist(dbo: Database, username: str, aid: int) -> int:
...
+    animals = dbo.query_list("SELECT ID FROM animal")
+    asm3.asynctask.set_progress_max(dbo, len(animals))
+    for i, a in enumerate(animals):
+        update_animal_figures_onshelter(dbo, a)
+        asm3.asynctask.set_progress_value(dbo, i)
+    return f"OK {len(animals)}"
+
+def update_animal_figures_onshelter(dbo: Database, animalid: int):
+    # Delete existing animalfiguresonshelter rows with this animalid
+    dbo.delete("animalfiguresonshelter", "AnimalID = %s" % animalid)
```
bobintetley: I'd do this delete right at the end just before the insert - otherwise if an error occurs we'll have deleted the rows and replaced them with nothing.

src/asm3/animal.py
```diff
@@ -7284,3 +7287,85 @@ def create_waitinglist(dbo: Database, username: str, aid: int) -> int:
...
+        month = int(f[0][0])
+        year = int(f[0][1])
+        values = {
+            "AnimalID": animalid,
+            "MonthMidPoint": datetime(year, month, 15),
+            "Month": month,
+            "Year": year,
+            "DaysOnShelter": f[1]
+        }
+        dbo.insert("animalfiguresonshelter", values, generateID=False, setCreated=False, writeAudit=False)
```
bobintetley: You should be able to get a big performance improvement here by using dbo.execute_many instead of running each insert one at a time. 

See functions like animal.update_all_animal_statuses for an example of how to use execute_many.

What it does is send all the queries to the DB in one go for execution, removing the overhead of grabbing/opening a valid connection, sending one query, checking and marshalling the result, returning the connection, repeating etc.

Issue #2101, opened 2026-06-10
Map view screen


bobintetley:
Similar to calendar view, add a new "Map view" option to the animal control menu. It will replace "Map of active incidents" in the menu. The map of recent incidents screen option will be removed.

The base of this will be the existing incident_map.js module. 

Similar to calendar view, it should have a bunch of checkboxes that enable and disable pins on the map. 

The categories will be:

* Active incidents (no completed date, dispatch lat long)
* Incidents completed in the last 90 days (dispatch lat long)
* Reclaims in the last 90 days (owner lat long)
* Non-shelter animals in the last 90 days (owner lat long)
* Lost animals (arealatlong - requires merging of #2097)
* Found animals (arealatlong - requires merging of #2097)

PR #2134, opened 2026-07-22
Upgrade Map Reports

adamdspencer77:
Plumb the map reports into the upgraded mapping library using the new pins.

Relies on https://github.com/sheltermanager/asm3/issues/2101 being merged first.

If you have a look at one of the map .rst files, you can see that map reports expect to have a column named "LatLong" for the geocode, and all of the other columns are joined together to make the description.

I think probably the way to do this is to have another named column that influences your style or icon/pin/whatever and have the report engine ignore that column when concatenating the others.


src/asm3/reports.py
```diff
@@ -1858,41 +1873,34 @@ def _GenerateMap(self) -> str:
...
-        self._Append( "}, 50);\n" )
-        self._Append("</script>")
+                
+            p.append({ "latlong": values[0], "POPUPTEXT": "".join(concat), "PINURL": f"static/images/mapping/{pincolour.lower()}.png" })
+
+        self._Append("<script type='text/javascript'>\n")
+        self._Append("  mapping.draw_map(\"embeddedmap\", 10, \"\", []);\n")
+        self._Append("  let ready = mapping.ready();\n")
+        self._Append("  ready.then(function() {\n")
+        self._Append("      window.setTimeout(async function() {\n")
```
bobintetley: This could probably be made a lot simpler. Put the window.setTimeout inside the mapping.ready function so that it is encapsulating the whole concept of being ready - not much point in having callers have to do it!

Then you can replace most of this with just:

``` 
mapping.draw_map(...);
mapping.ready().then(() => { let points = ... } 
```

src/static/js/common_map.js
```diff
@@ -37,7 +33,9 @@ const mapping = {
         let ticklength = 500;
         let maploaded = await mapping.check_map_loaded();
         let tickmessage = "Attempt " + attemptno + ". Mapping library unavailable, waiting " + ticklength + " milliseconds";
-        if (maploaded) { return true; }
+        if (maploaded) {
+            return true;
+        }
         window.setTimeout(async function() {
             log.info(tickmessage);
             maploaded = await mapping.check_map_loaded();
```
bobintetley: I think this whole _ready function is now redundant and can be removed?

src/asm3/reports.py

src/static/js/common_map.js
adamdspencer77: Done.

PR #2161, opened 2026-08-03
Online Forms - Postcode lookup button untidy when using Bootstrap

adamdspencer77:


src/asm3/onlineform.py
```diff
@@ -200,7 +200,6 @@ def _get_onlineform_html_bootstrap(dbo: Database, formid: int, completedocument:
             asm3.html.script_tag(TIMEPICKER_JS) + \
             asm3.html.asm_script_tag("onlineform_extra.js") + \
             "</head>"
-        header = header.replace("</head>", extra)
```
adamdspencer77, line 203: This was happening twice so I removed one of them.

PR #2170, opened 2026-08-05
Bulk Change Animals - Adopt after deletion

adamdspencer77:
Can select animals after deletion. If you then adopt, an adoption can be created that is not associated with an animal.

### DIFF
--- a/src/static/js/animal_bulk.js
+++ b/src/static/js/animal_bulk.js
@@ -191,10 +191,12 @@ $(function() {
                     let response = await common.ajax_post("animal_bulk", formdata);
                     header.hide_loading();
                     header.show_info(_("{0} animals successfully deleted.").replace("{0}", response));
+
                     // Remove deleted animals from animalchooser-multi to avoid them being reselected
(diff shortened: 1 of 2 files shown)


src/static/js/animal_bulk.js
```diff
@@ -191,6 +191,12 @@ $(function() {
                     let response = await common.ajax_post("animal_bulk", formdata);
                     header.hide_loading();
                     header.show_info(_("{0} animals successfully deleted.").replace("{0}", response));
+
+                    // Remove deleted animals from animalchooser-multi to avoid them being reselected
+                    $.each($("#animals").val().split(","), function(i, v) {
+                        $(".asm-animalchoosermulti-result a[href='animal?id=" + v + "']").parent().remove();
+                    });
+                    
                     $("#animals").animalchoosermulti("clear");
```
bobintetley: I'd put this code inside the animalchoosermulti widget behind a function, something like removeSelected so it just becomes a line in here for $("#animals").animalchoosermulti("removeSelected")

adamdspencer77: Done.

PR #2178, opened 2026-08-05
Lost and Found - Fill in area/postcode from person record

adamdspencer77:
If a person is selected and the arealost/areafound and areapostcode fields are blank, they are filled using the address and postcode from the person record.

### DIFF
--- a/src/static/js/lostfound_new.js
+++ b/src/static/js/lostfound_new.js
@@ -168,12 +168,12 @@ $(function() {
             
             // Auto fill area and postcode if not set
             $("#owner").on("change", async function(event, rec) {
-                let $areainput = $("#arealost");
+                let areainput = $("#arealost");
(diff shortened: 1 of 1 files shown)


src/static/js/lostfound_new.js
```diff
@@ -165,6 +165,18 @@ $(function() {
                 lostfound_new.update_breed_select();
                 additional.toggle_elements_by_species("additional", $("#species").val());
             });
+            
+            // Auto fill area and postcode if not set
+            $("#owner").on("change", async function(event, rec) {
+                let $areainput = $("#arealost");
```
bobintetley: wow, old school dollar variable name, nasty! just let areainput I think :D

adamdspencer77: Done.

PR #2179, opened 2026-08-06
lookups.py / insert_lookup and update_lookup

adamdspencer77:
Both of these functions have gotten out of hand as more and more lookups have been added to the system.

Refactor them. I think they could be hugely simplified by looking up the table in LOOKUP_TABLES and using the modifiers column to set the field dictionary to pass to a generate insert or update statement.

values = { t[LOOKUP_NAMEFIELD]: name }
if t[LOOKUP_DESCFIELD] != "": values[LOOKUP_DESCFIELD] = desc
if t[LOOKUP_MODIFIERS].find("cost") != -1: values["DefaultCost"] = defaultcost
...

Given how similar these functions are and that the only major difference is whether an update or insert is run, we could have a single "upsert_lookup" function and have a parameter that controls whether we're inserting or updating. insert_lookup and update_lookup could then call it and pass that parameter.

### DIFF
--- a/src/asm3/lookups.py
+++ b/src/asm3/lookups.py
@@ -1209,6 +1209,10 @@ def get_transport_types(dbo: Database) -> Results:
     return dbo.query("SELECT * FROM transporttype ORDER BY TransportTypeName")
 
 def upsert_lookup_from_form(dbo: Database, username: str, post: PostedData) -> int:
+    """
+    Insert or update a lookup using posted data, if the posted data contains a non-zero id, 
+    the row belonging to that ID is updated, otherwise a new row is inserted
+    """
     lookup = post["lookup"]
     name = post["lookupname"]
     desc = post["lookupdesc"]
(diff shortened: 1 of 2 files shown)


src/asm3/lookups.py
```diff
@@ -1425,6 +1208,50 @@ def get_transport_statuses(dbo: Database) -> Results:
 def get_transport_types(dbo: Database) -> Results:
     return dbo.query("SELECT * FROM transporttype ORDER BY TransportTypeName")
 
+def upsert_lookup_from_form(dbo: Database, username: str, post: PostedData) -> int:
```
bobintetley, line 1211: Comment to explain what the function does and requires an id for updates, etc

src/asm3/lookups.py
```diff
@@ -1425,6 +1208,50 @@ def get_transport_statuses(dbo: Database) -> Results:
...
+    if lookup in ("lkownerflags", "lkanimalflags"):
+        name = name.replace(",", " ").replace("|", " ").replace("'", " ") # Remove bad chars
+        name = asm3.utils.strip_duplicate_spaces(name) # Strip dup spaces Bad   Flag->Bad Flag
+    else:
+        # Only provide description in not owner or animal flags
+        data[desccol] = desc
+
+    data[namecol] = name
+    
+    if post.integer("id"):
```
bobintetley: I'd test post.integer("id") != 0 for clarity

src/asm3/lookups.py
adamdspencer77: Done.

src/asm3/lookups.py
adamdspencer77: Done.

PR #2183, opened 2026-08-11
Person multi chooser: Flag filter not being applied as part of the search

adamdspencer77:
As title says, when choosing a flag, it is not being submitted as part of the search to the backend. It's only being applied to the results of the search, which are only limited to 100 rows.

The flag filter should be left of the search button and included when clicking search so the results are filtered at the backend.

### DIFF
--- a/src/asm3/person.py
+++ b/src/asm3/person.py
@@ -637,7 +637,7 @@ def get_investigation(dbo: Database, personid: int, sort: int = ASCENDING) -> Re
     return dbo.query(sql, [personid])
 
 def get_person_find_simple(dbo: Database, query: str, username: str = "", classfilter: str = "all", typefilter: str = "all", 
-                           includeStaff: bool = False, includeVolunteers: bool = False, limit: int = 0, siteid: int = 0) -> Results:
+        includeStaff: bool = False, includeVolunteers: bool = False, limit: int = 0, siteid: int = 0, flags: List[str] = []) -> Results:
     """
     Returns rows for simple person searches.
     query: The search criteria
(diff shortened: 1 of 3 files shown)


src/asm3/person.py
```diff
@@ -681,14 +681,24 @@ def get_person_find_simple(dbo: Database, query: str, username: str = "", classf
         "individual":   " AND o.OwnerType=1",
         "organization": " AND o.OwnerType=2"
     }
+    flagfilterssql = ""
+    if flags:
+        flagfilterssql += " AND ("
+        flagfilters = []
+        colconcat = dbo.sql_concat(("'|'", "o.AdditionalFlags", "'|'"))
```
bobintetley: this is a bit tricksy! I'm not keen on colconcat, or the way flagfiltersql is being declared with the AND ( bit up here. Makes it hard to see what flagfilterssql actually contains or is being compared to when reading the code

src/asm3/person.py
```diff
@@ -681,14 +681,24 @@ def get_person_find_simple(dbo: Database, query: str, username: str = "", classf
...
+        for flag in flags:
+            pipedflag = f"%|{flag}|%"
+            flagfilters.append("%s LIKE '%s'" % (colconcat, pipedflag))
```
bobintetley: Just use a bracketed OR instead of concat above. Makes pipedflag redundant too.

```
flagfilters.append(f"(o.AdditionalFlags LIKE '{flag}|%' OR o.AdditionalFlags LIKE '|{flag}|%')")
```

src/asm3/person.py
```diff
@@ -681,14 +681,24 @@ def get_person_find_simple(dbo: Database, query: str, username: str = "", classf
...
+        flagfilterssql += " OR ".join(flagfilters)
+        flagfilterssql += ")"
```
bobintetley: I think it's nicer if we declare flagfiltersql here at this point after the filters for readability

```
flagfiltersql = " OR ".join(flagfilters)
flagfiltersql = f" AND ( {flagfiltersql} )"
```

src/asm3/person.py

src/asm3/person.py

src/asm3/person.py
adamdspencer77: Hopefully sorted after sorting other feedback below.

PR #2185, opened 2026-08-17
Incoming form permissions (DBU 35128)

adamdspencer77:
Some shelters would find it useful to be able to apply one or more roles to an online form. The roles would prevent users without one of those roles from seeing submissions of that type of form in the incoming screen.

For example, one online form might be surrender info that staff don't want volunteers to see, and another might be adoption applications that volunteers need to process.

Requires onlineformrole table.


src/asm3/onlineform.py
```diff
@@ -2336,3 +2353,15 @@ def create_animal_log(dbo: Database, username: str, collationid: int):
     else:
         raise asm3.utils.ASMValidationError(asm3.i18n._("Unable to match to an animal record (need animalname).", dbo.locale))
     return (collationid, animalid, animalname, 1)
+
+def reduce_incoming_results(dbo: Database, username: str, rows: Results) -> Results:
+    if len(rows) == 0: return rows
+    roles = asm3.users.get_roles_ids_for_user(dbo, username)
+    for r in rows.copy():
+        formid = dbo.query_int("SELECT ID FROM onlineform WHERE Name = ?", (r["FORMNAME"],))
+        viewroles = dbo.query_list("SELECT RoleID FROM onlineformrole WHERE OnlineFormID = ?", (formid,))
```
bobintetley: People can have thousands of form submissions sitting in the queue, so running 2 queries for every single iteration of this loop could have heavy performance impacts on databases with lots of forms in the queue.

Since the whole contents of these tables are never going to be that big, I would rewrite the function so that at the beginning you pull back a set of all IDs and Names from onlineform, and another set that is the entire contents of the onlineformrole table. 

You can then just have some simple nested functions that use loops to get the id for a form name, and retrieve any role permissions for that form.

src/asm3/onlineform.py
```diff
@@ -2336,3 +2353,15 @@ def create_animal_log(dbo: Database, username: str, collationid: int):
...
+        if not viewroles: continue
+        for role in roles:
+            if role not in viewroles:
+                rows.remove(r)
```
bobintetley: I think I gave you duff info, we do also need to check here if the person is a superuser. Superuser does still get to see everything even if they don't have the specific role.

src/asm3/onlineform.py

src/asm3/onlineform.py
adamdspencer77: Hopefully sorted. I put the superuser check in `get_onlineformincoming_headers`, I hope that's OK.

src/asm3/onlineform.py
```diff
@@ -852,15 +855,19 @@ def get_onlineformincoming_formfooter(dbo: Database, collationid: int) -> str:
         "INNER JOIN onlineformincoming oi ON oi.FormName = o.Name " \
         "WHERE oi.CollationID = ?", [collationid])
 
-def get_onlineformincoming_headers(dbo: Database) -> Results:
+def get_onlineformincoming_headers(dbo: Database, username: str, superuser: int = 0) -> Results:
```
bobintetley: I think it might be neater to pass the session object here rather than username and superuser. 

``` (dbo: Database, session: Session) ```

the caller in main.py will have access to o.session that you can pass here.

The session object has user and superuser properties. It also has one for "roles" and "roleids" so you don't need to re-read them from the user table.

File: src/asm3/onlineform.py [hunk 4]
```diff
@@ -2336,3 +2356,32 @@ def create_animal_log(dbo: Database, username: str, collationid: int):
     else:
         raise asm3.utils.ASMValidationError(asm3.i18n._("Unable to match to an animal record (need animalname).", dbo.locale))
     return (collationid, animalid, animalname, 1)
+
+def reduce_incoming_results(dbo: Database, username: str, rows: Results) -> Results:
+    if len(rows) == 0: return rows
+    forms = dbo.query("SELECT ID, Name FROM onlineform")
+    formroles = dbo.query("SELECT OnlineFormID, RoleID FROM onlineformrole")
+    roles = asm3.users.get_roles_ids_for_user(dbo, username)
```
bobintetley: If this function accepted the session, you can use session.roleids to get the user's roles without having to make a database call.

src/asm3/onlineform.py
```diff
@@ -852,15 +855,19 @@ def get_onlineformincoming_formfooter(dbo: Database, collationid: int) -> str:
...
         "(SELECT MAX(Value) FROM onlineformincoming WHERE CollationID=f.CollationID AND FieldName='mergeperson') AS MergePerson, " \
         "CASE WHEN EXISTS(SELECT Value FROM onlineformincoming WHERE CollationID=f.CollationID AND FieldName='processed') THEN 1 ELSE 0 END AS Processed, " \
         "CASE WHEN EXISTS(SELECT Value FROM onlineformincoming WHERE CollationID=f.CollationID AND FieldName='spam') THEN 1 ELSE 0 END AS Spam " \
         "FROM onlineformincoming f " \
         "GROUP BY f.CollationID, f.FormName, f.PostedDate, f.Host, f.Preview " \
         "ORDER BY f.PostedDate")
+    if superuser:
+        return rows
+    else:
+        return reduce_incoming_results(dbo, username, rows)
```
bobintetley: If you pass the session to reduce_incoming_results, it can do the short circuiting of returning the set if the user is a superuser, and it gets rid of this condition and keeps the code more readable (imo)

src/asm3/onlineform.py

src/asm3/onlineform.py

src/asm3/onlineform.py

src/asm3/onlineform.py
```diff
@@ -2336,3 +2353,29 @@ def create_animal_log(dbo: Database, username: str, collationid: int):
...
+        formname = r["FORMNAME"]
+        for form in forms:
+            if form["NAME"] == formname:
+                formid = form["ID"]
+                break
+        viewroles = []
+        for formrole in formroles:
+            if formrole["ONLINEFORMID"] == formid:
+                viewroles.append(formrole["ROLEID"])
+        if not viewroles: continue
```
bobintetley: You could create a dictionary before the outer loop, and store the viewroles list in it for the formid once you've constructed it.

Since many of the forms in the queue will be the same, it will save iterating the onlineformrole table every time for forms we already got the roles for.  You can just pull the roles from the dictionary for that form if we've seen it before.

PR #2187, opened 2026-08-21
Add caller, suspect and victim address and phone numbers to incident emails

adamdspencer77:
As title, incident emails only currently include the names of callers, victims and suspects. It has been requested we add the addresses and phone number information for these people as well.

### DIFF
--- a/src/static/js/incident.js
+++ b/src/static/js/incident.js
@@ -303,14 +303,33 @@ $(function() {
                     }
                 });
                 let i = controller.incident;
-                let msg = [ 
-                    _("Type") + ": " + i.INCIDENTNAME,
-                    _("Date/Time") + ": " + format.date(i.INCIDENTDATETIME) + " " + format.time(i.INCIDENTDATETIME),
-                    _("Address") + ": " + i.DISPATCHADDRESS + ' ' + i.DISPATCHTOWN + ' ' + i.DISPATCHCOUNTY + ' ' + i.DISPATCHPOSTCODE,
-                    _("Notes") + ": " + common.nulltostr(i.CALLNOTES),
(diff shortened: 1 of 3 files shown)


src/static/js/incident.js
```diff
@@ -303,14 +303,33 @@ $(function() {
...
+
+                let formatcontact = function(data) {
+                    let output = "";
+                    $.each(data, function(i, v) {
+                        if (v) {
+                            output += v + "<br />";
+                        }
+                    });
+                    return output.trim();
+                }
```
bobintetley: missing semicolon :)

src/static/js/incident.js
```diff
@@ -303,14 +303,33 @@ $(function() {
...
-                    _("Suspect") + ": " + common.nulltostr(i.OWNERNAME1)
+
+                let formatcontact = function(data) {
+                    let output = "";
+                    $.each(data, function(i, v) {
+                        if (v) {
+                            output += v + "<br />";
+                        }
+                    });
+                    return output.trim();
```
bobintetley: This whole function body could be replaced with:

``` return data.filter(v => v).join("<br>"); ```

Array.filter is ES6 so available to us. Filter accepts a callback function and only includes elements where the function returns true. If you return the value itself then it evaluates it for truthiness., hence why v => v (or function(v) { return v; }) works.

This seems like something we might find useful elsewhere. Maybe this could be added to common.js so that we can call common.join_not_null(array)

adamdspencer77: Went with common.join_not_falsy(array) instead.

src/static/js/incident.js
adamdspencer77: The function it was in has gone now.

PR #2195, opened 2026-09-02
Multiple waiting lists (DBU 35129)

adamdspencer77:
While the waiting list's original intent was for people surrendering animals, there are shelters in places where the demand for animals outstrips the supply. People sometimes use it as a waiting list for adopters.

Occasionally, we get a customer who wants to do both. Allow multiple waiting lists, requires a new lookup table lkwaitinglisttype - this will be a dropdown on the waiting list screen so that each list can be chosen. lkwaitinglisttype will have one default entry for "Surrenders". More types can be added under the lookup screen for Adopters or whatever someone might want to do.

animalwaitinglist will need a new column WaitingListTypeID

Rank needs to be calculated for each waiting list type.


src/static/js/waitinglist_results.js
```diff
@@ -272,7 +275,10 @@ $(function() {
                 rv = row.BREEDNAME;
             }
             else if (name == "WaitingListRemovalID") {
-                rv = row.WaitingListRemovalName;
+                rv = row.WAITINGLISTREMOVALNAME;
```
adamdspencer77: Spotted this unrelated typo and fixed it while I was in here.

src/asm3/dbupdate.py
```diff
@@ -35,7 +35,7 @@
     "jurisdiction", "licencetype", "lkadoptionsource", "lkanimalflags", "lkboardingtype", "lkclinicinvoiceitems", "lkclinictype", "lkcoattype", "lkmediaflags", 
     "lkcondition", "lksconditiontype", "lkownerflags", "lkproducttype", "lksaccounttype", "lksclinicstatus", "lksdiarylink", "lksdonationfreq", "lksentrytype",
     "lksex", "lksfieldlink", "lksfieldtype", "lksize", "lktaxrate", "lksloglink", "lksmedialink", "lksmediatype", "lksmedicaltype", 
-    "lksmovementtype", "lksoutcome", "lksposneg", "lksrotatype", "lksunittype", "lksyesno", "lksynun", "lksynunk", 
+    "lksmovementtype", "lksoutcome", "lksposneg", "lksrotatype", "lksunittype", "lksyesno", "lksynun", "lksynunk", "lkwaitinglisttype", 
     "lkstransportstatus", "lkurgency", "lkwaitinglistremoval", "lkworktype",
```
bobintetley, line 39: I would have put lkwaitinglisttype here next to the other waitinglist lookup table :D

src/asm3/waitinglist.py
```diff
@@ -90,7 +92,7 @@ def get_waitinglist_ranks(dbo: Database) -> Dict[int, int]:
 
 def get_waitinglist(dbo: Database, priorityfloor: int = 5, species: int = -1, size: int = -1, 
                     addresscontains: str = "", includeremoved: int = 0, namecontains: str = "", 
-                    descriptioncontains: str = "", siteid: int = 0) -> Results:
+                    descriptioncontains: str = "", siteid: int = 0, wltype: int = -1) -> Results:
     """
     Retrieves the waiting list
     priorityfloor: The lowest urgency to show (1 = urgent, 5 = lowest)
```
bobintetley, line 98: include wltype in the doc string

src/asm3/wordprocessor.py
```diff
@@ -1922,6 +1922,7 @@ def waitinglist_tags(dbo: Database, a: ResultRow) -> Tags:
         "CANAFFORDDONATION":        asm3.utils.iif(a["CANAFFORDDONATION"] == 1, _("Yes", l), _("No", l)),
         "URGENCY":                  a["URGENCYNAME"],
         "COMMENTS":                 a["COMMENTS"],
+        "WAITINGLISTTYPE":          a["WAITINGLISTTYPENAME"],
```
bobintetley, line 1925: little organisational thing, but a field as important as type should be near the top of the list rather than at the end :)

src/asm3/dbupdate.py
adamdspencer77: Done.

src/asm3/waitinglist.py
adamdspencer77: Done.

src/asm3/wordprocessor.py
adamdspencer77: Sorted.

PR #2196, opened 2026-09-03
New role permissions: Change animal/incident/owner permissions

adamdspencer77:
Add new role permissions for:

Change animal permissions
Change incident permissions
Change person permissions

These permissions control whether the user can change the role permissions when editing those types of records.

They should still be allowed to see what the permissions are, so this isn't a case of hiding the role multiselect. I can't remember if the selectmulti widget has the ability to be disabled. If it doesn't, this should be added to it.

I think disabling the selectmulti widget should hide its select dropdown and the close icons on any selected elements, so you can see but not interact with it.


src/static/js/common_widgets.js
```diff
@@ -1026,6 +1026,27 @@ $.fn.selectmulti = asm_widget({
         t.change(); 
     },
 
+    /** Leave widget visible but non-editable */
+    disable: function(t) {
+        t.parent().find(".asmSelect").hide();
+        t.parent().find("ol").hide();
+        let rolenames = [];
```
bobintetley: seems weird calling this variable rolenames when it's a generic widget disable function? They're just selected elements I think.

adamdspencer77: Changed variable name to "selected".

PR #2197, opened 2026-09-03
Home page: Animal quick link expander

adamdspencer77:
Some people set the number of animal quick links displayed on the home page to a huge amount (the customer that kicked this off set it to 200). This pushes virtually all of the rest of the home page content off screen so users can't see it.

While still allowing them to configure how many appear, the home page should show something like a maximum of 10 animal quick links and an expander that shows the rest of them.

### DIFF
--- a/src/static/js/main.js
+++ b/src/static/js/main.js
@@ -442,7 +442,7 @@ $(function() {
         },
 
         render_animal_links: function() {
-            let hidemorethan = 15;
+            let hidemorethan = 9;
             let s = [];
             let linknames = { "recentlychanged": _("Recently Changed"), 
                 "recentlyentered": _("Recently Entered Shelter"),
@@ -472,7 +472,7 @@ $(function() {
                     s.push('<p class="asm-menu-category" style="border-bottom: none;">'),
                     s.push('<a id="animallinktoggle" href="#">'),
                     s.push('<span id="animallinknav" class="ui-icon ui-icon-triangle-1-e"></span>'),
-                    s.push(_("Expand")),
+                    s.push('<span id="animallinklabel">' + _("more") + '</span>'),
                     s.push('</a>'),
                     s.push('</p>')
                 }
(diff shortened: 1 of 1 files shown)


src/static/js/main.js
```diff
@@ -442,6 +442,7 @@ $(function() {
         },
 
         render_animal_links: function() {
+            let hidemorethan = 15;
```
adamdspencer77: I thought that 15 looked better on screen than 10. Left this as a variable so it can be tweaked. I wondered if you would prefer this value to be plucked from sitedefs rather than hard coded at all.

src/static/js/main.js
```diff
@@ -458,10 +459,23 @@ $(function() {
...
+                    }
+
                     s.push(html.animal_link_thumb(a, { showlocation: true }));
                     s.push("</div>");
                 });
+                if (controller.animallinks.length > hidemorethan) {
+                    s.push('<p class="asm-menu-category" style="border-bottom: none;">'),
+                    s.push('<a id="animallinktoggle" href="#">'),
+                    s.push('<span id="animallinknav" class="ui-icon ui-icon-triangle-1-e"></span>'),
+                    s.push(_("Expand")),
```
adamdspencer77: I couldn't make up my mind what would be the best label here. I flip flopped between "Show all", "More" and "Expand". What do you think?

src/static/js/main.js
bobintetley: 15 seems like too many, also it's difficult because they aren't a uniform width and can expend. On mine, 15 overflows by one animal on a resolution of 2560x1440. 

I'd go with 9 as that is the default value in the config because it typically fits on one line at 1280px wide.

src/static/js/main.js
bobintetley: We switch between "more" and "less" for the expander text elsewhere on the screen, like the message board.

src/static/js/main.js
adamdspencer77: Done.

src/static/js/main.js
adamdspencer77: Done.

PR #2206, opened 2026-09-14
Allow tabbed content to be searchable. This would make finding particular options easier, amongst other things.

adamdspencer77:


src/static/js/common_widgets.js
```diff
@@ -1285,6 +1285,40 @@ $.fn.asmtabs = asm_widget({
             t.trigger("changeTab", [ self.active(t) ]);
         };
         t.tabs(options);
+        if (t.attr("data-searchable") == "true") {
+            t.find(".asm-tab-search").show();
+            $('li[aria-controls="tab-tab-tabsearch"]').on("click", function() {
+                console.log("Clicked!");
+                $("#tab-tab-tabsearch input").focus();
```
bobintetley, line 1291: console log left in :)

does the tab widget not focus the first widget automatically? seems odd that you have to do this

src/static/js/options.js
```diff
@@ -1227,6 +1227,7 @@ $(function() {
         },
 
         sync: function() {
+            $("#asm-options-panel").data("xoptions", {"searchable": true});
```
bobintetley: I don't think this needs to be here? You're already using the searchable HTML attribute to trigger search functionality

src/static/js/common_widgets.js
adamdspencer77: Console log removed.

adamdspencer77: Doesn't focus automatically when I remove the focus line.

src/static/js/options.js
adamdspencer77: You're right. Removed.

Issue #2191, opened 2026-08-26
Home page: Animal quick link expander


bobintetley:
Some people set the number of animal quick links displayed on the home page to a huge amount (the customer that kicked this off set it to 200). This pushes virtually all of the rest of the home page content off screen so users can't see it.

While still allowing them to configure how many appear, the home page should show something like a maximum of 10 animal quick links and an expander that shows the rest of them.

PR #2207, opened 2026-09-15
Expander Widget

adamdspencer77:
The thought occurs based on https://github.com/sheltermanager/asm3/issues/2191 that we have quite a few places now where we want to show expanded text and we're duplicating the same code to show the expander icon or a "more" style link.

Create a new widget in common_widgets.js called expander. I'm thinking that you feed it a div that has 2 div child elements. The first is the unexpanded content, the second is the expanded content.

<div id="ex1" class="asm-expander" style="display: none" mode="arrow">
    <div>unexpanded content</div>
    <div>expanded content</div>
</div>

The widget will then add the expander/arrow icon and use it to flip between showing the unexpanded and expanded content.

The mode attribute determines whether the expander shows the arrow icon to the left, or a "more..." or "less ..." type link at the end of the unexpanded content. We could have attributes to set the text of these for that mode.

Also needs to make sure the widget loads for the new asm-expander class in common.js/bind_widgets

Finally, we need to update some areas of the system to use the new widget - like the home page links, the message board on the home page, etc.

### DIFF
--- a/src/static/js/common_widgets.js
+++ b/src/static/js/common_widgets.js
@@ -1416,37 +1416,50 @@ $.fn.currency = asm_widget({
 
 /** Widget that allows you to display data, hiding more related data that can be shown by "expanding" the widget
  *  classname is the name of the class shared by elements that are subjec to this expander
- *  where more elements with classname than limit are present, the expander will hide he overflow
+ *  where more elements with classname than limit are present, the expander will hide the overflow
+ * 
+ *  If the target element has class "asm-expander" the widget will initialise automatically with the default options. 
+ *  To pass custom options, do not include the "asm-expander" class. Instead, initialise manually with your custom options
+ *  eg. $("#asm-homepage-animals").expander({limit: 9});
  */
-$.fn.expander = function(classname, limit) {
-    let container = this;
-    container.css("cursor", "pointer");
-    $.each(container.find("." + classname), function(i, v) {
-        if ( i >= limit ) {
-            $(v).addClass("asm-overflow");
-            $(v).hide();
+$.fn.expander = asm_widget({
+    options: {
+        limit: 10,
+        classname: "asm-expander-item"
+    },
+    _create: function(t, options) {
+        var container = t;
+        var classname = options.classname;
+        var limit = options.limit;
+        container.css("cursor", "pointer");
+        let elements = container.find("." + classname);
+        $.each(elements, function(i, v) {
+            if ( i >= limit ) {
+                $(v).addClass("asm-overflow");
+                $(v).hide();
+            }
+            
+        });
+        if ( container.find("." + classname).length > limit ) {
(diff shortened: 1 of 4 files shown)


src/static/js/common_widgets.js
```diff
@@ -1414,6 +1414,98 @@ $.fn.currency = asm_widget({
...
+ *  The item limit and classname may be set by adding data-expander-limit or data-expander-classname values to t
+ *  
+ *  The expander mode may be select by a adding data-expander-mode to t
+ *  "inline" will place an icon based toggle to the left of the content
+ *  "trailing" will place a text based toggle after the content
+ *  "below" will place both an icon and text based toggle below the content
+ *  
+ *  If no classname is set, raw text is assumed and limit refers to the number of characters before overflow
+ *  
+ *  eg. '<div class="asm-expander" data-expander-mode="inline" data-expander-limit="20" data-expander-classname="someclass">'
```
bobintetley: I'd include an example of each type of mode for clarity

src/static/js/main.js
```diff
@@ -540,23 +526,23 @@ $(function() {
...
             s.push('</tr></tbody></table>');
 
-            $.each(controller.mess, function(i, m) {
-                s.push('<input id="long' + m.ID + '" type="hidden" value="' + html.title(common.replace_all(m.MESSAGE, "\n", "<br/>")) + '" />');
-                s.push('<input id="short' + m.ID + '" type="hidden" value="' + html.title(html.truncate(m.MESSAGE)) + '" />');
-            });
+            // $.each(controller.mess, function(i, m) {
+            //     s.push('<input id="long' + m.ID + '" type="hidden" value="' + html.title(common.replace_all(m.MESSAGE, "\n", "<br/>")) + '" />');
+            //     s.push('<input id="short' + m.ID + '" type="hidden" value="' + html.title(html.truncate(m.MESSAGE)) + '" />');
+            // });
```
bobintetley: Not sure if either of these loops to output hidden fields are necessary any more since the expander widget has replaced the old toggle code?

src/static/lib/asmselect/1.0.4a/jquery.asmselect.js
```diff
@@ -42,7 +42,9 @@
 
             };
 
-        $.extend(options, customOptions); 
+        $.extend(options, customOptions);
+
+        console.log(this);
```
bobintetley: debug code left in :)

src/static/js/common_widgets.js

src/static/js/main.js
adamdspencer77: Removed.

src/static/lib/asmselect/1.0.4a/jquery.asmselect.js
adamdspencer77: Removed.

